PR #6, opened 2016-11-22
allow to configure bitrise config and bitrise secrets path

godrei:


erosdome:
@viktorbenei check dis

PR #21, opened 2017-01-24
load spec handler

godrei:


apiserver/api_server.go
```diff
@@ -253,6 +254,39 @@ func saveBitriseYMLFromJSONHandler(w http.ResponseWriter, r *http.Request) {
 	respondWithJSON(w, 200, SimpleResponse{Message: "OK"})
 }
 
+func loadSpecHandler(w http.ResponseWriter, r *http.Request) {
+	// Load steplib spec
+	cmd := command.New("stepman", "collections", "--format", "json")
```
viktorbenei: not good, `stepman` can't be accessed directly _outside_ of the `bitrise cli`

viktorbenei: maybe we should add a `stepman` command to `bitrise cli`, to expose it? so that `stepman collection ...` would be `bitrise stepman collection ...`?

apiserver/api_server.go
```diff
@@ -253,6 +254,39 @@ func saveBitriseYMLFromJSONHandler(w http.ResponseWriter, r *http.Request) {
 	respondWithJSON(w, 200, SimpleResponse{Message: "OK"})
 }
 
+func loadSpecHandler(w http.ResponseWriter, r *http.Request) {
+	// Load steplib spec
+	cmd := command.New("bitrise", "stepman", "collections", "--format", "json")
+	out, err := cmd.RunAndReturnTrimmedCombinedOutput()
+	if err != nil {
+		respondWithErrorMessage(w, "Failed to get steplib spec, error: %s", err)
```
viktorbenei: shouldn't this log the output too? at least with `log.Println(out)` or something like that?

PR #24, opened 2017-01-26
Api server

godrei:


apiserver/service/default_options.go
```diff
@@ -0,0 +1,24 @@
...
+
+// GetDefaultOutputsHandler ...
+func GetDefaultOutputsHandler(w http.ResponseWriter, r *http.Request) {
+	type EnvItmModel map[string]string
+
+	type ResponseModel struct {
+		FromBitriseCLI []EnvItmModel `json:"from_bitrise_cli"`
+	}
+
+	RespondWithJSON(w, 200, ResponseModel{
```
godrei: 200 -> should be http.StatusOk

Issue #65, opened 2017-03-05
Failed to load workflows.


bootstraponline:
I installed the latest version of bitrise cli & the workflow editor plugin. I created a folder with `bitrise.yml`. The workflow editor is failing to load the workflow. Am I doing something wrong?

<img width="593" alt="failed_to_load_workflow" src="https://cloud.githubusercontent.com/assets/1173057/23589430/e6fa9de2-019a-11e7-8d87-43a277077da2.png">

- https://gist.github.com/bootstraponline/01c50f24ecd7f971bf6394e1699abc99

```
$ bitrise version
1.5.4
$ go version
go version go1.8 darwin/amd64
$ bitrise :workflow-editor version
0.9.6
```

```
$ bitrise :workflow-editor
Starting API server at http://localhost:3645
Open workflow editor in browser ...
 => GET: /api/bitrise-yml.json - 10.065492ms ()
empty output data
```

```javascript
		}, function(error) {
			viewModel.workflows = undefined;
			console.log(error.message);
			viewModel.configureWorkflowsProgress.error(new Error("Failed to load workflows."));
		});
```

```
POST http://localhost:3645/api/spec 400 (Bad Request)
(anonymous) @ VM228:2
(anonymous) @ main.js:23338
sendReq @ main.js:23131
serverRequest @ main.js:22841
processQueue @ main.js:27448
(anonymous) @ main.js:27464
$eval @ main.js:28716
$digest @ main.js:28532
$apply @ main.js:28824
done @ main.js:23162
completeRequest @ main.js:23360
requestLoaded @ main.js:23301
main.js:57271 Error loading library
```

bootstraponline:
looks like the same issue as https://github.com/bitrise-io/bitrise-workflow-editor/issues/39 😭

bootstraponline:
closing as a dupe

bootstraponline:
seems to be a unique issue, reopening.

viktorbenei:
Did you run any `bitrise run ..` on your Mac?

bootstraponline:
I don't think so. maybe a long time ago

viktorbenei:
Error comes from here: https://github.com/bitrise-io/bitrise-workflow-editor/blob/23d73ccda3793c30420cbd14312cc30ad1dc7390/apiserver/tools/tools.go#L36

What's the output if you run:

```
bitrise "stepman" "collections" "--format" "json"
```

bootstraponline:
```
[...]

PR #66, opened 2017-03-05
Add usage documentation to readme

bootstraponline:
#65

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,14 @@
 # Bitrise Workflow Editor
 
+## Usage
+
+```
+brew update && brew install bitrise
+bitrise setup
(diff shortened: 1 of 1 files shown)


erosdome:
ReadMe updated, already available on master, will be included in next release, so closing this. Also, the workflow editor is now part of the Bitrise CLI core plugins, thus it is not needed to be installed as a separate plugin.

viktorbenei:
Yeah, just updated the README two days ago (https://github.com/bitrise-io/bitrise-workflow-editor/pull/126), forgot to check if there's any related issue. 

@bootstraponline let us know if the current setup & usage guide would not be enough!

PR #85, opened 2017-04-27
development now available in docker

erosdome:


erosdome:
@godrei please review, it makes changes to the API server

PR #151, opened 2017-05-31
ionic project type

doomsayer13:


erosdome:
@doomsayer13 resolve merge conflict - since then, fastlane project type has also been added

PR #183, opened 2017-07-18
is exposed for files and secrets

doomsayer13:


source/stylesheets/_tab-code_signing.scss.erb
```diff
@@ -484,4 +445,140 @@
...
+					}
+
+					.expose {
+						@include display-flex;
+
+						.checkbox {
+							width: 36px;
+							height: 16px;
+
+							svg[check] {
```
erosdome: Update with the span wrapper mentioned earlier

source/stylesheets/_tab-secrets-env_vars.scss.erb
```diff
@@ -122,6 +128,20 @@
 						}
 					}
 
+					.is-expand {
+						@include order(1);
```
erosdome: Order just changes the elements' order *visually*. But e. g. tabbing order is not changed. There should be a better solution.

source/stylesheets/_tab-code_signing.scss.erb
```diff
@@ -258,34 +259,48 @@
...
 
 					form.password {
 						@include display-flex;
 
+						.input-with-show-password {
+							position: relative;
+							@include display-flex;
+						}
+
 						input:not([type="submit"]) {
```
erosdome: This will also select the eye checkbox. You should add a class to the input field (like `input.password`, and modify this selector to use that, and not the `:not(...)`.

source/images/opened-eye.svg
```diff
@@ -0,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 764.3 400">
```
erosdome: Remake this, but not cropping the image area to this content, but to the crossed eye content, so they will be aligned to each other.

source/stylesheets/_tab-secrets-env_vars.scss.erb
doomsayer13: I just wanted to change the order visually :D

spec/javascripts/VariableSpec.js
```diff
@@ -330,6 +330,67 @@ describe("Variable", function() {
...
+			var meta = {
+				is_expose: true,
+				is_updated: false
+			};
+
+			expect(variable.meta(meta)).toEqual(meta);
+			expect(variable.userVariableConfig.opts.meta).toEqual(meta);
+		});
+
+	});
```
erosdome: Also write a spec in which the overriding meta does not contain a key that was originally included. Actually multiple specs: one where originally included in the default config, and one where in the user config, and one where in both.

spec/javascripts/VariableSpec.js
```diff
@@ -330,6 +330,67 @@ describe("Variable", function() {
...
+					}
+				}
+			}, defaultVariableConfig);
+
+			var meta = {
+				is_expose: true,
+				is_updated: false
+			};
+
+			expect(variable.meta(meta)).toEqual(meta);
```
erosdome: `variable.meta(meta)` is a setter. Although in our case it also returns the value it sets, it is not a required behaviour from getter-setter methods' setter mode (sometimes setters return true/false, indicating success/fail) and even in our case it might change in the future. So we should not expect the method to return the meta when used as a setter. So first use this separately as a setter, and in the expect, only use it as a getter.

source/javascripts/components/_Variable.js.erb
```diff
@@ -143,6 +146,22 @@ angular.module("BitriseWorkflowEditor").factory("Variable", function() {
...
+	};
+
+	Variable.prototype.isExpose = function(newIsExposedValue) {
+		var newMetaValue;
+		if (newIsExposedValue !== undefined) {
+			newMetaValue = angular.copy(this.meta());
+			newMetaValue.is_expose = newIsExposedValue;
+		}
+
+		var newMeta = optionGetterSetter(this, "meta", newMetaValue);
```
erosdome: `newMeta` -> `meta`. Also, we should rather call `newIsExposedValue` -> `isExpose` and `newMetaValue` -> `meta`.

source/javascripts/components/_Variable.js.erb
```diff
@@ -143,6 +146,22 @@ angular.module("BitriseWorkflowEditor").factory("Variable", function() {
...
+		
+		return newMeta.is_expose;
```
erosdome: Check if `newMeta` (or after the suggested change, `meta`) is defined. If not, return undefined.

PR #202, opened 2017-10-05
popover for is_expose toggle

doomsayer13:


source/stylesheets/_tab-code_signing.scss.erb
```diff
@@ -336,6 +345,7 @@
 
 					.expose {
 						@include display-flex;
+						width: auto;
```
erosdome: `width: auto` is the default, is this needed? Why?

data/strings.yml
```diff
@@ -120,6 +120,7 @@ workflows:
     details: Are you sure you want to delete the <workflow_id> workflow?
 code_signing:
   available_project_types_load: Loading, wait a sec...
+  expose_for_pr_popover_content: You can decide which files should be exposed for / available in Pull Request builds. <br> Be careful, exposing a secret is a potential security risk.
```
erosdome: Instead of *secret*, shouldn't it be something like *private file* or something? @viktorbenei so they won't confuse it with the secret env vars.

Also, remove whitespace before `<br>`

data/strings.yml
```diff
@@ -260,6 +261,7 @@ env_vars:
   add_new: Add new
   replace_variables_label: "Replace variables in inputs?"
   expose_for_pr_label: "Expose for Pull Requests?"
+  expose_for_pr_popover_content: You can decide which secrets should be exposed for / available in Pull Request builds. <br> Be careful, exposing a secret is a potential security risk.
```
erosdome: Remove whitespace before `<br>`

source/stylesheets/_tab-code_signing.scss.erb
doomsayer13: behave of the element was not the same without this

erosdome: That means that it is not `auto` without this. But I don't see where it is defined anything else. And if it is not defined anywhere, it is `auto`.

doomsayer13: here: https://github.com/bitrise-io/bitrise-workflow-editor/pull/202/files/b374a5b4f86fc505f878e854cdef3a53c30df3db#diff-5b6bc2fa079592de2108e6b49f8db44bR174

PR #203, opened 2017-10-06
popover z-index fix - mobile view

doomsayer13:


source/stylesheets/_constants.scss
```diff
@@ -30,12 +30,12 @@ $zindex-checkbox: 0;
 $zindex-step_description_open: 0;
 $zindex-file_uploader_input: 10;
 $zindex-dragged_element: 20;
+$zindex-popover: 45;
```
erosdome: Ok. In the future if e. g. popover is needed for the add step sidebar, we will add a new const `$zindex-add_step_sidebar_popover`

PR #219, opened 2017-11-20
Feature/code signing files with details

doomsayer13:


source/javascripts/controllers/_CertificateController.js.erb
```diff
@@ -18,8 +19,15 @@ angular.module("BitriseWorkflowEditor").controller("CertificateController", func
 	viewModel.certificateUploader;
 	viewModel.passwordEditedCertificate = null;
 	viewModel.certificateIsExposeGetterSetterConfigs;
+	viewModel.certificateSettings = [];
 	var passwordEditedCertificateOriginalPassword;
 
+	viewModel.selectedCertificate;
```
erosdome: Should be set to `null` here too.

PR #231, opened 2017-12-19
code sign tab - expiry status of identities (ProvProfile popup)

doomsayer13:


source/templates/code_signing.slim
```diff
@@ -96,13 +96,16 @@
...
 									.icon-wrapper.missing-icon-wrapper = svg("triangle")
 									.icon-wrapper.ok-icon-wrapper = svg("check-in-circle")
 								.details
-									.common-name[ng-bind="identity.commonName"]
-									.serial[ng-bind="identity.serial"]
+									.column
+										.common-name[ng-bind="identity.commonName"]
+										.serial[ng-bind="identity.serial"]
+									.column
+										.is-identity-expired[ng-if="provProfileCtrl.isIdentityExpired(identity.endDate)"] == data[:strings][:code_signing][:prov_profile][:prov_profile_details_popup][:identities][:expired]
```
erosdome: Simply call the CSS class "expired". It is not an element which says something like "is it expired?", this says "expired".

source/javascripts/controllers/_ProvProfileController.js.erb
```diff
@@ -152,6 +152,10 @@ angular.module("BitriseWorkflowEditor").controller("ProvProfileController", func
 		return _.any(certificateService.certificates, function(aCertificate) {
 			return _.any(aCertificate.details, { serial: identitySerial });
 		});
+	}
+
+	viewModel.isIdentityExpired = function(identityExpiryDateTime) {
+		return new Date(identityExpiryDateTime) < new Date();
```
erosdome: `new Date(identityExpiryDateTime)` -> simply `identityExpiryDateTime`, since it is already a Date type object.

source/javascripts/controllers/_ProvProfileController.js.erb
```diff
@@ -152,6 +152,10 @@ angular.module("BitriseWorkflowEditor").controller("ProvProfileController", func
 		return _.any(certificateService.certificates, function(aCertificate) {
 			return _.any(aCertificate.details, { serial: identitySerial });
 		});
+	}
+
+	viewModel.isIdentityExpired = function(identityExpiryDateTime) {
```
erosdome: Function name indicates that input argument will be identity. Either have it as an identity (and later on `identity.expireAt < new Date()`) or have a separate "is date expired" function in the date service, and use that function everywhere instead of having `viewModel.isIdentityExpired` defined.

PR #233, opened 2018-01-03
discard button fix on triggers page

doomsayer13:


source/javascripts/controllers/_TriggersController.js.erb
```diff
@@ -103,7 +103,9 @@ angular.module("BitriseWorkflowEditor").controller("TriggersController", functio
 
 	viewModel.deleteTrigger = function(trigger) {
 		var index = _.indexOf(viewModel.selectedTriggerType.triggers, trigger);
-		viewModel.selectedTriggerType.triggers.splice(index, 1);
+		if (index >= 0) {
```
erosdome: I would rather not do this, but instead in the `configureWithAppConfig` function, right after it is clearing all trigger types' triggers:

```
_.each($scope.triggerTypes, function(aTriggerType) {
	aTriggerType.triggers = [];
});

if (addTriggerProgress.isInProgress) {
	addTriggerProgress.reset();
}
```

Because `addTriggerProgress` should be consistent with the fact that a trigger adding is in progress. This fact is tied to the add-in-progress trigger being part of the current trigger type. Which stops to be true at the point when each trigger type's triggers are cleared, because the add-in-progress trigger is cleared as well.

PR #234, opened 2018-01-03
filtered empty key/value env vars and secrets

dbreuer:
Changes:
- filtered empty key/value pairs from Environment variables list after save event triggered
- filtered empty key/value pairs from Secrets list after save event triggered


source/javascripts/services/_appService.js.erb
```diff
@@ -147,6 +147,17 @@ angular.module("BitriseWorkflowEditor").service("appService", function($q, reque
 	appService.saveAppConfig = function(requestConfig) {
 		var variables = [];
 		if (appService.appConfig.app && appService.appConfig.app.envs) {
+			var variablesWithoutKeyAndValue = _.filter(_.map(appService.appConfig.app.envs, function(anEnvVarConfig, ind) {
```
doomsayer13: you can remove "ind" if it is not used

PR #236, opened 2018-01-04
allow * as initial character when entering trigger pattern

erosdome:


dbreuer:
Could we use empty field instead of the default *, so when the user start editing that then its not bothering the user input.

PR #238, opened 2018-01-05
delete all functionality for provisioning profile files and certificates

doomsayer13:
TODO before release


source/templates/code_signing.slim
```diff
@@ -24,28 +24,33 @@
 			h2.label.expose[ng-if="provProfileCtrl.provProfiles.length > 0"]
 				span.long-version == data[:strings][:code_signing][:prov_profile][:header][:expose_for_pull_requests]
 				span.short-version == data[:strings][:code_signing][:prov_profile][:header][:expose_for_pr]
+			.actions[ng-class="{'hidden': provProfileCtrl.provProfiles.length <= 1}" floating-menu toggle="provProfileCtrl.actionsForAllProvisioningProfiles.isMenuVisible"]
+				button.dropdown-toggle[ng-if="provProfileCtrl.provProfiles.length > 1" ng-click="provProfileCtrl.actionsForAllProvisioningProfiles.isMenuVisible = !provProfileCtrl.actionsForAllProvisioningProfiles.isMenuVisible" ng-disabled="provProfileCtrl.deleteAllProvProfilesProgress.isInProgress"] = svg("three-horizontal-dots")
```
erosdome: If there are no prov profiles, dropdown toggle will not be present, thus the dropdown will never be opened, thus actions will be an empty div? That's not cool.

source/templates/code_signing.slim
```diff
@@ -134,11 +139,16 @@
 			h2.label.expose[ng-if="certificateService.certificates.length > 0"]
 				span.long-version == data[:strings][:code_signing][:certificate][:header][:expose_for_pull_requests]
 				span.short-version == data[:strings][:code_signing][:certificate][:header][:expose_for_pr]
+			.actions[ng-class="{'hidden': certificateService.certificates.length <= 1}" floating-menu toggle="certificateCtrl.actionsForAllCertificates.isMenuVisible"]
```
erosdome: Same things for this block as above.

PR #241, opened 2018-01-16
Use both namespaced and non-namespaced handling for secrets' "is expose" flag - temporarily

erosdome:


source/javascripts/components/_Meta.js.erb
```diff
@@ -0,0 +1,65 @@
...
+	var Meta = function(userMetaConfig, defaultMetaConfig) {
+		this.userMetaConfig = userMetaConfig;
+		if (!this.userMetaConfig) {
+			this.userMetaConfig = {};
+		}
+
+		this.defaultMetaConfig = defaultMetaConfig;
+	};
+
+	Meta.prototype.valueGetterSetter = function(value, key, bundleID) {
```
gaborszakacs: Let's move the `value` to the last position, it's more convenient to use it in getter mode.

source/javascripts/components/_Meta.js.erb
```diff
@@ -0,0 +1,65 @@
...
+		if (value !== undefined) {
+			if (!this.userMetaConfig[bundleID]) {
+				this.userMetaConfig[bundleID] = {};
+			}
+			this.userMetaConfig[bundleID][key] = value;
+
+			// TODO: remove the following if block, it is only required during secrets' "is expose" migration
+			if (key == "is_expose" && bundleID == "bitrise.io.workflow-editor") {
```
gaborszakacs: I would prefer using the `bitrise.io` namespace defined in a constant.

source/javascripts/components/_Meta.js.erb
```diff
@@ -0,0 +1,65 @@
...
+				this.userMetaConfig.is_expose = value;
+			}
+
+			if (value === null) {
+				delete this.userMetaConfig[bundleID][key];
+			}
+			else if (this.defaultMetaConfig && this.defaultMetaConfig[bundleID] && this.defaultMetaConfig[bundleID][key] !== undefined && angular.equals(this.defaultMetaConfig[bundleID][key], value)) {
+				delete this.userMetaConfig[bundleID][key];
```
gaborszakacs: Use an `OR` instead of an `else if` you want to perform the same action.

PR #243, opened 2018-01-16
Fixes menu zindex while scrolling

pedrommone:
# Now

error

# With the fix

fixed


viktorbenei:
Awesome, thanks for the PR @pedrommone ! :)

@erosdome pls review! ;)

source/stylesheets/_constants.scss
```diff
@@ -32,7 +32,7 @@ $zindex-step_description_open: 0;
 $zindex-file_uploader_input: 10;
 $zindex-dragged_element: 20;
 $zindex-popover: 45;
-$zindex-sticky_header: 50;
+$zindex-sticky_header: 76;
```
erosdome: Please move the z-index a few lines below, so they are in ascending order.

pedrommone:
@erosdome first time a I see something like this, but this is totally fair. Fixed!

PR #250, opened 2018-01-30
update matching certificates popup

doomsayer13:


data/strings.yml
```diff
@@ -159,13 +159,17 @@ code_signing:
         devices: Devices
         capabilities: Capabilities
       device_registration_details:
-          not_registered: Not registered on bitrise
-          registered: Registered on Bitrise by
+        not_registered: Not registered on bitrise
+        registered: Registered on Bitrise by
+      matching_certificates_details:
+        missing_or_expired_notice: "There is no matching certificate right now. They are either missing or expired.<br><strong>Upload at least one certificate that is matching with one of your provisioning profile.</strong></br>"
```
erosdome: `<br>` does not need a closing tag. Also, I would separate these to `missing_or_expired_notice_part_1` and `missing_or_expired_notice_part_2` and put them in separate paragraphs in the template.

source/javascripts/controllers/_ProvProfileController.js.erb
```diff
@@ -147,6 +147,12 @@ angular.module("BitriseWorkflowEditor").controller("ProvProfileController", func
 		});
 	}
 
+	viewModel.hasValidIdentity = function() {
+		return !_.isUndefined(_.find(viewModel.selectedProvProfile.developerCertificateIdentities, function(anIdentity) {
```
erosdome: Simpler: use `_.any(...` instead of `_.isUndefined(_.find(...`.

source/stylesheets/_tab-code_signing.scss.erb
```diff
@@ -1142,6 +1155,7 @@
 
 						.udid {
 							word-break: break-all;
+							word-wrap: break-word;
```
erosdome: This way if there is a too long UDID (and on mobile, all of them will be too long), it won't be broken down to multiple lines, since it does not have white space. Is that what we want?

source/stylesheets/_tab-code_signing.scss.erb
```diff
@@ -1194,6 +1208,7 @@
 
 		@include media-below-mid {
 			display: block;
+			overflow-y: auto;
```
erosdome: This won't make `.content` scrollable, isn't that a problem?

source/templates/code_signing.slim
```diff
@@ -100,24 +100,27 @@
...
-									.icon-wrapper.missing-icon-wrapper = svg("triangle")
-									.icon-wrapper.ok-icon-wrapper = svg("check-in-circle")
+									.icon-wrapper.missing-icon-wrapper = svg("warning-in-circle")
+									.icon-wrapper.ok-icon-wrapper = svg("tick-in-circle")
 								.details
 									.column
 										.common-name[ng-bind="identity.commonName"]
 										.serial[ng-bind="identity.serial"]
 									.column
+										.included[ng-if="!provProfileCtrl.isIdentityExpired(identity) && provProfileCtrl.identityStatusCSSclass(identity) == 'ok'"] == data[:strings][:code_signing][:prov_profile][:prov_profile_details_popup][:identities][:included]
```
erosdome: `.identityStatusCSSclass()` and other functions that return CSS classes should not be used for condition checks like an `ng-if`, only for `ng-class`. Use `isIdentityWithSerialAddedToApp()` instead, you will have to make it a `viewModel.` function.

source/stylesheets/_tab-code_signing.scss.erb
doomsayer13: The issue appears only in firefox's quantum, there we should add this line, I found this advice somewhere on stackoverflow. :D

source/stylesheets/_tab-code_signing.scss.erb
doomsayer13: this made the content scrollable...

data/strings.yml
doomsayer13: There is no <b>

source/stylesheets/_tab-code_signing.scss.erb
dbreuer: I tested with Firefox 58.0.2 and did not worked. 
I suggest to use height if you have a flex element with overflow scroll.
for example `.vertical-scroll { height: inherit; }` and
`.popup .popup-body.snowflake article { max-height: calc(100% - $article-header-size); }` where the `$article-header-size = 36px;` as the header height - article padding `52 - 16`

doomsayer13: I also checked and the issue is that the `overflow-y: auto` only added for the small screen size(`@include media-below-mid`).
screen shot 2018-02-13 at 14 58 33
screen shot 2018-02-13 at 14 58 44


So, I added a class(`.vertical-scroll`) to the popup's article tag, which is doing the same thing now is everything is OK, but please test it again.

data/strings.yml
```diff
@@ -156,17 +156,21 @@ code_signing:
...
+        matching_certificates: Included Certificates
         devices: Devices
         capabilities: Capabilities
       device_registration_details:
-          not_registered: Not registered on bitrise
-          registered: Registered on Bitrise by
+        not_registered: Not registered on Bitrise
+        registered: Registered on Bitrise by
+      matching_certificates_details:
+        missing_or_expired_notice: "There is no matching certificate right now. They are either missing or expired.<br><strong>Upload at least one certificate that is matching with one of your provisioning profile.</strong>"
```
erosdome: I would separate these to `missing_or_expired_notice_part_1` and `missing_or_expired_notice_part_2` and put them in separate paragraphs in the template.

source/templates/code_signing.slim
```diff
@@ -106,24 +106,28 @@
...
+									.icon-wrapper.missing-icon-wrapper = svg("dot")
+									.icon-wrapper.expired-icon-wrapper = svg("warning-in-circle")
+									.icon-wrapper.ok-icon-wrapper = svg("tick-in-circle")
 								.details
 									.column
 										.common-name[ng-bind="identity.commonName"]
 										.serial[ng-bind="identity.serial"]
 									.column
-										.expired[ng-if="provProfileCtrl.isIdentityExpired(identity)"] == data[:strings][:code_signing][:prov_profile][:prov_profile_details_popup][:identities][:expired]
+										.uploaded[ng-if="provProfileCtrl.identityStatusCSSclass(identity) == 'ok'"] == data[:strings][:code_signing][:prov_profile][:prov_profile_details_popup][:identities][:uploaded]
```
erosdome: `.identityStatusCSSclass()` and other functions that return CSS classes should not be used for condition checks like an `ng-if`, only for `ng-class`. Use `isIdentityWithSerialAddedToApp()` instead, you will have to make it a `viewModel.` function.

source/templates/code_signing.slim
```diff
@@ -106,24 +106,30 @@
 						.menu.small
 							select.menu-selector[ng-model="provProfileCtrl.selectedMenu" ng-model-options="{getterSetter: true}" ng-options="menu.title for menu in provProfileCtrl.menus" ng-class="{ 'selected' : menu == provProfileCtrl.selectedMenu }"]
 					.content.vertical-scroll[ng-switch="provProfileCtrl.selectedMenu.id"]
+						.error-notice[ng-if="!provProfileCtrl.hasValidIdentity()" ng-switch-when="matching-certificates"]
+							div == data[:strings][:code_signing][:prov_profile][:prov_profile_details_popup][:matching_certificates_details][:missing_or_expired_notice_part_1]
```
erosdome: Two sentences - should either be `span` and `strong`, or plain text and `strong` or two `p`s, one of them having a `strong` inside, holding the text.

source/templates/code_signing.slim
```diff
@@ -106,24 +106,30 @@
...
+									.icon-wrapper.expired-icon-wrapper = svg("warning-in-circle")
+									.icon-wrapper.ok-icon-wrapper = svg("tick-in-circle")
 								.details
 									.column
 										.common-name[ng-bind="identity.commonName"]
 										.serial[ng-bind="identity.serial"]
 									.column
-										.expired[ng-if="provProfileCtrl.isIdentityExpired(identity)"] == data[:strings][:code_signing][:prov_profile][:prov_profile_details_popup][:identities][:expired]
+										.uploaded[ng-if="provProfileCtrl.identityStatus(identity) == 'ok'"] == data[:strings][:code_signing][:prov_profile][:prov_profile_details_popup][:identities][:uploaded]
+										.expired[ng-if="provProfileCtrl.isIdentityExpired(identity) == 'expired'"] == data[:strings][:code_signing][:prov_profile][:prov_profile_details_popup][:identities][:expired]
```
erosdome: `provProfileCtrl.isIdentityExpired`'s return value is a boolean - true or false, as its name indicates it. This condition will never be true.

Issue #244, opened 2018-01-18
Add new workflow, then discard -> Most of the workflow manage buttons disappear


erosdome:
As if there were no workflows, only the add button remains visible.

erosdome:
fixed in new release:

https://discuss.bitrise.io/t/workflow-editor-v1-1-12-released/4268

PR #254, opened 2018-02-06
Fix disappearing workflow manage buttons

doomsayer13:
Issue report: 
https://github.com/bitrise-io/bitrise-workflow-editor/issues/244


erosdome:
Closing this, as it may work but is only a workaround. Real issue lies in the selection caching, which is fixed here: https://github.com/bitrise-io/bitrise-workflow-editor/pull/259

PR #255, opened 2018-02-06
Handle protected state of app files

doomsayer13:


doomsayer13:
2. Certificate: screen shot 2018-02-08 at 12 44 49

PR #273, opened 2018-03-20
Add and show den_tags input field if "self hosted agent" stack is selected

doomsayer13:


source/javascripts/services/_appService.js.erb
```diff
@@ -331,13 +334,18 @@ angular.module("BitriseWorkflowEditor").service("appService", function($q, reque
 	}
 
 	appService.stackHasUnsavedChanges = function() {
-		return !angular.equals(appService.stack, appService.savedStack) || appService.dockerImage !== appService.savedDockerImage;
+		return !angular.equals(appService.stack, appService.savedStack) || appService.dockerImage !== appService.savedDockerImage || appService.denTags !== appService.savedDenTags;
 	};
 
 	appService.saveStack = function() {
-		return requestService.saveStack(appService.stack.id, appService.dockerImage).then(function() {
+		var bitriseDENtagsArray = appService.denTags ? _.map(appService.denTags.split(","), function(string) {
```
erosdome: It would be better if it was `undefined` instead of `[]` in the "else" case.

source/javascripts/services/_requestService.js.erb
```diff
@@ -542,6 +543,9 @@ angular.module("BitriseWorkflowEditor").service("requestService", function($q, $
 				if (dockerImage !== null) {
 					requestData.docker_image_name = dockerImage;
 				}
+				if (denTags !== null) {
```
erosdome: You can pass it without the `if` instead. In the case of `dockerImage` the null is a special value, meaning the default docker image. So it was needed to not be sent in that case, thus the if. There is no such thing for den tags, and also with the suggested change above, it will be assigend with undefined for cases when den tag setting is not an available feature, which is fine.

PR #286, opened 2018-05-08
Fix duplicate matches in steplist

doomsayer13:


source/javascripts/services/_stepSourceService.js.erb
```diff
@@ -283,7 +283,15 @@ angular.module("BitriseWorkflowEditor").factory("stepSourceService", function($q
 			}
 
 			return requestService.libraryFetch(notLoadedLibraryURLs, requestConfig).then(function(data) {
-				_.each(data, function(aLibraryData, aLibraryURL) {
+				var newLibraries = _.pick(data, function(aLibraryData, aLibraryURL) {
```
erosdome: We already do this check above at line:273. `notLoadedLibraryURLs` also indicates this: libraries which have not been loaded yet.

If this code change solved any issue, it might have been due to race condition between two calls of this same function. But the solution to that should not be doing the same thing twice.

PR #287, opened 2018-06-07
UX revision of android keystore file's input fields

doomsayer13:


source/templates/code_signing.slim
```diff
@@ -302,27 +302,27 @@
 							p
 								span> == data[:strings][:code_signing][:android_keystore][:keystore_password_as]
 								code.env-var[ng-bind="KeystoreFile.androidKeystorePasswordEnvVarKey | prettifiedKey"]
-							input.env-var[type="text" ng-if="androidKeystoreCtrl.showPassword" name="password" required ng-model="androidKeystoreCtrl.keystoreFile.password" placeholder="#{ data[:strings][:code_signing][:android_keystore][:password_placeholder] }"]
+							input.env-var[type="text" ng-if="androidKeystoreCtrl.showPassword" name="password" ng-required="androidKeystoreCtrl.keystoreFile.password" ng-model="androidKeystoreCtrl.keystoreFile.password" placeholder="#{ data[:strings][:code_signing][:android_keystore][:password_placeholder] }"]
 							input.env-var[type="text" ng-if="!androidKeystoreCtrl.showPassword" ng-model="codeSigningCtrl.placeholderForText(androidKeystoreCtrl.keystoreFile.password)" ng-model-options="{getterSetter: true}" disabled]
-							.show-password
+							.show-password[ng-if="androidKeystoreCtrl.keystoreFile.password"]
```
erosdome: It should not be this way. It should be visible always. Whenever the file loads (or has been uploaded), toggle it on based on whether the input is empty or not. And whenever the fields have been saved, do the same toggling with the same condition.

source/javascripts/controllers/_AndroidKeystoreController.js.erb
```diff
@@ -74,6 +70,10 @@ angular.module("BitriseWorkflowEditor").controller("AndroidKeystoreController",
 				viewModel.keystoreFile = null;
 			}
 
+			viewModel.showPassword = viewModel.keystoreFile ? _.isUndefined(viewModel.keystoreFile.password) : false;
```
erosdome: These should only be done in the case when there is already a file uploaded. It is the block starting at line 46 `if (data !== null) {`.

For the case when there is no uploaded file yet, but user later uploads one, do this same thing in the upload method's success block.

Also, do this same thing after saving the metadata - see other comment below.

And it does not need to be checked whether it is undefined or not, simply `!viewModel.keystoreFile.password` should be enough, so it handles empty strings as well.

PR #296, opened 2018-07-04
Fix secrets reveal button

doomsayer13:


doomsayer13:
screen shot 2018-07-04 at 10 28 09
screen shot 2018-07-04 at 10 28 28
screen shot 2018-07-04 at 10 28 47

Chrome, Firefox and Safari

PR #310, opened 2018-07-30
Automate release

erosdome:


bitrise.yml
```diff
@@ -225,9 +225,84 @@ workflows:
   # ----------------------------------------------------------------
   # --- workflows for Releasing
   create-release:
+    title: Create release
+    before_run:
+    - create-binaries
```
erosdome: Not cool, we need to bump version number prior to creating binaries.

PR #318, opened 2018-08-21
Discard button fix for yaml editor

dbreuer:
Change:
- Fixing YAML editor discard change to revert YAML to the saved version.


source/javascripts/controllers/_YMLController.js.erb
```diff
@@ -8,6 +8,15 @@ angular.module("BitriseWorkflowEditor").controller("YMLController", function($sc
 	var editor;
 	viewModel.downloadAppConfigYMLPath = requestService.mode == "website" ? requestService.appConfigYMLDownloadPath() : null;
 
+	// discard change detection
```
erosdome: Code should not contain comments unless it is absolutely necessary for some reason. Here it is quite self-explanatory what this watcher does.

PR #319, opened 2018-08-21
Add option to remove sensitive input value

dbreuer:
change:
- Add option to remove sensitive input value


source/stylesheets/_rebo.scss
```diff
@@ -194,3 +194,22 @@ a.rebo {
 		@include size(8px);
 	}
 }
+
+.rebo.inline {
```
erosdome: Move it to another file and call it another class, this class is reserved for that one specific button style we use in most of the cases.

PR #349, opened 2018-11-14
Replace colors with corresponding constants #9tv6f

erosdome:


erosdome:
Good question, I just chose it because it had very little difference, but now I also checked it with design team and they said it is okay.

PR #370, opened 2019-04-29
Fix GitHub spelling

friederbluemle:
Just a minor typo fix: `Github` -> `GitHub`.


sylank:
@friederbluemle 

First of all, thank you for your contribution, it is really important for us. After some review, we would not like to approve this PR because these changes are present in 3rd party libraries. 

After all, nice catch and don't give up. If you would l to contribute check the open issues and help to close them or find new releases from the libraries and send a new PR.

Thank you 😺

friederbluemle:
Makes sense, thanks @sylank 👌

PR #371, opened 2019-05-13
Saving bitrise.yml: display warnings #a3k79

sylank:
Saving bitrise.yml: display warnings to the user that was given by the CLI

ClickUp card link: https://app.clickup.com/t/a3k79

TODO before release:

- [ ] Deploy this project first then the backend

### DIFF
--- a/apiserver/utility/utility.go
+++ b/apiserver/utility/utility.go
@@ -12,12 +12,18 @@ import (
 
 // ValidationResponse ...
 type ValidationResponse struct {
-	Warnings map[string][]string `json:"warnings"`
+	Warnings *WarningItems `json:"warnings"`
+}
+
+// WarningItems ...
+type WarningItems struct {
+	Config  []string `json:"config"`
+	Secrets []string `json:"secrets"`
 }
 
 // ValidateBitriseConfigAndSecret ...
 // `bitriseConfig` and `secretsConfig` can be either YML or JSON, both are accepted
-func ValidateBitriseConfigAndSecret(bitriseConfig, secretsConfig string) (map[string][]string, error) {
+func ValidateBitriseConfigAndSecret(bitriseConfig, secretsConfig string) (*WarningItems, error) {
 	bitriseConfigBase64 := base64.StdEncoding.EncodeToString([]byte(bitriseConfig))
 	secretsConfigBase64 := base64.StdEncoding.EncodeToString([]byte(secretsConfig))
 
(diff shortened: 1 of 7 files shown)


apiserver/utility/utility.go
```diff
@@ -53,18 +59,25 @@ func ValidateBitriseConfigAndSecret(bitriseConfig, secretsConfig string) error {
...
-		return fmt.Errorf("bitrise validation command failed, error: %s", validationOutput.Error)
+		return nil, fmt.Errorf("bitrise validation command failed, error: %s", validationOutput.Error)
 	}
 
 	if cmdErr != nil {
-		return fmt.Errorf("bitrise validation command failed, error: %s", cmdErr)
+		return nil, fmt.Errorf("bitrise validation command failed, error: %s", cmdErr)
+	}
+
+	warningsArray := []string{}
```
gaborszakacs: Pls don't include the type in the variable name.

apiserver/utility/utility.go
```diff
@@ -53,18 +59,25 @@ func ValidateBitriseConfigAndSecret(bitriseConfig, secretsConfig string) error {
...
+	warnings := validationOutput.Data.Config.Warnings
+	warnings = append(warnings, validationOutput.Data.Secrets.Warnings...)
```
gaborszakacs: So we are putting both secret and config related warnings into one slice? I would consider using a more structured format at that point (eg. by creating a custom type), the utility shouldn't be aware how others will use its output, there can be a use case of being able to differentiate between secret and config warnings. (I'm not sure if our use case is one of those, for example.)

apiserver/utility/utility.go
```diff
@@ -53,18 +59,25 @@ func ValidateBitriseConfigAndSecret(bitriseConfig, secretsConfig string) error {
...
+	if warnings != nil && len(warnings) > 0 {
+		warningsArray = append(warningsArray, warnings...)
 	}
 
-	return nil
+	return warningsArray, nil
```
gaborszakacs: Pls add tests to showcase this new feature.

apiserver/utility/utility.go
```diff
@@ -12,12 +12,12 @@ import (
 
 // ValidationResponse ...
 type ValidationResponse struct {
-	Warnings []string `json:"warnings"`
+	Warnings map[string][]string `json:"warnings"`
```
gaborszakacs: Let's make this into a type.

PR #379, opened 2019-06-11
(FE) Saving bitrise.yml: display warnings to the user that was given by the CLI #a3k5r

sylank:
Using free text in sensitive inputs will result in a warning when the user saves the bitrise.yml. We might want to notify the user about this. It will be displayed in the top of the build log anyway, so the user will be notified anyway, if we don't do this.

ClickUp link: https://app.clickup.com/t/a3k5r


source/templates/yml.slim
```diff
@@ -1,4 +1,10 @@
 .tab-yml.menu-content-padding[ng-controller="YMLController as ymlCtrl"]
+	section.cardo.warn[ng-if="appService.validationWarnings.length!=0"]
+		.icon
+		.message
+			.title == data[:strings][:yml][:warn][:title]
+			ul
+				li[ng-repeat="warning in appService.validationWarnings"] {{warning}}
```
doomsayer13: please change it to: li[ng-repeat="warning in appService.validationWarnings" ng-bind="warning"]

PR #380, opened 2019-06-12
Allow update of protected Secret values

doomsayer13:


source/javascripts/controllers/_SecretsController.js.erb
```diff
@@ -168,7 +168,49 @@ angular.module("BitriseWorkflowEditor").controller("SecretsController", function
...
+								return secret.value();
+							}
+						}
+					}
+					else {
+						if (secret.customProperties.isEditMode === true) {
+							return secret.value(value);
+						}
+						else {
+							return secret.value(value);
```
nyikos-zoltan: These two branches are the same?

PR #384, opened 2019-09-02
Tool 709 review workflow editor validation

lszucs:
Bitrise CLI calls were used to validate models. Validation involved calling the binary and processing the output string.

The task was to refactor the validation of bitrise.yml and model validation to be consistent and to not use calls to the Bitrise CLI binary.

Direct calls to the underlying validation functions of the Bitrise CLI had been introduced.

Other than that, the Bitrise CLI will not issue an error if an empty secrets file is given. In such cases, the output will simply omit the secrets key.

```
$ bitrise -l=panic validate --format=json --config-base64 Zm9ybWF0X3ZlcnNpb246IDEuMy4w --inventory-base64 ""
$ {"data":{"config":{"is_valid":true}}}
```

Because the Workflow Editor would not accept empty secret files, this was previously solved by depending on Golang trickery using default values for missing fields in the Bitrise CLI JSON output.

Checking for empty secrets config has also been made more explicit.

Also note: had to update dependencies due to transitive conflicts.


apiserver/utility/utility.go
```diff
@@ -2,12 +2,13 @@ package utility
...
 	"encoding/base64"
-	"encoding/json"
+	// "encoding/json"
 	"fmt"
 	"os"
 	"strings"
 
-	"github.com/bitrise-io/go-utils/command"
+	"github.com/bitrise-io/bitrise/cli"
+	// "github.com/bitrise-io/go-utils/command"
```
trapacska: pls remove these commented out imports

apiserver/utility/utility.go
```diff
@@ -27,61 +28,34 @@ func ValidateBitriseConfigAndSecret(bitriseConfig, secretsConfig string) (*Warni
 	bitriseConfigBase64 := base64.StdEncoding.EncodeToString([]byte(bitriseConfig))
 	secretsConfigBase64 := base64.StdEncoding.EncodeToString([]byte(secretsConfig))
 
-	validateCmd := command.New("bitrise",
-		"-l=panic", "validate", "--format=json",
-		"--config-base64", bitriseConfigBase64,
-		"--inventory-base64", secretsConfigBase64)
+	fmt.Println("`" + secretsConfig + "`" + secretsConfigBase64 + "`")
```
trapacska: so, we are absolutely sure about this?

trapacska: and also about this?

lszucs: 🤦‍♂

apiserver/utility/utility.go
```diff
@@ -27,61 +26,31 @@ func ValidateBitriseConfigAndSecret(bitriseConfig, secretsConfig string) (*Warni
...
+	if bitriseErr != nil {
+		errorStrs = append(errorStrs, "Config validation error: "+bitriseErr.Error())
 	}
-	if !validationOutput.Data.Secrets.IsValid {
-		errorStrs = append(errorStrs, "Secret validation error: "+validationOutput.Data.Secrets.Error)
+	if secretsErr != nil {
+		errorStrs = append(errorStrs, "Secret validation error: "+secretsErr.Error())
+	} else {
+		if secretsConfigBase64 == "" {
+			errorStrs = append(errorStrs, "Secret validation error: Validation failed: Empty secrets configuration")
```
godrei: In the PR comments you state: `Other than that, the Bitrise CLI will not issue an error if an empty secrets file is given. In such cases, the output will simply omit the secrets key.`

Than why do we raise this error?

apiserver/utility/utility.go
```diff
@@ -27,61 +26,31 @@ func ValidateBitriseConfigAndSecret(bitriseConfig, secretsConfig string) (*Warni
...
-
 	warningItems := &WarningItems{Config: []string{}, Secrets: []string{}}
-	if len(validationOutput.Data.Config.Warnings) > 0 {
-		warningItems.Config = validationOutput.Data.Config.Warnings
+	if len(bitriseWarns) > 0 {
+		warningItems.Config = bitriseWarns
 	}
-	if len(validationOutput.Data.Config.Warnings) > 0 {
-		warningItems.Secrets = validationOutput.Data.Secrets.Warnings
+	if len(bitriseWarns) > 0 {
```
godrei: i think in case of secret validation we do not throw warnings, other than that, we do you add the bitrise.yml warnings to the secret validation item?

apiserver/utility/utility.go
lszucs: Sorry, the phrasing wasn't clear. 

Although the CLI does not raise an error, because the output does not contain key in the validation result JSON, `validationOutput.Data.Secrets.IsValid` will be `false` and so the Workflow Editor will eventually treat this as an error case.

Related test case: https://github.com/bitrise-io/bitrise-workflow-editor/blob/master/apiserver/utility/utility_test.go#L117

apiserver/utility/utility.go
lszucs: 🤦‍♂ Assigning `bitriseWarns` to `warningItems.Secrets` was a mistake.

apiserver/utility/utility.go
```diff
@@ -27,61 +26,31 @@ func ValidateBitriseConfigAndSecret(bitriseConfig, secretsConfig string) (*Warni
...
-
-	var validationOutput BitriseCLIValidateOutputModel
-	if outputParseErr := json.NewDecoder(strings.NewReader(combinedOut)).Decode(&validationOutput); outputParseErr != nil {
-		if cmdErr != nil {
-			return nil, fmt.Errorf("Failed to run bitrise validate command, error: %s | 'bitrise validate' command output: %s", cmdErr, combinedOut)
-		}
-		return nil, fmt.Errorf("Failed to parse bitrise validate output, error: %s | 'bitrise validate' command output: %s", outputParseErr, combinedOut)
-	}
+	_, bitriseWarns, bitriseErr := cli.CreateBitriseConfigFromCLIParams(bitriseConfigBase64, "")
+	_, secretsErr := cli.CreateInventoryFromCLIParams(secretsConfigBase64, "")
```
godrei: this should be only called if the secrets base 64 is not empty otherwise, it can pick up the secrets file in the current working directory, which should not happen: https://github.com/bitrise-io/bitrise/blob/518e23eca82dc826ef7ee1e27e15878dd8118bb3/cli/run_util.go#L252

apiserver/utility/utility.go
```diff
@@ -27,61 +26,34 @@ func ValidateBitriseConfigAndSecret(bitriseConfig, secretsConfig string) (*Warni
...
-
 	warningItems := &WarningItems{Config: []string{}, Secrets: []string{}}
-	if len(validationOutput.Data.Config.Warnings) > 0 {
-		warningItems.Config = validationOutput.Data.Config.Warnings
+	if len(bitriseWarns) > 0 {
+		warningItems.Config = bitriseWarns
 	}
-	if len(validationOutput.Data.Config.Warnings) > 0 {
-		warningItems.Secrets = validationOutput.Data.Secrets.Warnings
+	if secretsErr != nil {
```
godrei: you can remove this case: secret validation will never raise warnings.
other than that, the secret error is stored in `errorStrs` already.

PR #410, opened 2020-01-14
[WEB-2963] Update Verified and Official step badges

simonmarton:
The steps that were `verified` previously are now labeled `official`.

Steps that are maintained by the community and are approved by the Bitrise team will be labeled `verified`.

----
**Screenshots**

Editing a WF:
Screenshot 2020-01-14 at 18 11 28

Step picker:
Screenshot 2020-01-14 at 18 12 04


gergoovari:
A few alignments would be handy, let's discuss it.

source/javascripts/services/_stepSourceService.js.erb
```diff
@@ -163,17 +166,17 @@ angular.module("BitriseWorkflowEditor").factory("stepSourceService", function($q
...
+		// 	var step;
+		// 	try {
+		// 		step = stepSourceService.stepFromCVS(aCVS);
+		// 	}
+		// 	catch (error) {
+		// 		return false;
+		// 	}
+
+		// 	return step.libraryURL !== undefined;
+		// });
```
tadam313: can we remove this?

source/javascripts/services/_stepSourceService.js.erb
```diff
@@ -283,6 +286,7 @@ angular.module("BitriseWorkflowEditor").factory("stepSourceService", function($q
 			}
 
 			return requestService.libraryFetch(notLoadedLibraryURLs, requestConfig).then(function(data) {
+				console.log('libraryFetch 1');
```
tadam313: let's remove console logs

source/javascripts/components/_AddStep.js.erb
```diff
@@ -127,6 +127,7 @@ angular.module("BitriseWorkflowEditor").service("addStepService", function() {
 						return aLibrary.steps[stepID][aStepVersion].isDeprecated();
 					}), function(aStepVersion, stepID) {
 						var step = angular.copy(aLibrary.steps[stepID][aStepVersion]);
+						step.info = angular.copy(aLibrary.steps[stepID].info)
```
erosdome: If I understand it correctly, this is not a changeable property of a step. So it is okay that by reference it is the same object for two step instances since we won't be changing any of them. The copy one line above is necessary because it contains changeable properties.
```suggestion
						step.info = aLibrary.steps[stepID].info;
```

source/javascripts/services/_stepSourceService.js.erb
erosdome: I guess we can, it is not used anywhere.

PR #411, opened 2020-01-15
[WEB-3334] version strings extraction

tadam313:
_STATUS_: user can choose wildcard versions on the UI we set it on the left side step control and also write it to the yml. All the ancillary tasks (highlight step updates on the UI etc) will come in separate tasks. 

**Contains:** 
* created semver service for version related logics
* created angular filter to simplify UI data binding
*  indentation, tests


source/javascripts/services/_stepSourceService.js.erb
```diff
@@ -1,335 +1,328 @@
...
+                    if (!library) {
+                        break;
+                    }
+
+                    if (!library.steps[step.id]) {
+                        throw new Error("<%= data[:strings][:step_source_service][:step_from_cvs][:no_step_with_id_error_prefix] %>" + step.id);
+                    }
+
+                    var normalizedVersion = semverService.normalizeVersion(step.version);
+                    var requestedVersion = semverService.resolveVersion(normalizedVersion, step.id, library);
```
tadam313: resolving version against the library

source/javascripts/services/_stepSourceService.js.erb
```diff
@@ -1,335 +1,328 @@
...
+
+                    if (!library.steps[step.id][requestedVersion]) {
+                        throw new Error("<%= data[:strings][:step_source_service][:step_from_cvs][:no_step_with_version_error_prefix] %>"+ requestedVersion);
+                    }
+
+                    step.version = normalizedVersion;
```
tadam313: this is bound to the UI so need to use the normalized version (wildcards)

source/javascripts/services/_stepSourceService.js.erb
```diff
@@ -1,335 +1,328 @@
...
+
+            if (version !== null) {
+                step.cvs += "@" + semverService.removeWildcardVersionParts(version);
+                step.version = version;
+            }
+            else {
+                step.version = library.latestStepVersions[step.id];
+            }
+
+            var stepVersion = semverService.resolveVersion(step.version, step.id, library);
```
tadam313: for the configuration we need to use the resolved version of the step

source/templates/workflows.slim
```diff
@@ -95,7 +95,7 @@
 							section.version[ng-if="workflowsCtrl.selectedStep.version !== undefined"]
 								span.icon[ng-class="{'latest': stepSourceService.isLatestStepVersion(workflowsCtrl.selectedStep)}"]
 								label[for="selected-step-version-select"] == data[:strings][:workflows][:steps][:version]
-								select#selected-step-version-select[ng-options="(version ? version : '#{ data[:strings][:workflows][:steps][:always_latest] }') for version in (workflowsCtrl.selectedStep.libraryURL !== undefined ? [null] : []).concat(stepSourceService.versionsOfStep(workflowsCtrl.selectedStep))" ng-model="workflowsCtrl.requestedStepVersion"]
+								select#selected-step-version-select[ng-options="(version ? version : '#{ data[:strings][:workflows][:steps][:always_latest] }') for version in (workflowsCtrl.selectedStep | wildcardVersions)" ng-model="workflowsCtrl.requestedStepVersion"]
```
tadam313: simplified the logic with a filter

package.json
```diff
@@ -4,7 +4,8 @@
   "description": "Bitrise workflow editor",
   "main": "index.js",
   "scripts": {
-    "test": "MODE=WEBSITE bundle exec middleman build && karma start spec/karma.conf.js",
+    "karma": "karma start spec/karma.conf.js",
```
tadam313: if you don't need middleman compilation, it's faster

spec/javascripts/semverService.spec.js
```diff
@@ -0,0 +1,102 @@
+describe("semverService", () => {
```
tadam313: **NOTE**: in tests we could use ES6 systax as JSDOM supports it

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -277,8 +277,6 @@ angular.module("BitriseWorkflowEditor").controller("MainController", function($s
 		if (originalOnbeforeunloadHandler) {
 			originalOnbeforeunloadHandler(event);
 		}
-
-		window.location.replace();
```
tadam313: @erosdome `location.replace(url)` expects an url to be passed. Everytime I refreshed the browser I got an error, so I removed this I guess it's ok right? If the purpose of this was to trigger the reload: that happens regardless we don't need to do it explicitly

source/javascripts/main.js
```diff
@@ -6,3 +6,7 @@
 //= require_tree ./components/
 //= require_tree ./directives/
 //= require_tree ./controllers/
+
+_.templateSettings = {
+    interpolate: /\{\{(.+?)\}\}/g
+};
```
tadam313: Original underscore template syntax clashed with ruby on rails one :)

spec/karma.conf.js
```diff
@@ -22,6 +22,7 @@ module.exports = (config) => {
     // list of files / patterns to load in the browser
     files: [
       'javascripts/angular_cache.js',
+      'javascripts/vendor.js',
```
tadam313: it contains underscore dependency, so had to list it here explicitly to load before the main app code.

source/javascripts/controllers/_StackController.js.erb
```diff
@@ -1,135 +1,135 @@
 (function() {
```
tadam313: only indentation updates in this file

source/javascripts/components/_Step.js.erb
```diff
@@ -146,6 +146,14 @@ angular.module("BitriseWorkflowEditor").factory("Step", function($injector, Vari
 		return (this.info || {}).maintainer === 'community';
 	};
 
+	Step.prototype.isLocal = function() {
+		return !!this.localPath;
+	};
+
+	Step.prototype.isLibraryStep = function() {
+		return !!this.libraryURL;
+	};
```
tadam313: convenience methods here

source/javascripts/services/_stepSourceService.js.erb
```diff
@@ -1,143 +1,140 @@
...
-				step.defaultStepConfig = library.steps[step.id][requestedVersion].defaultStepConfig;
-		}
+			return step;
+		};
 
-		return step;
-	};
+		stepSourceService.versionsOfStep = function(step) {
+			if (!step.isLibraryStep()) {
+				return step.isLocal() ? null : [step.version];
```
erosdome: The code at some places counts on step version being undefined instead of null here:
* ~~stepSourceService: changeStepToVersion~~ nvm I see you update it in this PR
* workflows.slim: step list, version selector

Either use the orginal undefined in this case, or update the above mentioned code.

source/javascripts/services/_semverService.js.erb
```diff
@@ -0,0 +1,99 @@
...
+    var isVersionCompatible = function(wildcard, version) {
+        var semverParts = version.split('.');
+        var wildcardParts = wildcard.split('.');
+
+        var matchingVersionPart = true;
+        for (var i = 0; matchingVersionPart && i < semverParts.length; i++) {
+            matchingVersionPart = semverParts[i] == wildcardParts[i] || wildcardParts[i] == WILDCARD;
+        }
+
+        return matchingVersionPart;
```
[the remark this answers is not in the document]
tadam313: 👍

source/javascripts/services/_stepSourceService.js.erb
tadam313: I'd rather use `null` instead of `undefined`. Reason is: `undefined` is given by a runtime to indicate missing value, although `null` is explicit. Also has some better support for equality checks etc...

source/javascripts/components/_Step.js.erb
```diff
@@ -146,6 +146,14 @@ angular.module("BitriseWorkflowEditor").factory("Step", function($injector, Vari
...
+
```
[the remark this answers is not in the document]
tadam313: will do

tadam313:
@erosdome I changed `undefined` check to simple falsy check which works both with `null` and `undefined`. Also spotted I forgot to resolve wildcard version at step latest version check, I did that as well

source/javascripts/services/_stepSourceService.js.erb
tadam313: @erosdome I've just tested the templates (we don't have e2e tests, so manually for now) and it seems `undefined` check is there for a good reason, since `null` means latest and that's ok, we should only hide sections when the version is `undefined`. I agree it's not super clear now, but let's say it's out of scope for this task.

PR #416, opened 2020-01-21
[WEB-3337] disable non wildcard versions in version selector

tadam313:
* additional helper filters
* removed explicit latest version insertion


source/javascripts/main.js
```diff
@@ -7,6 +7,12 @@
...
 
+// underscore template settings
 _.templateSettings = {
     interpolate: /\{\{(.+?)\}\}/g
 };
+
+// polyfill for IE
+if (!Number.MAX_SAFE_INTEGER) {
+    Number.MAX_SAFE_INTEGER = Math.pow(2, 53) - 1;
+}
```
tadam313: Based on this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER IE needs polyfill

source/javascripts/services/_semverService.js.erb
```diff
@@ -71,10 +84,8 @@ angular.module("BitriseWorkflowEditor").factory("semverService", function() {
             })
             .flatten()
             .tap(function(wVersions) {
-                if (step.version) {
-                    // adding step own version if any
-                    wVersions.push(step.version);
-                }
+                // adding step own version
+                wVersions.push(step.version);
```
tadam313: As discussed with Gergo we should stop adding explicit latest version AKA: `null` but we still need to display whether it's set to latest in yml. So had to add step version here regardless.

spec/javascripts/semverService.spec.js
```diff
@@ -88,7 +88,7 @@ describe("SemverService", () => {
             ]);
         });
 
-        it("should only add step version if not null", () => {
+        it("should add step version even if null to the beginning", () => {
```
simonmarton: 🤔 not sure I understand this test's intent

source/javascripts/main.js
tadam313: hmm yeah a browser support doc would be useful. I'll leave it here just in case

spec/javascripts/semverService.spec.js
simonmarton: Thanks, clear. A description update would be great!

PR #420, opened 2020-01-27
Skip API disconnection on browser close when running in development mode

erosdome:


source/javascripts/_apihandler.js.erb
```diff
@@ -1,11 +1,11 @@
 window.onload = function() {
-	if ("<%= mode %>" == "cli") {
+	if ("<%= mode %>" == "cli" && "<%= build? %>" == "true") {
```
tadam313: @erosdome a suggestion, should not assign a variable to indicate why we are doing this? In the source code it will look sth like this: "if('true' == 'true')". My suggestion:

```javascript
var shouldAbortConnTermination = <%= mode == "CLI" && build ==  %>;
if (shouldCancelApi) { ... }
```

source/javascripts/_apihandler.js.erb
```diff
@@ -1,11 +1,11 @@
...
 		cancelAPIConnectionClose();
 	}
 };
 
 window.onbeforeunload = function() {
-	if ("<%= mode %>" == "cli") {
+	if ("<%= mode %>" == "cli" && "<%= build? %>" == "true") {
```
tadam313: same herE:
```javascript
var shouldCloseConnection = <%= mode == "CLI" && build ==  %>;
if (shuldCloseConnection) { ... }
```

source/javascripts/_apihandler.js.erb
erosdome: Good point, I moved the conditions to an `isPluginMode` variable, somewhat different from your suggestion, but should indicate clearly what scenario it refers to.

source/javascripts/_apihandler.js.erb
```diff
@@ -1,11 +1,13 @@
+var isPluginMode = "<%= mode %>" == "cli" && "<%= build? %>" == "true";
```
tadam313: how about `var isPluginMode = <%= mode == "cli" && build %>`?

erosdome: Good point, my solution still results in ugly looking static code. With this one it looks ok.

PR #421, opened 2020-01-27
Lazy load add step sidebar icons [WEB-2014]

erosdome:


source/javascripts/components/_AddStep.js.erb
```diff
@@ -7,7 +7,19 @@ angular.module("BitriseWorkflowEditor").service("addStepService", function() {
 	var addStepService = {
 		isVisible: false,
 		stepSelected: undefined,
-		steps: undefined
+		steps: undefined,
+		stepIconLoadedStatesByStepIds: {},
+		stepIntersectionObserverCallbacks: {},
+		stepIntersectionObserver: new IntersectionObserver(function(entries) {
```
tadam313: @erosdome what is our browser support guideline? see this: https://caniuse.com/#search=IntersectionObserver it's not supported in many browsers like IE. Should we use a polyfill?

source/javascripts/components/_AddStep.js.erb
```diff
@@ -463,14 +461,32 @@ angular.module("BitriseWorkflowEditor").service("addStepService", function() {
 		}
 	};
 
-}).directive("addStepSidebarStep", function($animate) {
+}).directive("addStepSidebarStep", function($filter, addStepService, imageService) {
```
tadam313: can we add injection names? like `["$filter", "addStepService", "imageService", function($filter, addStepService, imageService) {...}]`? that's the only way if we want to minify it later etc, that's the best practice.

source/javascripts/components/_AddStep.js.erb
```diff
@@ -463,14 +461,32 @@ angular.module("BitriseWorkflowEditor").service("addStepService", function() {
...
 	return {
 		restrict: "C",
 		templateUrl: "<%= mode_dependant_asset_path(data[:routes][:templates][:add_step_sidebar_step] )%>",
 		scope: {
 			step: "=",
 			selected: "&selectCallback"
+		},
+		link: function(scope, element) {
+			scope.addStepService = addStepService;
+			addStepService.stepIntersectionObserverCallbacks[scope.step.id] = function() {
```
tadam313: should not we hide `stepIntersectionObserver` in the `addStepService`? We could expose an `observer` and `unobserve` method from the service, but how that does it only concerns the service not this component. So I'd recommend implement something like this: `addStepService.objserve(element[0], function callback() { .... })` and `stepService.unobserve(element[0])`;

source/templates/add_step_sidebar_step.slim
```diff
@@ -1,7 +1,7 @@
 .step
 	button.select.angular-animated[ng-click="selected()"] = svg("plus")
 	.step-content
-		.icon[ng-style="{'background-image': (step | normalizedStepIconURL | backgroundImagePropertyValue)}"]
+		.icon[ng-style="{'background-image': (addStepService.stepIconLoadedStatesByStepIds[step.id] ? (step | normalizedStepIconURL | backgroundImagePropertyValue) : null)}"]
```
tadam313: `addStepService.isIconLoaded(step.id) && (step | normalizedStepIconURL | backgroundImagePropertyValue)`

tadam313: and shouldn't we use a default image until the icon is loaded?

source/javascripts/components/_AddStep.js.erb
erosdome: I'd rather do it separately, we currently use this format for dependency injection everywhere in the Workflow Editor. In a separate PR we could collectively replace all the dependency injections to fix the issue. I created an internal maintenance task for it.

source/templates/add_step_sidebar_step.slim
erosdome: I did that in the stylesheet instead: when there is no background image defined, the `:before` element below will be exposed, which contains a loading spinner.

PR #428, opened 2020-01-31
[WEB-3338] step update notifier

tadam313:
* added icon assets
* created some utility service to check versioning conditions on the UI
* refactored templating with tests

Screenshot 2020-01-31 at 11 34 32

Screenshot 2020-01-31 at 11 34 20


source/images/step/latest.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+<circle cx="12" cy="12" r="10" fill="#0FC389"/>
+<path d="M7.57666 12.2071L9.06365 10.7929L10.9488 12.5858L14.7191 9L16.2061 10.4142L10.9488 15.4142L7.57666 12.2071Z" fill="white"/>
```
erosdome: Remove `fill` from these elements so we can
- specify the colors with our constants in the stylesheets
- reuse the same .svg files in other places, with different colors

source/images/step/upgrade.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+<circle cx="12" cy="12" r="10" fill="#FF2158"/>
+<path d="M13 11L15 13L16.5 11.5L12 7L7.5 11.5L9 13L11 11V17H13V11Z" fill="white"/>
```
erosdome: Same as above.

source/stylesheets/main.scss
```diff
@@ -382,3 +382,7 @@ select:hover, select:focus, .mak:hover, .mak:focus {
 .vertical-scroll {
 	overflow-y: auto;
 }
+
+[ng-click] {
+	cursor: pointer;
+}
```
erosdome: This might result in disabled buttons which have `ng-click` defined: they could get pointer cursor, incorrectly.

Define pointer cursor for the element you're aiming for specifically, in its stylesheet (`.icon.icon-danger`, though it would be better if it would also receive a class indicating its functionality).

source/templates/workflows.slim
```diff
@@ -95,9 +95,15 @@
 									button[ng-click="workflowsCtrl.shouldShowDescription = !workflowsCtrl.shouldShowDescription"]
 									= svg("arrow-down")
 							section.version[ng-if="workflowsCtrl.selectedStep.version !== undefined"]
-								span.icon[ng-class="{'latest': stepSourceService.isLatestStepVersion(workflowsCtrl.selectedStep)}"]
-								label[for="selected-step-version-select"] == data[:strings][:workflows][:steps][:version]
-								select#selected-step-version-select[ng-options="(version | versionLabel) disable when (version | isVersionDisabled) for version in (workflowsCtrl.selectedStep | stepVersions)" ng-model="workflowsCtrl.requestedStepVersion"]
+								.version-info
+									.resolved-version
+										span.icon.icon-ok[ng-show="stepSourceService.isLatestStepVersion(workflowsCtrl.selectedStep)"] = svg("step/latest")
+										span.icon.icon-danger[ng-hide="stepSourceService.isLatestStepVersion(workflowsCtrl.selectedStep)" ng-click="workflowsCtrl.updateStep(workflowsCtrl.selectedStep, workflowIndex)"] = svg("step/upgrade")
```
erosdome: Make it a `<button>` instead?

source/images/step/latest.svg
tadam313: I'd leave it here since it provides a reasonable default in case we forget to specify it in CSS. A wrong color is better than a broken UI.

source/stylesheets/main.scss
tadam313: good point, will change it

source/templates/workflows.slim
tadam313: aye

source/stylesheets/_default.scss
```diff
@@ -695,3 +695,17 @@ $screen-minimum: 320px;
...
+
+		fill: $color;
+
+		@if $shade-color {
+			&:hover {
+				fill: $def-danger-shade;
+			}
+		}
+	}
+}
```
erosdome: Define the mixin locally, inside the `.icon` selector: the *_default.scss* is reserved for general mixins, whereas this is very specific for the below 2 cases.

tadam313: moved out `circle` part so made it more reusable

PR #429, opened 2020-02-05
Webpack integration

tadam313:
* added webpack
* created custom rails transformer 
* changed couple things to accommodate new transpilation logic

**As a result**: we have sourcemaps!, can include react and bitkit!


package.json
```diff
@@ -26,21 +29,39 @@
...
     "karma-jasmine": "^3.1.0",
     "karma-jsdom-launcher": "^8.0.0",
     "karma-spec-reporter": "^0.0.32",
-    "onchange": "^6.1.0"
+    "mini-css-extract-plugin": "^0.9.0",
+    "monaco-editor-webpack-plugin": "^1.7.0",
+    "node-sass": "^4.13.1",
+    "onchange": "^6.1.0",
+    "sass-loader": "^8.0.2",
+    "shell-loader": "git@github.com:tadam313/shell-loader.git#add-script-options",
```
tadam313: I had to make a change in there to being able to pass options to node `exec`. I pushed my changes back: https://github.com/localjo/shell-loader/pull/1

rails/transformer.rb
```diff
@@ -0,0 +1,81 @@
...
+        "<link href=\"#{href}\" rel=\"#{rel}\" type=\"#{type}\" sizes=\"#{sizes}\" />"
+    end
+
+    def include_slim(name, options = {}, &block)
+        path = "#{Dir.pwd}/../source/templates/"
+        Slim::Template.new("#{path}#{name}.slim", options).render(self, &block)
+    end
+
+    puts Slim::Template.new { input }.render
+end
```
tadam313: it can transform `.erb` and `.slim` files.

source/javascripts/index.js
```diff
@@ -0,0 +1,22 @@
...
+    // templates
+    require.context("../templates", true, /^[^_].*\.slim$/)
+];
+
+import "../index.slim";
+import "../stylesheets/main.scss";
+
+ctxs.forEach(function(ctx) {
+    ctx.keys().forEach(ctx);
+});
```
tadam313: entry file for out code, can include templates and stylesheets

source/javascripts/vendor.js
```diff
@@ -1,12 +1,27 @@
...
+import 'js-yaml'
+
+$(document).ready(function() {
+    document.body.addEventListener('DOMSubtreeModified', function() {
+        _.each($('*[sanitized-markdown] a[href]'), function(anAnchor) {
+            $(anAnchor).attr('target', '_blank');
+            $(anAnchor).attr('rel', 'noreferrer noopener nofollow');
+        });
+    });
+});
```
tadam313: vendor file, later we can break it down even more if it's too big

source/templates/_auto_code_signing.slim
```diff
@@ -21,7 +21,7 @@
 						span After running the script,
 						strong< upload all exported files
 						span< below:
-					= partial "templates/tab-code_signing-certificates"
+					== include_slim "_tab-code_signing-certificates"
```
tadam313: single `=` supposed to html encode included templates according to the doc, no idea how this could work so far

webpack.config.js
```diff
@@ -0,0 +1,160 @@
...
+    main: "./javascripts/index.js",
+  },
+
+  optimization: {
+    minimizer: isProd ? [
+      new UglifyJsPlugin({
+        uglifyOptions: {
+          cache: true,
+          parallel: true,
+          mangle: false
```
tadam313: we can't mangle our variable names, since we omitted them in angular DI :( it uses function argument names now for lookup so it's impossible to minify entirely. Might worth writing some regex replace as a next step

webpack.config.js
```diff
@@ -0,0 +1,160 @@
...
+        "!rename",
+        "!smartSelect",
+        "!snippets",
+        "!suggest",
+        "!toggleHighContrast",
+        "!toggleTabFocusMode",
+        "!transpose",
+        "!wordHighlighter",
+        "!wordOperations",
+        "!wordPartOperations"
```
tadam313: excluded basically everything -> could save 1MB in bundle

rails/strings.rb
```diff
@@ -0,0 +1,1062 @@
+def strings
+    {
```
tadam313: moved every string here since reading a yml on every transformer run would impact performance quite heavily. This is a tradeoff to have strings in ruby file and require them from the transformer. In the long run we need to figure out strings management.

bitrise.yml
```diff
@@ -75,33 +75,31 @@ workflows:
             docker-compose rm -f -v
             docker-compose build
 
-  middleman-build:
+  client_build:
```
simonmarton:
```suggestion
  client-build:
```

PR #431, opened 2020-02-07
remove unnecessary check

lszucs:
empty secret file check seemed unnecessary, because everything seems to run fine with
- no secrets file at all
- secrets file with single comment (i.e. semantically empty secrets file)

tests are green after removing the check


apiserver/utility/utility_test.go
```diff
@@ -114,15 +114,6 @@ workflows:
 		require.True(t, strings.Contains(err.Error(), "Validation failed: Config validation error: Failed to get config (bitrise.yml) from base 64 data, err: Failed to parse bitrise config, error: missing format_version"), err.Error())
 	}
 
-	t.Log("Invalid secrets - empty")
```
lpusok: Could add a positive test case, instead of removing it completely.

lszucs: See my latest commit please. There was a test for valid cases, I extended the list.

PR #435, opened 2020-02-11
Fix webpack deploy

tadam313:
**DESC:** assets path was not correct, since ruby transformation logic had to be replicated from webpack

steps: 

* added `publicPath` calculation logic to webpack config
* explicitly copied favicons to the output (since these are not required from code)
* added prod `NODE_ENV` to binary release
* fixed `image_path` rails transformation
* removed unused favicons


package.json
```diff
@@ -57,7 +58,7 @@
     "node-sass": "^4.13.1",
     "onchange": "^6.1.0",
     "sass-loader": "^8.0.2",
-    "shell-loader": "tadam313/shell-loader#add-script-options",
+    "shell-loader": "^1.2.0",
```
tadam313: it is released 👍

webpack.config.js
```diff
@@ -1,12 +1,18 @@
...
 
 const OUTPUT_FOLDER = path.join(__dirname, "build");
 const CODEBASE = path.join(__dirname, "source");
-const isProd = process.env.NODE_ENV === "prod";
+
+const { NODE_ENV, RELEASE_VERSION, MODE } = process.env;
+const isProd = NODE_ENV === "prod";
+
+const urlPrefix = MODE === "website" ? "bitrise_workflow_editor-" : "";
+const publicPath = isProd ? `/${urlPrefix}${RELEASE_VERSION}/` : "";
```
tadam313: this: https://github.com/bitrise-io/bitrise-workflow-editor/blob/master/rails/transformer.rb#L19 had to be duplicated here

PR #436, opened 2020-02-13
master deploy fixes

tadam313:
* `ng-init` only run once so selected step wasn't updated -> had to revert back to use selected step
* change `font-url` to `url`, webpack didn't load the font properly
* safety check for $digest cycle

### DIFF
--- a/source/javascripts/controllers/_WorkflowsController.js.erb
+++ b/source/javascripts/controllers/_WorkflowsController.js.erb
@@ -646,7 +646,10 @@ angular.module("BitriseWorkflowEditor").controller("WorkflowsController", functi
 		}
 
 		viewModel.selectedStep = step;
-		$scope.$digest();
+
+		if (!$scope.$$phase) {
+			$scope.$digest();
+		}
 	};
 
 	$scope.$watch(function() {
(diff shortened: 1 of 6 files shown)


source/javascripts/controllers/_WorkflowsController.js.erb
```diff
@@ -646,7 +646,10 @@ angular.module("BitriseWorkflowEditor").controller("WorkflowsController", functi
 		}
 
 		viewModel.selectedStep = step;
-		$scope.$digest();
+
+		if (!$scope.$$phase) {
+			$scope.$digest();
+		}
```
tadam313: this is needed since we are calling it both from react and angular context

webpack.config.js
```diff
@@ -11,7 +11,7 @@ const CODEBASE = path.join(__dirname, "source");
 const { NODE_ENV, RELEASE_VERSION, MODE } = process.env;
 const isProd = NODE_ENV === "prod";
 
-const urlPrefix = MODE === "website" ? "bitrise_workflow_editor-" : "";
+const urlPrefix = MODE === "WEBSITE" ? "bitrise_workflow_editor-" : "";
```
tadam313: it's uppercase here 🙈

PR #438, opened 2020-02-14
[WEB-3293] Handle null response at YML validation

slapec93:


source/javascripts/services/_appService.js.erb
```diff
@@ -334,6 +334,9 @@ angular.module("BitriseWorkflowEditor").service("appService", function($q, reque
 			appService.savedAppConfig = undefined;
 			appService.savedAppConfigYML = appService.appConfigYML;
 
+			if (validationData.warnings === null) {
```
tadam313: can we change it to `!validationData.warnings`

apiserver/utility/utility.go
```diff
@@ -44,12 +45,15 @@ func ValidateBitriseConfigAndSecret(bitriseConfig, secretsConfig string) (*Warni
 		return nil, fmt.Errorf("Validation failed: %s", strings.Join(errorStrs, " | "))
 	}
 
-	warningItems := &WarningItems{Config: []string{}, Secrets: []string{}}
+	warningItems := WarningItems{}
 	if len(bitriseWarns) > 0 {
 		warningItems.Config = bitriseWarns
 	}
```
tadam313: can't we do? `if len(bitriseWarns) > 0 {} else { return nil, nil }` instead of using `reflect`?

tadam313:
@slapec93 also go testing failed

slapec93:
Reopening in other PR

PR #442, opened 2020-02-18
Fix unnecessary polling [WEB-2017]

erosdome:


source/javascripts/controllers/_CertificateController.js.erb
```diff
@@ -30,20 +32,30 @@ angular.module("BitriseWorkflowEditor").controller("CertificateController", func
...
-	function loadCertificates() {
-		viewModel.loadCertificatesProgress.start("<%= data[:strings][:code_signing][:certificate][:load_progress][:in_progress] %>");
+	viewModel.init = function() {
+		loadCertificates();
+	};
+
+	function loadCertificates(shouldClearPreviousLoad) {
+		if (shouldClearPreviousLoad === undefined) {
+			shouldClearPreviousLoad = false;
+		};
```
tadam313: why we need this?

source/javascripts/controllers/_GenericFileController.js.erb
```diff
@@ -22,20 +24,29 @@ angular.module("BitriseWorkflowEditor").controller("GenericFileController", func
...
 
-	function loadGenericFiles() {
-		viewModel.loadGenericFilesProgress.start("<%= data[:strings][:code_signing][:generic_file][:load_progress][:in_progress] %>");
+	viewModel.init = function() {
+		loadGenericFiles();
+	}
+
+	function loadGenericFiles(shouldClearPreviousLoad) {
+		if (shouldClearPreviousLoad === undefined) {
+			shouldClearPreviousLoad = false;
```
tadam313: same, I don't think we need this if we only check it as a boolean

erosdome: Yep, it's just to make it more explicit what will be the default value, but we can remove it, functionally it's okay without this too.

source/javascripts/controllers/_AndroidKeystoreController.js.erb
```diff
@@ -101,7 +110,7 @@ angular.module("BitriseWorkflowEditor").controller("AndroidKeystoreController",
 			viewModel.actionProgress.success();
 
 			viewModel.uploadedKeystoreFile = new KeystoreFile("");
-			loadKeystoreFile();
+			loadKeystoreFile(true);
```
tadam313: can we change it to resolve logic by using `shouldHideLoading` or sth like that, so we concentrate `true` parameter passing to recursion only.

source/templates/_tab-code_signing-certificates.slim
```diff
@@ -1,4 +1,4 @@
-.files.certificates[ng-controller="CertificateController as certificateCtrl" ng-if="codeSigningCtrl.sectionConfigurations.certificate.shouldShow && !codeSigningCtrl.isSectionFilteredByFileTypeMenu(codeSigningCtrl.sectionConfigurations.certificate)"]
+.filese.certificates[ng-controller="CertificateController as certificateCtrl" ng-init="certificateCtrl.init()" ng-if="codeSigningCtrl.sectionConfigurations.certificate.shouldShow && !codeSigningCtrl.isSectionFilteredByFileTypeMenu(codeSigningCtrl.sectionConfigurations.certificate)"]
```
tadam313: can you change it back please? 🙂

PR #450, opened 2020-02-26
Add Workflows tab specification for Workflow control bar, Step list, Step details, Step Inputs (part of it)

erosdome:


tadam313:
@erosdome  I think we can merge this to master, it won't harm

PR #455, opened 2020-03-02
Fix search deprecated steps, List steps by ID not CVS

simonmarton:


source/javascripts/services/_stepSourceService.js.erb
```diff
@@ -376,9 +386,11 @@
 			stepSourceService.loadLatestOfficialSteps = function(projectType) {
 				var projectTypes = projectType && [projectType];
 
-				return stepLibSearchService.list({ latestOnly: true, projectTypes: projectTypes }).then(function(steps) {
-					addStepsToDefaultLibrary(steps);
-				});
+				return stepLibSearchService
+					.list({ latestOnly: true, projectTypes: projectTypes, includeDeprecated: false })
```
tadam313: do we need to filter for deprecated here? what if a user is using a deprecated step and we can not display the latest version of that on the UI?

simonmarton: Yup, this is called in the Search sidebar. We're not filtering for deprecated when loading Step info for WFs.

PR #458, opened 2020-03-08
[WEB-3686] e2e tests kickoff

tadam313:
* cypress with cucumber parser
* some test cases
* ci integration
* example bitrise yml

**Idea: upon every step we discard changes, so every step starts with clean state -> steps are independent from each other**


_scripts/run_api.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+EXECUTABLE=./_bin/workflow-editor-Linux-x86_64
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    EXECUTABLE=./_bin/workflow-editor-Darwin-x86_64
+fi
+
+PORT=$PORT BITRISE_CONFIG=$1 $EXECUTABLE
```
tadam313: small util starts API platform agnostic

bitrise.yml
```diff
@@ -312,8 +318,8 @@ workflows:
             #!/bin/bash
             set -ex
 
-            cat bitrise.yml | sed -e "s/^  \- RELEASE_VERSION\: .*$/  - RELEASE_VERSION: ${NEW_RELEASE_VERSION}/" > new_bitrise.yml
-            mv new_bitrise.yml bitrise.yml
+            sed -i "" -e "s/^  \- RELEASE_VERSION\: .*$/  - RELEASE_VERSION: ${NEW_RELEASE_VERSION}/" bitrise.yml
+            sed -i "" "s/\"version\": \".*\"/\"version\": \"$NEW_RELEASE_VERSION\"/" package.json
```
tadam313: we need to keep `package.json` up-to date going forward

spec/integration/workflow-steps.feature
```diff
@@ -0,0 +1,46 @@
...
+    And I confirm on "alert" popup
+    And I click on "Step Latest Version Updater"
+    Then I should see "Version: 2.2.2" in "Step Versions"
+    And "Step Version Success Icon" should "be visible"
+
+  Scenario: Always latest vs latest major lock
+    When I click on "Third step"
+    Then I should see "Version: 4.0.5" in "Step Version"
+    When I select "4.x.x" from "Version selector"
+    Then I should see "Version: 4.0.5" in "Step Versions"
```
tadam313: this is automated

spec/integration/workflows.feature
```diff
@@ -0,0 +1,26 @@
...
+    Then "Workflow Add Popup" should "not be visible"
+    And Workflow appeared with name "Test"
+
+  Scenario: Delete Workflow
+    Given Workflow with name "ToBeDeleted"
+    When I click on "Delete Workflow Button"
+    And I confirm on "default" popup
+    Then I should not see "ToBeDeleted" in "Selected Workflow Name"
+    And I should see "wf1" in "Selected Workflow Name"
+
```
tadam313: this is automated

webpack.config.js
```diff
@@ -4,15 +4,17 @@ const MonacoWebpackPlugin = require("monaco-editor-webpack-plugin");
...
 const CopyPlugin = require("copy-webpack-plugin");
 const { ProvidePlugin } = require("webpack");
+const { version } = require("./package.json");
 
 const OUTPUT_FOLDER = path.join(__dirname, "build");
 const CODEBASE = path.join(__dirname, "source");
 
 const { NODE_ENV, RELEASE_VERSION, MODE } = process.env;
 const isProd = NODE_ENV === "prod";
+const wfVersion = RELEASE_VERSION || version;
```
tadam313: outside of the release flow we don't have version number -> need to fall back when env var is empty

spec/integration/test_bitrise.yml
```diff
@@ -0,0 +1,132 @@
...
+        inputs:
+        - access_token: $ACCESS_KEY
+  wf3:
+    after_run:
+    - wf1
+    - wf2
+    envs:
+    - opts:
+        is_expand: false
+      project: vrbo
```
tadam313: test yml, needs to be refined

bitrise.yml
```diff
@@ -183,11 +180,20 @@ workflows:
...
-        title: Karma tests
+        title: Unit tests
         inputs:
-        - content: npm test
+        - content: npm run karma
+    - script:
+        title: E2E tests
+        inputs:
+        - content: npm run e2e-ci
+    - deploy-to-bitrise-io: {}
```
tadam313: **opinion needed**: do we wanna make binaries available as artifacts at every build?

cypress.json
```diff
@@ -0,0 +1,14 @@
+{
+  "projectId": "rnnyfb",
+  "integrationFolder": "spec/integration",
+  "testFiles": "**/*.feature",
+  "pluginsFile": "spec/integration/cypress/plugins",
+  "supportFile": "spec/integration/cypress/support",
+  "screenshotsFolder": "spec/integration/cypress/screenshots",
+  "videosFolder": "spec/integration/cypress/videos",
+  "videoUploadOnPasses": false,
```
tadam313: @HHogg idea behind it: we only need videos for investigations, so I didn't wanna overload our account. WDYT? should we instead upload every video?

package.json
```diff
@@ -1,88 +1,100 @@
 {
```
tadam313: indentation change, sorry

spec/integration/common.js
```diff
@@ -0,0 +1,74 @@
...
+export const select = (value, element) => {
+  $(element).select(value);
+};
+
+export const type = (text, element) => {
+  $(element).type(text, { force: true }).trigger('input');
+};
+
+export const popupConfirm = (type) => {
+  let selector = '#default-confirm-popup-body .yes';
```
HHogg: I really suggest avoiding using CSS selectors and tags for selecting element. It makes them quite brittle to changes (for example changing the CSS classes or HTML element). Another approach is to use data-* attributes with randomly generated IDs, with the sole responsibility for being used in the ATs. It's then also helpful to store these inside a single config file with more identifiable names like ...

```json
{ 
  "WorkflowEditorConfirmButton": "ad5c7bad1f57"
}
```

```jsx
import atids from '../path/to/atids.json';

<button data-at-id={ atids.WorkflowEditorConfirmButton }>Confirm</button>
```

I use  the below to generate IDs. 

```
cat /dev/random | head -1 | shasum | cut -c 1-12
```

spec/integration/workflow-steps.feature
```diff
@@ -0,0 +1,46 @@
...
+  The selected Step's Inputs' properties can be viewed & some of them can be edited.
+  Steps of any Workflow are listed, they can be selected, and their name, version, icon & badges are displayed.
+
+  Background:
+    Given editor is open
+
+  Scenario: Step's selected
+    When I click on "First step"
+    Then I should see "GitHub Status" in "Step Title"
+    Then I should see "2.2.2" in "Step Version"
```
HHogg: Is testing an exact version number likely to need updating all of the time? Is this adding anything over testing under this  that the steps inputs have been rendered?

spec/integration/workflow-steps.feature
```diff
@@ -0,0 +1,46 @@
+ Feature: Workflow steps
+
+  The selected Step's properties can be viewed & some of them can be edited.
+  The selected Step's Inputs' properties can be viewed & some of them can be edited.
+  Steps of any Workflow are listed, they can be selected, and their name, version, icon & badges are displayed.
+
+  Background:
+    Given editor is open
+
+  Scenario: Step's selected
```
HHogg: I'm not a huge fan of cucumber, but thats just an opinion. Having used it in the past, I've found that when developers are writing the tests, they are more comfortable with using a programatic API, which can be a bit more flexible especially when trying to debug a test.

tadam313: I know it's controversial, we already have large number of test cases written in it (which are not automated) so it made sense for us to use it. I've used it as well many times, one advantage of this is readability and expressiveness over programatic apis. But we have underlying functions as well so if we think it's giving us pain we can get rid of this and use plain reusable functions quite easily.

spec/integration/workflow-steps.feature
tadam313: this step version is locked in a test fixture file, it has nothing to do with editor version, so this is essentially predefined test data.

spec/integration/common.js
tadam313: totally agree with you and CSS selectors are not ideal, but we wanted to balance between effort and result. Editor was not prepared for testing and most of elements can not be identified easily. However theory we wanted to follow is not to touch the editor only after we have e2e coverage. If we wanted to go with test data attributes we need to touch the editor now and not sure if we wanna risk that. So I'd agree with this but maybe in the next round. WDYT?

spec/integration/workflow-steps.feature
HHogg: Ah ok, sorry I'm not familiar with the setup of the workflow editor. So is there a locally running API this is running against then?

tadam313: yep, we already had a binary built with rice (https://github.com/GeertJohan/go.rice) which acts as a server for the tests. https://github.com/bitrise-io/bitrise-workflow-editor/blob/9afb393ec3047716a79780821dbdb9c27029173b/package.json#L15

package.json
```diff
@@ -13,7 +13,7 @@
     "e2e:dev": "cypress open",
     "e2e:run": "cypress run",
     "e2e:api": "PORT=4000 ./_scripts/run_api.sh ./spec/integration/test_bitrise.yml",
-    "e2e-ci": "concurrently -k -s first \"npm:e2e:api\" \"cypress run --record --key 7af6cee4-8cc2-4d68-bf06-dda30b737dab\"",
+    "e2e-ci": "concurrently -k -s first \"npm:e2e:api\" \"cypress run --record --key $CYPRESS_API_KEY\"",
```
tadam313: the key was regenerated

spec/integration/common.js
HHogg: I think it's a pattern that shouldn't really be started, if the intention is to not use it going forward that is. Adding data attributes to the elements shouldn't cause any issues, and if it did it should be picked up by the tests in this PR 🙃 

I'm just expressing my opinion though and wouldn't think of block anything, it's absolutely your call.

webpack.config.js
erosdome: I think we do, the `RELEASE_VERSION` env var is not specific to the releasing Workflow, it is app-level. We can assume it will always be set, just like we do with the `MODE` env var.

spec/integration/common.js
tadam313: Problem is I'm not sure how to use these identifiers easily in angular templates 🤔 React is ok but in angular tempaltes we don't really have an easy way to use data from a simple json

webpack.config.js
tadam313: problem is it failed without this. For some reason `RELEASE_VERSION` was empty on the CI and on locally as well. Not sure why but I could not get it to work without this. But since we have a way to use package json easily from here I can't see a huge issue or is it @erosdome ?

spec/integration/common.js
HHogg: Ah ok, sorry I don't know about that. It's been a few years since I've done any Angular.

tadam313: ok let me investigate this, there is might be an easy way but definitely more complex than with react, so it'd be a follow up task

PR #459, opened 2020-03-11
[WEB-3746] Added Datadog logging

simonmarton:
Had quite some trouble with running Karma tests properly :(


source/javascripts/services/logger-service-instance.js.erb
```diff
@@ -0,0 +1,10 @@
+import { datadogLogs } from "@datadog/browser-logs";
+import DataDogLoggerService from "./logger-service";
+
+var loggerService = new DataDogLoggerService(datadogLogs, "<%= data[:constants][:datadog][:apiKey] %>");
+
+angular.module("BitriseWorkflowEditor").factory("loggerService", function() {
```
tadam313: can we call itt `logger`?

source/javascripts/_BitriseWorkflowEditor.js.erb
```diff
@@ -68,4 +59,11 @@
 			$httpProvider.defaults.xsrfHeaderName = "X-CSRF-TOKEN";
 		}
 	]);
+
+	BitriseWorkflowEditor.factory("$exceptionHandler", function($log, loggerService) {
+		return function myExceptionHandler(error) {
+			loggerService.error(error.message, error);
+			$log.error(error);
```
tadam313: what does `$log` do?

source/javascripts/services/logger-service-instance.js.erb
```diff
@@ -0,0 +1,10 @@
+import { datadogLogs } from "@datadog/browser-logs";
+import DataDogLoggerService from "./logger-service";
+
+var loggerService = new DataDogLoggerService(datadogLogs, "<%= data[:constants][:datadog][:apiKey] %>");
```
tadam313: what if we say `global.logger = new DataDogLoggerService(...) ... factory(function { return global.logger; })`. So we have logger both on global and angular scope. That's nasty but at least we can mock it via karma by `global.logger = new Mock()` WDYT? Or we can also not test the logger itself, the logic is pure simply in it anyway.

source/javascripts/controllers/_WorkflowsController.js.erb
```diff
@@ -204,6 +205,7 @@ import { safeDigest } from "../services/react-compat";
 										}
 									);
 								} catch (error) {
+									loggerService.error(error.message, error);
```
tadam313: what does it log here? should we use `reject` part of the promise? Also couple lines above there is a `console.error` entry should we replace that as well? :)

source/javascripts/_BitriseWorkflowEditor.js.erb
simonmarton: This is the default functionality for the exception handler, it just logs to the browser console by default.

source/javascripts/services/logger-service-instance.js.erb
simonmarton: That's a good idea, I don't think it's that nasty, having it tested is a good trade-off. I'll try it.

source/javascripts/_BitriseWorkflowEditor.js.erb
tadam313: do we want it in production?

simonmarton: I'd say so, at least until we have a stable release and monitoring solution that has proven we actually see errors on our end.

tadam313: ok good point

source/javascripts/services/logger.ts
```diff
@@ -0,0 +1,42 @@
...
+	error(message: string, ctx?: Context): void {
+		this.logger.error(message, ctx);
+	}
+}
+
+// testing purposes
+(<any>window).datadogLogs = datadogLogs;
+
+export default (token: string|undefined): Logger => new DataDogLoggerService(
+    token || (<any>window).DATADOG_API_KEY,
```
tadam313: @simon can you please declare it on window

PR #468, opened 2020-03-23
[WEB-3759] Step input e2e tests

simonmarton:


spec/integration/workflows-steps-inputs.feature
```diff
@@ -12,3 +12,67 @@ Feature: Workflow steps inputs
...
     And "Step Inputs" should "be visible"
     And "Save Button" should "be disabled"
+
+  Scenario: Inputs without categories don't get wrapped into expandable container
+    When I click on "First step"
+    Then "Step Inputs Without Category" should "be visible"
+
+  Scenario: Inputs with categories get wrapped into expandable container
+    When I click on "Second step"
+    Then I should see "Config" in "Second Step Input Category Title"
```
tadam313: can't we use generic element here like `Selected Step Input Category Title`? I believe only one input element will present in the DOM at a single time and that's only the selected, or am I mistaken? Or is it purely a naming issue?

spec/integration/workflows-steps-inputs.feature
```diff
@@ -12,3 +12,67 @@ Feature: Workflow steps inputs
...
+    When I click on "Second Step Input Category"
+    Then "Second Step Input Category Inputs" should "be visible"
+
+  Scenario: Expanding and Closing Input categories' container
+    When I click on "Second step"
+    Then I should see "Config" in "Second Step Input Category Title"
+    When I click on "Second Step Input Category"
+    Then "Second Step Input Category Inputs" should "be visible"
+    When I click on "Second Step Input Category Toggle Button"
+    Then "Second Step Input Category Inputs" should "not be visible"
```
tadam313: Generally I think we could use "selected step...." elements here

spec/integration/workflows-steps-inputs.feature
```diff
@@ -12,3 +12,67 @@ Feature: Workflow steps inputs
...
+
+  Scenario: A changeable Input
+    When I click on "First step"
+    And I click on "First Step Input"
```
tadam313: here as well only "step Input"?

spec/integration/workflows-steps-inputs.feature
```diff
@@ -12,3 +12,67 @@ Feature: Workflow steps inputs
...
+    Then "First Step Input Insert Variable Button" should "be visible"
+    When I click on "First Step Input Insert Variable Button"
+    Then "Insert Variable Popup" should "be visible"
+
+  Scenario: Input in non-edit mode
+    When I click on "First step"
+    Then "First Step Input Change Button" should "be visible"
+  # And "First Step Input Description" should "be visible" # Fails locally
```
tadam313: ? 😮

spec/integration/workflows-steps-inputs.feature
```diff
@@ -12,3 +12,67 @@ Feature: Workflow steps inputs
...
+    When I click on "First step"
+    Then I should see "GitHub auth token" in "First Step Input Title"
+    Then "First Step Input Required Badge" should "be visible"
+
+  Scenario: Empty required Input
+    When I click on "First step"
+    Then I should see "GitHub auth token" in "First Step Input Title"
+    And I click on "First Step Input"
+    Then "Selected Input Textarea" should "be empty"
+# Then "Selected Input Textarea" should have "purple" "border-color" style
```
tadam313: what happened with this?

spec/integration/workflows-steps-inputs.feature
```diff
@@ -12,3 +12,67 @@ Feature: Workflow steps inputs
...
+    Then "First Step Input Sensitive Badge" should "be visible"
+    And I click on "First Step Input"
+    Then "First Step Input Clear Button" should "be visible"
+
+  Scenario: Clearing the sensitive Input's value
+    When I click on "First step"
+    And I click on "First Step Input"
+    When I click on "First Step Input Clear Button"
+    Then "Selected Input Textarea" should "be empty"
+    Then "First Step Input Clear Button" should not exist
```
tadam313: can't we use regular `should` rule for this? just to enforce reusability

spec/integration/workflows-steps-inputs.feature
simonmarton: These are all about the second input in the selected step (not the second step). I did think about naming it `Selected Step Second Input ...` WDYT @tadam313 ?

spec/integration/workflows-steps-inputs.feature
simonmarton: That fails due to jQuery not finding the element, and some other `should` tests failed when using `cy.get` unfortunately

spec/integration/workflows-steps-inputs.feature
tadam313: I'd suggest to do that yeah, I think it would be more clear what it's referring to

spec/integration/workflows-steps-inputs.feature
tadam313: did it fail because of using `eq` stuff? 🤔 Element selector is now refactored so it might work, I'll check it

tadam313: @simonmarton yeah it works for me now in the current master 🎉

tadam313:
```suggestion
    Then "First Step Input Clear Button" should "not exist"
```

spec/integration/workflows-steps-inputs.feature
```diff
@@ -12,3 +12,68 @@ Feature: Workflow steps inputs
...
+    When I click on "First step"
+    And I click on "Selected Step First Input"
+    When I click on "Selected Input Clear Button"
+    Then "Selected Input Textarea" should "be empty"
+    Then "Selected Input Clear Button" should "not exist"
+
+  Scenario: Required Input
+    When I click on "First step"
+    Then I should see "GitHub auth token" in "Selected Step First Input Title"
+    Then "Selected Step First Input Required Badge" should "be visible"
```
tadam313: can we merge this scenario to the one below this one? In order to save some execution time. Also `Then` can be combined with `And` as well

PR #469, opened 2020-03-24
Move core API functionality to separate library 

slapec93:


tadam313:
@slapec93 I'll close this PR for now as this goes to engagement team and also they might plan to make WFE as a service in long term etc..

PR #472, opened 2020-03-30
Plugin update + logging refinement

tadam313:
* include stack trace to logs
* dep update for CLI


source/javascripts/_BitriseWorkflowEditor.js.erb
```diff
@@ -71,8 +71,8 @@ window.mode = "<%= mode %>";
 	]);
 
   BitriseWorkflowEditor.factory("$exceptionHandler", function($log, logger) {
-		return function(error) {
-			logger.error(error.message, error);
+		return function(error, cause) {
+			logger.error(error, { cause: cause });
```
tadam313: logger will take care of how to log the errors

source/javascripts/services/logger.ts
```diff
@@ -60,7 +60,12 @@ class DataDogLoggerService implements Logger {
 		this.logger.warn(message, ctx);
 	};
 
-	error = (message: string, ctx?: Context) => {
+	error = (error: Error, ctx?: Context) => {
+		if (!error) {
+			return;
+		}
+
+		const message = `${error.message}\n${error.stack || 'No stack'}`;
```
tadam313: construct message and stack information for every error

PR #476, opened 2020-04-16
[WEB-3843] Increase E2E test coverage

slapec93:


spec/integration/common.js
```diff
@@ -130,4 +130,8 @@ Then("{string} should have {string} {string} style", (element, cssValue, cssProp
 	$(element).should("have.css", cssProperty, styleValueSelector(cssValue));
 });
 
+Then("{string} should have attribute {string} with value {string}", (element, attrName, attrValue) => {
+	$(element).should("have.attr", attrName, attrValue);
+});
```
tadam313, line 135: this could be written as `should "have attr: target: _blank"`, should handle could understand this format. Have you tried that?

slapec93: @tadam313 I just thought this way the Gherkin test case is more readable

PR #478, opened 2020-04-17
[WEB-3776] Follow S3 folder structure

simonmarton:
Related to https://github.com/bitrise-io/bitrise-website/pull/4073


webpack.config.js
```diff
@@ -9,19 +9,19 @@ const { version } = require("./package.json");
...
 const CODEBASE = path.join(__dirname, "source");
 
-const { NODE_ENV, MODE } = process.env;
+const { NODE_ENV, MODE, PUBLIC_URL_ROOT } = process.env;
 const isProd = NODE_ENV === "prod";
 
-const urlPrefix = MODE === "WEBSITE" ? "bitrise_workflow_editor-" : "";
-const publicPath = `/${urlPrefix}${version}/`;
+const urlPrefix = MODE === "WEBSITE" ? PUBLIC_URL_ROOT : "";
+const publicPath = isProd ? `${urlPrefix}/${version}` : "";
```
tadam313, line 16: I don't think we need `isProd` check here, locally we are running CLI mode when the `urlPrefix ` is empty

webpack.config.js
```diff
@@ -40,7 +40,7 @@ const assetExporter = (regex, folder) => ({
 			options: {
 				outputPath: folder,
 				name: "[name].[ext]",
-				publicPath: path.join(publicPath, `/${folder}`)
+				publicPath: isProd ? `${publicPath}/${folder}` : path.join(publicPath, folder)
```
tadam313, line 43: what's the difference here?

simonmarton: path join break http urls
```js
> path.join('http://whate.ver', 'blah')
'http:/whate.ver/blah'
```

tadam313: ohh ok then let's remove `path.join` and only use the first version

webpack.config.js
```diff
@@ -9,19 +9,19 @@ const { version } = require("./package.json");
...
 const CODEBASE = path.join(__dirname, "source");
 
-const { NODE_ENV, MODE } = process.env;
+const { NODE_ENV, MODE, PUBLIC_URL_ROOT } = process.env;
 const isProd = NODE_ENV === "prod";
 
-const urlPrefix = MODE === "WEBSITE" ? "bitrise_workflow_editor-" : "";
-const publicPath = `/${urlPrefix}${version}/`;
+const urlPrefix = MODE === "WEBSITE" ? PUBLIC_URL_ROOT : "";
+const publicPath = `${urlPrefix}/${version}`;
```
tadam313, line 16: hmm I've just checked locally, we need that trailing `/` at the end otherwise it breaks locally :(

tadam313: E2E test does not break unfortunately because the problem only exists on bitrise.yml editor page @simonmarton

simonmarton: ack 😢

PR #482, opened 2020-04-24
[Web-3684] Workflow editor ui bugs #plugin

zoltanszabo-bitrise:
- Add missing step title tests
- Add missing step title test data
- Parse out the title from CVS property of the step, whenever the title is not provided
- Extend tooltip with CVS to let the user find the information lost with the title change
- Fix `StepItem` usage in `workflows.slim` file (wrong attribute casing: camelCase -> kebab-case)
- Rename `StepItem.title` to `StepItem.displayName` (in usage it becomes display-name)
- Fix `StepItem` badge icons
- Extend `StepItem` badge icons with the `community-created` as in case of the step details
- Amend the required input field test to expect `red` `border-color`
- Changed required empty input filed `border-color` to `red`


source/javascripts/_componentRegister.js
```diff
@@ -12,7 +12,7 @@ angular
 		"rStepItem",
 		register(StepItem, [
 			"step",
-			"title",
+			"displayName",
```
tadam313, line 15: sorry don't have context here, can you elaborate why `displayName` is better than `title`? what are the differences?

source/javascripts/components/StepItem/StepItem.tsx
```diff
@@ -65,9 +66,10 @@ const StepItem: FC<StepItemProps> = ({
 		<span className="info">
 			<strong>
 				<Text className="title" ellipsis>
-					{title}
+					{displayName}
 				</Text>
-				{isVerified && <img className="verified" src={verifiedIcon} />}
+				{isVerified && !isDeprecated && <img className="verified" src={verifiedIcon} />}
+				{!isVerified && !isDeprecated && <img className="community-created" src={communityCreatedIcon} />}
```
tadam313, line 72: ohh you should catch up with @simonmarton he changed this part and ported it to Bitkit

source/javascripts/components/_Step.js
```diff
@@ -29,9 +29,31 @@ import { normalizeIconUrl } from "./StepItem/StepItem";
...
+
+		function diplayNameFromCvs(cvs) {
+			var lastDelimiter = cvs.lastIndexOf("/");
+
+			if (lastDelimiter != -1) {
+				cvs = cvs.substring(lastDelimiter + 1);
+			}
+
+			return cvs;
+		}
```
tadam313, line 55: can you please add unit tests to the new logics in here?

source/javascripts/components/_Step.js
```diff
@@ -29,9 +29,31 @@ import { normalizeIconUrl } from "./StepItem/StepItem";
...
-			return this.cvs;
+			return diplayNameFromCvs(this.displayCvs());
 		};
 
+		Step.prototype.displayCvs = function() {
+			return (
+				(this.cvs.startsWith("git::") || this.cvs.startsWith("path::")) 
+					? this.cvs.replace(/(git|path)::/g, "")
+					: this.cvs
+			);
```
tadam313, line 40: can't we tackle this in a single regex? like replace `/^(git|path)::/g`, it will only replace this wen the cvs starts with that prefix. You should also add unit tests for this :)

source/javascripts/_componentRegister.js
zoltanszabo-bitrise: Since `title` is a reserved/existing HTML attribute, which is used to add tooltip info to an element, it showed up accidentally whenever hovering over the step with the wrong message. (step.displayName() since that was the input prop fro the react element)
Basically, it is the fix for https://bitrise.atlassian.net/browse/WEB-3943

tadam313: ahh nice, good catch ok 👍

source/javascripts/components/StepItem/StepItem.tsx
zoltanszabo-bitrise: I think it would be better than to let @simonmarton's PR go into master first, and I would rebase my changes to that one.

source/javascripts/components/_Step.js
zoltanszabo-bitrise: Sure, will do.

source/javascripts/components/_Step.js
zoltanszabo-bitrise: Will do.

zoltanszabo-bitrise:
Waiting for https://github.com/bitrise-io/bitrise-workflow-editor/pull/486 to be merged

source/javascripts/components/_Step.js
```diff
@@ -36,9 +36,27 @@ import { normalizeIconUrl } from "./StepItem/StepItem";
...
-			return this.cvs;
+			return diplayNameFromCvs(this.displayCvs());
 		};
 
+		Step.prototype.displayCvs = function() {
+			return this.cvs.replace(/^(git|path)::/g, "");
+		}
+
+		Step.prototype.displayTooltip = function() {
+			return `${this.displayName()}<br>${this.displayCvs()}`;
```
tadam313, line 47: ahh sorry just realised, we do not have ES6 support in plain js files, can you please backport it to ES5? :)

zoltanszabo-bitrise: Ohh, don't we have a Babel loader for taking care of the transpile for js files?

tadam313: nope and that's intentional to force people to use typescript instead and speed up the migration 😄

zoltanszabo-bitrise: OK, got it :)

PR #483, opened 2020-04-24
[WEB-3412] fix(workflow-selector): Long names

zoltanszabo-bitrise:
In case the workflow name is too long and two workflows only differ at their ends,
the user was not able to distinguish them.

To make it more flexible the following changes were made:
- Allow word-wrapping to handle harder edge cases
- On desktop viewport (+580px) make the selector wider, to give more room
- Under 580px width, let the selector fill up 100% of the width

This required some fine-tuning of the current style:
- center the dot, whenever the name of the workflow wraps into a new line (not selected case)
- make text behave more like when it is selected

The solution was to utilize `display: flex` on the non-selected items as well.

### DIFF
--- a/source/stylesheets/_tab-workflows.scss.erb
+++ b/source/stylesheets/_tab-workflows.scss.erb
@@ -60,11 +60,14 @@ section.manage-workflows {
 			$li_padding: 12px;
 			$li_fontsize: 16px;
 
-			@include position(absolute, 100%, 0);
+			@include position(absolute, 100%, null, null, 0);
 			z-index: $zindex-workflow_selector;
-			width: 100%;
 			margin-top: 10px;
 			margin-bottom: 0;
+			width: 548px;
+			@include media-below(580px) {
+				width: 100%;
+			}
 			border: $border;
 			@include border-radius($def-radius);
 			padding-left: 0;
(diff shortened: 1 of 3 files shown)


source/stylesheets/_tab-workflows.scss.erb
```diff
@@ -60,11 +60,14 @@ section.manage-workflows {
...
 			$li_fontsize: 16px;
 
-			@include position(absolute, 100%, 0);
+			@include position(absolute, 100%, null, null, 0);
 			z-index: $zindex-workflow_selector;
-			width: 100%;
 			margin-top: 10px;
 			margin-bottom: 0;
+			width: 548px;
+			@include media-below(580px) {
```
tadam313, line 68: can you use a variable instead `580px`? I think we have some constants about this

zoltanszabo-bitrise:
**Desktop**
image

**Mobile**
image

source/stylesheets/_tab-workflows.scss.erb
zoltanszabo-bitrise: There is no 580px constant yet, but most likely it can be calculated somehow based on the `manage-workflows` section since, in the case of 579px, all the buttons except the `workflow-selector` are wrapped into a new line.

PR #490, opened 2020-05-17
[WEB-3980] Always hide a protected secret's value

akobor-bitrise:
JIRA card link: https://bitrise.atlassian.net/browse/WEB-3980


README.md
```diff
@@ -62,6 +62,13 @@ npm start          # start both local plugin api and webpack dev server
 1. In your browser, you can reach the Workflow Editor on `localhost:4000/{version}`. Be aware that you usually have to wait a while until dev server starts up (then refresh)
 1. By default, the Workflow Editor will open the `test_bitrise.yml` from integration folder (used for integration testing). Please do not commit this file if you have any changes with it (e2e tests would fail).
 
+If you would like to run the Workflow Editor in `website` mode, you have to set the following two environment variables:
+
+```bash
+export PUBLIC_URL_ROOT=http://localhost:4000
+export MODE=WEBSITE
```
tadam313, line 69: could we do an npm task to start the editor in website mode?

source/javascripts/services/app-service.ts
```diff
@@ -9,3 +9,11 @@ export const getAppSlug = (): string|null => {
 
   return null;
 };
+
+export const handleSecretAfterSave = (secret: any): void => {
```
tadam313, line 13: can we not use `any` in this case? just define a type for that in this file. Or if that's `Value` we can also port that to TS if easy enough. If that's complex, then you can just define a `type` for that which we later port to a `class`.

akobor-bitrise: Yeah, now it's definitely ugly, you are right :( I was wondering, what could be the most elegant and efficient way to create/port `Variable` to TS (the logic in `_Variable.js.erb` is pretty complex). Do you have any battle-proof suggestion?

tadam313: yeah I see that. I'm ok with creating a type from that for now, it would mean scope explosion

package.json
```diff
@@ -14,6 +14,7 @@
 		"build:plugin": "MODE=CLI ANALYTICS=true npm run build && bitrise run create-binaries",
 		"start:dev": "webpack-dev-server",
 		"start": "concurrently \"npm:start:dev\" \"DEV=true ./_scripts/run_api.sh 4000\"",
+		"start:website": "export MODE=WEBSITE && export PUBLIC_URL_ROOT=http://localhost:4000 && npm run start",
```
tadam313, line 17: do we need exports here? 🤔

akobor-bitrise: I was wondering too, but I can't get it work the way it works one line above :( If you have any suggestion, I'll iterate an other round here

tadam313: can't we do `MODE=WEBSITE PUBLIC_URL_ROOT=http://localhost:4000 npm start`?

akobor-bitrise: hmm, last time I tried this, it didn't work, now it's fine :/ anyway, I fixed it :)

PR #496, opened 2020-06-19
Fix step version update inaccuracy [WEB-4217]

zoltanszabo-bitrise:
Due to multiple things
- multiple version is flagged as latest for the same step
- step version info is unordered when calculating latest version
caused inaccuracies in the workflow editor.

For example:
deploy-to-bitrise-io step latest version was calculated as 1.9.7, however, a newer 1.10.0 was already available.
Due to this, selecting 1.10.x or 1.x.x from the version selector still showed the update available state.


source/javascripts/services/semver-service.ts
```diff
@@ -36,21 +36,26 @@ class SemverService {
...
 		// make WILDCARD always the lowest priority than everything else
-		ver1[notEqInd] = ver1[notEqInd] === this.WILDCARD ? "-1" : ver1[notEqInd];
-		ver2[notEqInd] = ver2[notEqInd] === this.WILDCARD ? "-1" : ver2[notEqInd];
+		ver1[notEqInd] = ver1[notEqInd] === this.WILDCARD ? "-1" : ver1[notEqInd] || "0";
+		ver2[notEqInd] = ver2[notEqInd] === this.WILDCARD ? "-1" : ver2[notEqInd] || "0";
 
 		// comply with compare interface
-		return Math.sign(parseInt(ver1[notEqInd]) - parseInt(ver2[notEqInd])) * -1;
+		const diff = Math.sign(parseInt(ver1[notEqInd]) - parseInt(ver2[notEqInd]));
+		return diff === 0 ? 0 : diff * -1;
```
simonmarton, line 54:
```suggestion
		return -diff;
```

zoltanszabo-bitrise: The reason I differentiate the 0 case, because the previous implementation could even return -0 (not to mention undefined), however, the expected outcomes for a comparison method are -1, 0, 1.

So I would leave it as is: or a small improvement could be:
```diff
return diff === 0 ? 0 : -diff;
```

What do you think?

simonmarton: I'd handle the `undefined` case separately. But why is `-0` an issue, if `0 === -0`?

I meant the commit related to `semverService.spec.js`, as I see there's a lot of update that's only whitespace and indentation update.

Zolee007: That one was VSCode entirely, but I keep an eye on it next time ;)

zoltanszabo-bitrise: Pushed the requested change ;)

PR #500, opened 2020-07-29
[EN-189] Welcome splash modal

Ozonid:
image


source/javascripts/components/SplashModal/SplashModal.scss
```diff
@@ -0,0 +1,13 @@
+.ModalHeader {
+	height: 200px;
+	background: url("https://www.warrenphotographic.co.uk/photography/rubs/03656.jpg");
```
akobor-bitrise, line 3: 🦆 ❤️

source/javascripts/components/SplashModal/SplashModal.scss
```diff
@@ -0,0 +1,13 @@
+.ModalHeader {
+	height: 200px;
```
erosdome, line 2: Do it with `$grid * 50` instead. You will need to include _constants.scss for this.

source/javascripts/components/SplashModal/SplashModal.scss
```diff
@@ -0,0 +1,13 @@
...
+	background: url("https://www.warrenphotographic.co.uk/photography/rubs/03656.jpg");
+	background-position: center;
+	background-size: cover;
+	border-radius: inherit;
+	border-bottom-left-radius: 0;
+	border-bottom-right-radius: 0;
+
+	a.Base--clickable.Link {
+		color: #fff;
```
erosdome, line 11: Do it with a constant, put it in _constants.scss as `$def-white`.

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -59,6 +61,21 @@
 				viewModel.menuSelected(initialMenu);
 			}
 
+			function checkRepoYmlSplashModalVisibility () {
+				$q.all([requestService.getCurrentUserBetaTags(), appService.getPipelineConfig()])
+					.then(function(result) {
+						var betaTags = result[0];
+						
+						var featureAvailable = betaTags.indexOf('yml-from-repo') > -1;
```
erosdome, line 69:
```suggestion
						var featureAvailable = _.contains(betaTags, 'yml-from-repo');
```

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -59,6 +61,21 @@
 				viewModel.menuSelected(initialMenu);
 			}
 
+			function checkRepoYmlSplashModalVisibility () {
```
erosdome, line 64: I think `checkRepoYmlSplashModalVisibility` indicates it will just do checks, not alter anything, whereas it does.

```suggestion
			function showRepoYmlSplashModalIfNecessary () {
```

source/javascripts/services/_requestService.js.erb
```diff
@@ -543,6 +543,45 @@ import jsyaml from "js-yaml";
...
+							new Error(
+								"<%= data[:strings][:request_service][:get_pipeline_config][:not_website_mode_error] %>"
+							)
+						);
+				}
+
+				return $q(function(resolve, reject) {
+					$http.get(requestURL, requestConfig).then(
+						function(response) {
+							resolve(response.data);
```
erosdome, line 570: Do a snake_case to camelCase transformation with the fields here. So from later on, the fetched data's keys look pretty.

source/stylesheets/_tab-workflows.scss.erb
```diff
@@ -619,6 +619,7 @@ header.sticky section.manage-workflows {
 
 								button.update-step {
 									$size: 14px;
+									border: 1px solid white;
```
erosdome, line 622: Use constant for white.

source/javascripts/controllers/_MainController.js.erb
Ozonid: it does the same in the background, i don't see a benefit of using a lodash method here

source/javascripts/components/SplashModal/SplashModal.scss
Ozonid: i'm not a big fan of using magic/global variables in CSS to tie together unrelated things especially with element sizing. the variable name is really non-descriptive

source/javascripts/components/SplashModal/SplashModal.scss
Ozonid: I'll use the BitKit globals instead `var(--color-white)`

source/javascripts/components/SplashModal/SplashModal.scss
erosdome: The grid is what all our elements need to be the multiple of, regarding width & height. There is probably a Bitkit global for this, it would be better to use thet, I wasn't aware it's accessible here.

source/javascripts/controllers/_MainController.js.erb
erosdome: Easier readability, I think it's worth it.

source/javascripts/services/_requestService.js.erb
erosdome: We've decided to later come up with a new way for this. For now, this alteration will happen, but later we will use some kind of Typescript type definition for these.

PR #503, opened 2020-08-04
[EN-178] Create request service in TypeScript

erosdome:


source/javascripts/services/request-service.ts
```diff
@@ -0,0 +1,158 @@
...
+			abortedPromise.then(() => {
+				controller.abort();
+			});
+		}
+
+		let requestURL = "";
+
+		switch (this.mode) {
+			case RequestServiceMode.Website:
+				requestURL = StringService.stringReplacedWithParameters(window["routes"].website.yml_get, {
```
Ozonid, line 46: Not sure if i agree with collecting all endpoint paths in a separate file. I don't think we ever want to reuse these string and it adds an extra layer of unnecessary complexity.

source/javascripts/services/request-service.ts
```diff
@@ -0,0 +1,158 @@
...
+			const error = this.errorFromResponseBody(
+				responseBody,
+				window["strings"].request_service.load_app_config.default_error
+			);
+			this.logErrorWithLevel(error, response.status < 500 ? StatusType.warn : StatusType.error);
+			throw error;
+		}
+
+		try {
+			return await response.text();
```
Ozonid, line 101: What does `.text()` do?

source/javascripts/services/request-service.ts
```diff
@@ -0,0 +1,158 @@
...
+		return new Error(responseBody.error || responseBody.error_msg || defaultMessage);
+	}
+
+	private logErrorWithLevel(error: Error, level: StatusType): void {
+		if (!this.logger) {
+			return;
+		}
+		const logger = (this.logger as unknown) as Logger;
+
+		switch (level) {
```
Ozonid, line 137: I think we could be simplified, only "error" differs from the others, so
`level===StatusType.error ? logger.error(error) : logger[level](error.message)`

source/javascripts/services/request-service.ts
```diff
@@ -0,0 +1,158 @@
...
+		this.configure({ logger });
+	}
+
+	public configure({ mode, appSlug, logger }: RequestServiceConfigureOptions = {}): void {
+		this.mode = mode || (this.modeFromEnvVars() as string);
+		this.appSlug = appSlug || (getAppSlug() as string);
+		this.logger = logger;
+	}
+
+	public async getAppConfigYML(
```
Ozonid, line 32: I think this method is way too long.

- I think we should create some kind of abstract wrapping for `fetch`. In my opinion this function should look something like
```
try {
let result = await abstractApiCall(url, params, etc.)
} catch (ex) {
// handle call-specific errors
}
// return or do some mapping
```

- Also we should "standardize" general errors (aborted, not found, 500 etc. - maybe we should create custom error classes - a thin layer over Error - so we can later check for them?) and only handle individual failure cases in the different API call methods.

source/javascripts/services/request-service.ts
```diff
@@ -0,0 +1,158 @@
...
+		const controller = new AbortController();
+		if (abortedPromise) {
+			abortedPromise.then(() => {
+				controller.abort();
+			});
+		}
+
+		let requestURL = "";
+
+		switch (this.mode) {
```
Ozonid, line 44: We'll only ever have 2 cases here, I think it's better to user a ternary for this

source/javascripts/services/request-service.ts
```diff
@@ -0,0 +1,158 @@
...
+
+		return RequestServiceMode.Cli;
+	}
+
+	private prefixedError(message: string, messagePrefix: string): Error {
+		return new Error(messagePrefix + message);
+	}
+
+	private requestAbortedError(messagePrefix: string): Error {
+		return this.prefixedError(window["strings"].request_service.request.aborted, messagePrefix);
```
Ozonid, line 121: I'd put the error string here instead of fetching it from a file, I don't think we want to use it outside of request service.

source/javascripts/services/request-service.ts
```diff
@@ -0,0 +1,158 @@
+import { getAppSlug } from "./app-service";
+import StringService from "./string-service";
+import { Logger } from "./logger";
+import { StatusType } from "@datadog/browser-logs";
+
+enum RequestServiceMode {
+	Website = "website",
```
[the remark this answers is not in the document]
erosdome: I think the recommended TypeScript naming convention is using PascalCase for that. For the values we could use all-caps strings, but it currently break existing functionality - the `requestService` fetches its `mode` from `RequestService`.

source/javascripts/services/request-service.ts
erosdome: I think we should do that in a separate iteration. Though this URL is not used anywhere else, so we could move it here, but I think it would result in additional confusion whenever someone checks the file containing the routes, and not seeing all routes in there.

source/javascripts/services/request-service.ts
erosdome: It reads the response body, assuming it is not JSON but a plain text - like in our case.

https://developer.mozilla.org/en-US/docs/Web/API/Body/text

source/javascripts/services/request-service.ts
erosdome: Though it would be simpler, I think it's dangerous: the fact that the `StatusType` enum has the same keys as the `Logger` interface's methods is coincidental. If we were to rename methods of the `Logger` interface - which we would assume to be a safe change - it would break this functionality.

source/javascripts/services/request-service.ts
erosdome: Same as with routes, I think we should do it in a separate iteration.

source/javascripts/services/request-service.ts
```diff
@@ -97,12 +91,7 @@ class RequestService {
 			throw error;
 		}
 
-		try {
-			return await response.text();
-		} catch (error) {
-			this.logErrorWithLevel(error, StatusType.error);
-			throw new Error(window["strings"].request_service.load_app_config.default_error);
-		}
+		return await this.convertResponseToText(response, window["strings"].request_service.load_app_config.default_error);
```
Ozonid, line 85: this await is not harmful here, but it doesn't do anything here, you can just simply return the promise here

source/javascripts/services/request-service.ts
```diff
@@ -113,12 +102,46 @@ class RequestService {
...
+		} catch (error) {
+			error = new Error(`${method} ${url} - ${error.message}`);
+			this.logErrorWithLevel(error, StatusType.warn);
+			throw error;
+		}
+
+		return response;
+	}
+
+	private async convertResponseToJson(response: Response, defaultErrorMessage: string): Promise<any> {
```
Ozonid, line 120: either remove async or make use of it inside the function body ;)

source/javascripts/services/request-service.ts
```diff
@@ -113,12 +102,46 @@ class RequestService {
...
+		return response.json().catch(error => {
+			this.logErrorWithLevel(error, StatusType.error);
+			throw new Error(defaultErrorMessage);
+		});
+	}
+
+	private async convertResponseToText(response: Response, defaultErrorMessage: string): Promise<string> {
```
Ozonid, line 127: either remove async or make use of it inside the function body ;)

PR #505, opened 2020-08-06
[EN-191] Save workflow to repo modal

Ozonid:


source/javascripts/controllers/_MainController.js.erb
```diff
@@ -338,16 +349,48 @@ import _ from "underscore";
...
 				}
-
+				
 				viewModel.saveProgress.start("<%= data[:strings][:main][:load_progress][:saving] %>");
 
+				if (appService.pipelineConfig.usesRepositoryYml && _.contains(appService.betaTags, 'repository-yml-usage')) {
+					viewModel.openUpdateYmlInRepositoryModal();
+					return;
+				}
+
```
erosdome, line 390: This should not be done on all tabs, e. g. not on secrets or stacks tab. Use the switch below.

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -338,16 +349,48 @@ import _ from "underscore";
 				}
 			};
 
+			viewModel.openUpdateYmlInRepositoryModal = function() {
+				var variables = appService.getCurrentVariables();
+				appService.validateVariables(variables)
+					.then(function() {
+							viewModel.showUpdateYmlInRepositoryModal = true;
+					})
+			}
```
erosdome, line 358: This should only be done in the Workflows/Env vars/Triggers tab. The bitrise.yml tab needs a separate one, which only has to show the modal, no validation is required. It might even fail, if the page is initially loaded with the bitrise.yml tab, because in that case there will be no `appService.appConfig`.

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -338,16 +349,48 @@ import _ from "underscore";
...
+				}, 0)
+			}
+
+			viewModel.onSaveToRepoComplete = function() {
+				$timeout(function() {
+					viewModel.closeUpdateYmlInRepositoryModal();
+				}, 2000)
+
+				$timeout(function() {
+					appService.afterAppConfigSaved();
```
erosdome, line 373: This should only be done in the Workflows/Env vars/Triggers tab. The bitrise.yml tab needs a separate method like this - see the success block of `appService.saveAppConfigYML`.

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -31,6 +31,7 @@ import _ from "underscore";
 			$scope.requestService = requestService;
 			$scope.osService = osService;
 
+			viewModel.showUpdateYmlInRepositoryModal = false;
```
erosdome, line 34: Just some nitpicking - call it like `viewModel.isUpdateYmlInRepositoryModalVisible`, so it indicates it's a boolean flag, not an action to call.

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -99,7 +104,13 @@ import _ from "underscore";
 												"<%= data[:strings][:triggers][:update_deprecated_popup][:details_1] %>",
 												"<%= replaced_string(data[:strings][:triggers][:update_deprecated_popup][:details_2], [data[:routes][:blog][:new_triggers]])%>"
 											])
-												.then(viewModel.save)
+												.then(function() {
+													if (!appService.pipelineConfig.usesRepositoryYml) {
```
erosdome, line 108: Remove the negating `!` and switch the bodies of the if-else, so it's easier to read.

PR #506, opened 2020-08-07
[EN-201] [EN-215] Finalize texts for repo yml

Ozonid:


source/javascripts/components/YmlStorageSettings/YmlStorageSettingsNotifications.tsx
```diff
@@ -0,0 +1,41 @@
...
+				text="valid syntax of the bitrise.yml file."
+				url="https://devcenter.bitrise.io/builds/bitrise-yml-online/"
+			/>
+		</Text>
+	</Notification>
+);
+
+export const GenericBackendError: FC<MonolithError> = ({ error_msg: errorMessage }: MonolithError) => (
+	<Notification type="alert">{errorMessage}</Notification>
+);
```
erosdome, line 41: If it's a generic backend error, I think it should be in a separate file as it's not specific to YML storage settings notifications.

source/javascripts/components/YmlStorageSettings/YmlStorageSettingsNotifications.tsx
```diff
@@ -0,0 +1,41 @@
...
+			the{" "}
+			<InlineLink
+				underline
+				color="red-4"
+				text="valid syntax of the bitrise.yml file."
+				url="https://devcenter.bitrise.io/builds/bitrise-yml-online/"
+			/>
+		</Text>
+	</Notification>
+);
```
erosdome, line 37: Most of these components just wrap an already simple component, not much functionality is encapsulated. Is it worth it instead of just using the encapsulated content in the parent component instead? It might be a valid design pattern though, not sure.

Ozonid: I was deliberating the same thing when i did these. My "pro" arguments were:

- we only have to maintain the look/text of these messages in one place
- make the components that use these a bit smaller (they have quite some logic in there already, sparing couple of lines makes it cleaner for me)

The first one to move here was `YmlInRepositoryInvalidError` as it has specific styling tied to it, then I decided to move the other "common" notifications as well.

I don't really have a strong preference here, we can remove these changes if you feel like it's unnecessary

source/javascripts/components/YmlStorageSettings/YmlStorageSettingsNotifications.tsx
Ozonid: Yep, this should be in the common components

source/javascripts/components/YmlStorageSettings/YmlStorageSettingsNotifications.tsx
erosdome: I can see both solutions being reasonable, we can keep it. I think this should be a part of a bigger discussion when we're planning the ideal structure & arrangement of the project later.

PR #508, opened 2020-08-12
[EN-207] Fix before-after workflow step selection bug

erosdome:


source/javascripts/services/workflows-selection-service.ts
```diff
@@ -78,7 +78,11 @@ export class WorkflowsSelectionService {
 	};
 
 	restoreSelection = (viewModel: WorkflowViewModel): void => {
-		this.rearrangeSelection(viewModel, this.findSelectedWorkflow(viewModel));
+		this.rearrangeSelection(
+			viewModel,
+			this.findSelectedWorkflow(viewModel),
+			this.store.lastEditedWorkflowID || undefined
```
Ozonid, line 84: Isn't this value already undefined if the `this.store.lastEditedWorkflowID` is falsy? 🤔

erosdome: Well, by type it expects string or undefined (related code: `editedId?: string`). And `this.store.lastEditedWorkflowID` is string or null, so that would be a type error.

PR #512, opened 2020-08-15
[EN-242] Extend validation error with backend message

Ozonid:
- [x] error handling on save modal
- [x] hide repo yml actions success notification after API call _NOTE: notification is hidden with a timer_
- [x] add backend error context to valdiation error
- [x] stop loading when repo.yml is invalid _NOTE: the cleanest solution seemed to be hiding the tab content if it happens_
- [x] replace splash modal header image

Extend validation error message
image

Add error notifications to save modal
image
image

Updated modal header
image


source/javascripts/components/common/notifications/CreatingYmlOnWebsiteProgress.tsx
```diff
@@ -0,0 +1,8 @@
+import React, { FC } from "react";
+import { Notification } from "@bitrise/bitkit";
+
+const CreatingYmlOnWebsiteProgress: FC = () => (
+	<Notification type="progress">Creating bitrise.yml on bitrise.io...</Notification>
```
gaborszakacs, line 5: Can we avoid these duplications? Seems like only the text is different in these components.

source/templates/_main.slim
```diff
@@ -54,7 +54,7 @@ main[ng-controller="MainController as mainCtrl" ng-init="mainCtrl.init()"]
 		p.progress-indicator[progress-model="mainCtrl.loadAppConfigProgress"]
 		p.progress-indicator[progress-model="mainCtrl.changeMenuProgress"]
 		p.progress-indicator[progress-model="mainCtrl.initAppProgress"]
-		.tab-content[ng-view ng-if="mainCtrl.initAppProgress.isIdle && !mainCtrl.menuProgress.isError && !mainCtrl.loadAppConfigProgress.isInProgress  && mainCtrl.changeMenuProgress.isIdle"]
+		.tab-content[ng-view ng-if="mainCtrl.currentMenu.id === 'yml' || (mainCtrl.initAppProgress.isIdle && !mainCtrl.menuProgress.isError && !mainCtrl.loadAppConfigProgress.isInProgress && !mainCtrl.loadAppConfigProgress.isError  && mainCtrl.changeMenuProgress.isIdle)"]
```
gaborszakacs, line 57: Why is this needed?

source/javascripts/components/common/notifications/CreatingYmlOnWebsiteProgress.tsx
Ozonid: Well the reason behind creating these components was to avoid having the same text in multiple places (in case we have to change the wording) and to make the React components using them a bit easier to read (most cases these notifications are used in ternary operators, for me it's easier to understand them if it's just one simple tag).

source/templates/_main.slim
Ozonid: When app is in repo mode and bitrise.yml is missing or invalid in the repo, the app had an error on top and a stuck loading indicator below. 
By adding `!mainCtrl.loadAppConfigProgress.isError` to this logic, the tab content won't show up if the appConfig has issues. However I wanted to keep the `yml` tap available to give the users an option to switch back to website mode even if the the said error is present.

PR #513, opened 2020-08-17
[EN-253] Add fallback in case last edited WF is not part of current chain

Ozonid:


source/javascripts/services/workflows-selection-service.ts
```diff
@@ -143,7 +143,12 @@ export class WorkflowsSelectionService {
 
 		// save it to the store
 		editedId = editedId || wf.id;
-		const editedIndex = viewModel.selectedWorkflowChain.findIndex(({ workflow }) => workflow.id === editedId);
+		let editedIndex = viewModel.selectedWorkflowChain.findIndex(({ workflow }) => workflow.id === editedId);
```
gaborszakacs, line 146: Is this something we should cover in E2E tests?

PR #516, opened 2020-08-25
[EN-264] Track local WFE usage with Freshpaint

Ozonid:


package.json
```diff
@@ -10,7 +10,7 @@
 		"test-ci": "npm run lint && npm run karma",
 		"test-watch": "onchange -ik -d 500 'source/javascripts/**/*.js.erb' -- npm test",
 		"build": "webpack --bail -p",
-		"build:website": "MODE=WEBSITE ANALYTICS=true HOTJAR=true FRESHPAINT=true npm run build",
+		"build:website": "MODE=WEBSITE ANALYTICS=true HOTJAR=true npm run build",
```
erosdome, line 13: What is the reason behind wanting to enable Freshpaint always unless opted out from it? I think it would be safer if you would need to opt in in every mode where you want to use it.

erosdome (changes requested):

package.json
Ozonid: It was already included for website mode, and the purpose of these changes is to track the local usage. As this covers all cases we have I decided to include it by default and provide a way to opt out of it.

erosdome: In that case we should set `DISABLE_FRESHPAINT=true` to `start:dev`, no?

Ozonid: If you check the package.json, start calls start:dev as well.

		"start:dev": "webpack-dev-server",
		"start": "concurrently \"npm:start:dev\" \"DEV=true ./_scripts/run_api.sh 4000\"",

I've added dotenv (js) to avoid having freshpaint on during development (in any mode), I could add a `.env.example` file so it's more visible. We could also add it to the docs

erosdome: No need for dotenv, as I think adding the env var to the script is simpler, and also avoids the issue when someone does not have an .env file with `DISABLE_FRESHPAINT=true`. Adding it to the script will make sure it disables it for everyone, without further env var configuration needed.

erosdome (approved):

PR #519, opened 2020-09-04
Show title as step output description #plugin

lpusok:
Step Output title is always filled, show that instead of summary/description. This is the first phase, the second being a similar behavior for step outputs as inputs (title showed always, summary or description showed below). See https://bitrise.atlassian.net/browse/PLAT-47.

### DIFF
--- a/source/templates/workflows.slim
+++ b/source/templates/workflows.slim
@@ -135,11 +135,11 @@
 										.parameter.key
 											input[value="{{ output | prettifiedVariableKey }}" readonly]
 										.parameter.description
-											p[ng-if="!output.shortestDescription()"] == data[:strings][:workflows][:outputs][:no_description]
-											div[ng-if="output.shortestDescription()"]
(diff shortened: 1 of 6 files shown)


source/templates/workflows.slim
```diff
@@ -135,11 +135,11 @@
 										.parameter.key
 											input[value="{{ output | prettifiedVariableKey }}" readonly]
 										.parameter.description
-											p[ng-if="!output.shortestDescription()"] == data[:strings][:workflows][:outputs][:no_description]
-											div[ng-if="output.shortestDescription()"]
+											p[ng-if="!output.title()"] == data[:strings][:workflows][:outputs][:no_description]
+											div[ng-if="output.title()"]
```
Ozonid, line 139: If title is always available, there's no real point in keeping the `ng-if`. However, I've seen weirder things happen, so it might be the best to keep it as a safety net... 🤔 It's your call wether to keep it or not, I'm fine with it either way.

Ozonid (approved):

PR #525, opened 2020-09-24
Add missing pipeline config fetching

Ozonid:


source/javascripts/controllers/_MainController.js.erb
```diff
@@ -74,6 +74,7 @@ import _ from "underscore";
 							var loadPromises = [appService.getAppConfig()];
 							if (requestService.mode == "website") {
 								loadPromises.push(Stack.getAll().then(appService.getStack));
+								loadPromises.push(appService.getPipelineConfig());
```
erosdome, line 77: As this is a website-only feature, it's ok that we only add it to the loading process in website mode.

erosdome (approved):

PR #529, opened 2020-10-07
Fix missing pipeline config in CLI mode #plugin

Ozonid:


source/javascripts/services/_appService.js.erb
```diff
@@ -38,6 +38,10 @@
 				betaTags: undefined
 			};
 
+			if (requestService.mode === "cli") {
+				appService.pipelineConfig = {}
+			}
+
```
erosdome, line 44: This implies that pipeline config makes sense in CLI environment, whereas it doesn't.

I think it would be better to wherever `appService.pipelineConfig` is used, prefix it with checking whether we're in website mode or not.

This current solution works for simple properties like `.usesRepositoryYml`, but would break, if we were to introduce a nested property & do checks for it, e. g.

`if (appService.pipelineConfig.repoSettings.provider)`

erosdome (changes requested):

source/javascripts/controllers/_YMLController.js.erb
```diff
@@ -10,7 +10,7 @@ import * as monaco from "monaco-editor/esm/vs/editor/editor.api.js";
 			var editor;
 
 			viewModel.appSlug = appService.appDetails.slug;
-			viewModel.shouldShowYmlStorageSettings = false;
+			viewModel.shouldShowYmlStorageSettings = requestService.isWebsiteMode();
```
erosdome, line 13: I think at this point this should be false, and only set it to (potentially) true after the asynchronous loading of the pipeline config.

Otherwise, if a user visits this tab on the website & the pipeline config is not yet loaded, this flag will be `true` for a short period, which makes `r-yml-storage-settings` appear with not yet initialized properties.

erosdome (changes requested):

erosdome (approved):

PR #538, opened 2020-11-04
[EN-243] Fix step delete issue

Ozonid:
step_delete

### DIFF
--- a/source/javascripts/controllers/_WorkflowsController.js.erb
+++ b/source/javascripts/controllers/_WorkflowsController.js.erb
@@ -707,17 +707,7 @@ import { safeDigest } from "../services/react-compat";
 			viewModel.deleteStepSelected = function() {
 				var index = _.indexOf(viewModel.editedWorkflow.steps, viewModel.selectedStep);
 				viewModel.editedWorkflow.steps.splice(index, 1);
-
-				if (index === 0) {
-					var workflowIndex = _.findIndex(viewModel.selectedWorkflowChain, function(item) {
-						return item.workflow.id === viewModel.editedWorkflow.id;
-					});
-					
-					viewModel.stepSelected(null);
-					viewModel.editWorkflowAtIndex(workflowIndex);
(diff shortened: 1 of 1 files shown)


source/javascripts/controllers/_WorkflowsController.js.erb
```diff
@@ -707,7 +708,16 @@ import { safeDigest } from "../services/react-compat";
 				var index = _.indexOf(viewModel.editedWorkflow.steps, viewModel.selectedStep);
 				viewModel.editedWorkflow.steps.splice(index, 1);
 
-				viewModel.stepSelected(null);
+				if (index === 0) {
+					var workflowIndex = _.findIndex(viewModel.selectedWorkflowChain, function(item) {
+						return item.workflow.id === viewModel.editedWorkflow.id;
```
erosdome, line 713: This can be a different "instance" of the same workflow: let's say I have `primary` workflow with `setup` workflow added **twice** as before-workflows, and I delete the first step from the second `setup` workflow. This code will start to edit the previous `setup` workflow, not the one I deleted the step from.

Why is it even needed to update the edited workflow? Afaik it is a valid scenario that a step gets deselected while the workflow still remains edited. In this case we will be seeing the workflow's description in the main content area, instead of step details.

erosdome (changes requested):

source/javascripts/controllers/_WorkflowsController.js.erb
Ozonid: True, I forgot that you can have a workflow in the chain multiple times.

When the user deletes a step and there are steps left in the workflow, we jump to the previous one. But when there are no more steps in there, I wanted to show the "description screen" (when you click on the workflow's label in the left side of the workflows tab).

Also I did this before fixing the other method, I'll check how it behaves now with `viewModel.stepSelected(null);`

erosdome (approved):

PR #539, opened 2020-11-05
[EN-331] Add pipelines to trigger target dropdown

Ozonid:
pipeline_trigger

### DIFF
--- a/source/javascripts/components/_Trigger.js.erb
+++ b/source/javascripts/components/_Trigger.js.erb
@@ -45,19 +45,33 @@
 				return "tag";
 			}
 		};
+
+		Trigger.prototype.target = function(target) {
+			if (!target) {
+				return this.targetType() + "#" + this.targetValue()
+			}
+
+			var parts = target.split("#");
+			this.targetType(parts[0]);
+			this.targetValue(parts[1]);
+		}
 		
-		Trigger.prototype.setTarget = function(target) {
-			delete this.triggerConfig.workflow;
-			delete this.triggerConfig.pipeline;
-			this.triggerConfig[target.type] = target.value;
+		Trigger.prototype.targetValue = function(targetValue) {
+			if (!targetValue) {
+				return this.triggerConfig[this.targetType()];
+			}
+
+			this.triggerConfig[this.targetType()] = targetValue;
 		};
 
-		Trigger.prototype.getTarget = function(target) {
-			var type = this.triggerConfig.workflow ? "workflow" : "pipeline"
-			return {
-				type: type,
-				value: this.triggerConfig[type]
+		Trigger.prototype.targetType = function(targetType) {
+			if (!targetType) {
+				return this.triggerConfig.hasOwnProperty("workflow") ? "workflow" : "pipeline";
 			}
+
+			var target = this.targetValue();
+			delete this.triggerConfig[this.targetType()];
+			this.triggerConfig[targetType] = target;
 		};
 
 		Trigger.prototype.pushBranchPattern = function(pushBranchPattern) {
--- a/source/javascripts/components/_Trigger.js.erb
+++ b/source/javascripts/components/_Trigger.js.erb
@@ -48,20 +48,20 @@
 
 		Trigger.prototype.target = function(target) {
 			if (!target) {
-				return this.targetType() + "#" + this.targetValue()
+				return this.targetType() + "#" + this.targetId()
 			}
 
 			var parts = target.split("#");
 			this.targetType(parts[0]);
-			this.targetValue(parts[1]);
+			this.targetId(parts[1]);
 		}
 		
-		Trigger.prototype.targetValue = function(targetValue) {
-			if (!targetValue) {
+		Trigger.prototype.targetId = function(targetId) {
+			if (!targetId) {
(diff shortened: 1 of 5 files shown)


erosdome (changes requested):

source/javascripts/components/_Trigger.js.erb
```diff
@@ -45,13 +45,19 @@
...
+		};
 
-		Trigger.prototype.workflowID = function(workflowID) {
-			if (workflowID !== undefined) {
-				this.triggerConfig.workflow = workflowID;
+		Trigger.prototype.getTarget = function(target) {
+			var type = this.triggerConfig.workflow ? "workflow" : "pipeline"
+			return {
+				type: type,
+				value: this.triggerConfig[type]
```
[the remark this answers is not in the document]
Ozonid: I'm not really a fan of that pattern, using the same method for two completely different purposes is rather confusing for me. Also setting the type itself is rather tricky as it needs a value at the same time. And we need to know where to set the value, so basically both of them need the other half to be viable.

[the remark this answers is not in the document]
erosdome: Does not seem that complex to me, they do use each other but I don't think that's a problem:

`.target`: if it's a setter, do `this.triggerConfig[this.targetType()] = target`; and for both setter & getter cases, return `this.triggerConfig[this.targetType()]`.

`.targetType`: if it's a setter, do `var target = this.triggerConfig.target()` then `delete this.triggerConfig[this.targetType()]`, finally `this.triggerConfig[targetType] = target`; and for both setter & getter cases, return either `workflow` or `pipeline`, whichever is defined as key in `this.triggerConfig`.

My problem is merging the two just as "target" is that now it is a non-primitive that is not documented anywhere (meaning `{type: string, value: string}`), instead of having 2 separately handled string "properties" (I mean getter-setters) for them. Also, whenever you want to set the type or the value, you need to define both of them, whereas if you use separate methods for them here, only one of them is needed.

As for the pattern, it tries to mimic how a regular property would behave: you can either get its value, or assign one to it. We use it throughout this project in case of all the other objects like `Step`, `Variable`, etc.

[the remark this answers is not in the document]
Ozonid: I still think it's a horrible pattern, but I don't want to introduce inconsistencies with the other parts of the application. Maybe we could revisit the topic once we start migrating these parts to TS! :)

source/javascripts/components/_Trigger.js.erb
```diff
@@ -46,12 +46,32 @@
...
+				return this.targetType() + "#" + this.targetValue()
 			}
 
-			return this.triggerConfig.workflow;
+			var parts = target.split("#");
+			this.targetType(parts[0]);
+			this.targetValue(parts[1]);
+		}
+		
+		Trigger.prototype.targetValue = function(targetValue) {
```
erosdome, line 59: One more nitpicking, I'd suggest using **target ID** instead of **target value** everywhere.

source/javascripts/controllers/_TriggersController.js.erb
```diff
@@ -75,12 +75,27 @@
...
-					aWorkflowConfig,
-					workflowID
-				) {
-					return new Workflow(workflowID, aWorkflowConfig);
-				});
+				viewModel.targets = Object.keys(appService.appConfig.workflows)
+					.map(function(workflowId) {
+						return { value: "workflow#"+ workflowId, groupLabel: "Workflows", label: workflowId, disabled: workflowId[0] === "_" };
+					})
+					.sort(function(a,b) {
```
erosdome, line 82:
```suggestion
					.sort(function(a, b) {
```

source/javascripts/controllers/_TriggersController.js.erb
```diff
@@ -75,12 +75,27 @@
...
+					.sort(function(a, b) {
+						return b.disabled ? -1 : 0;
+					});
+
+				if (appService.betaTags && appService.betaTags.indexOf("can-trigger-pipeline") > -1) {
+					var pipelines = Object.keys(appService.appConfig.pipelines)
+					.map(function(pipelineId) {
+						return { value: "pipeline#"+ pipelineId, groupLabel: "Pipelines", label: pipelineId };
+					})
+					.sort(function(a,b) {
```
erosdome, line 94:
```suggestion
					.sort(function(a, b) {
```

source/javascripts/controllers/_TriggersController.js.erb
```diff
@@ -190,10 +206,6 @@
 			};
 
 			viewModel.triggerIsInvalid = function() {
-				if (viewModel.editedTrigger.workflowID()[0] == "_") {
-					return true;
-				}
```
erosdome, line 195: This should still be checked, but with first doing a check if the workflow's ID is defined.

erosdome (changes requested):

source/javascripts/controllers/_TriggersController.js.erb
Ozonid: Well I made them unselectable from the dropdown, but leaving an extra check shouldn't hurt

erosdome (approved):

PR #549, opened 2020-11-26
Make workflow selector scrollable

Ozonid:


erosdome (changes requested):
We're not planning to make the dropdown scrollabel, but adding a filter field to it.

Ozonid:
> We're not planning to make the dropdown scrollable, but adding a filter field to it.

S

PR #550, opened 2020-11-27
[EN-380] Migrate workflow selector to React and add search input

Ozonid:
Screen_Recording_2020-11-27_at_16 00 12


source/javascripts/controllers/_WorkflowsController.js.erb
```diff
@@ -517,12 +519,17 @@ import { safeDigest } from "../services/react-compat";
 				);
 			};
 
-			viewModel.renameWorkflowConfirmed = function() {
-				var oldWorkflowID = viewModel.renamedWorkflow.id;
-				var newWorkflowID = viewModel.renamedWorkflowID;
+			viewModel.wrappedRenameWorkflowConfirmed = function(workflow, newWorkflowID) {
```
erosdome, line 522: I think you should either do the `$timeout` wrapping in the original method, as you do in `videModel.selectWorkflow`, or do a "wrapped" method for `viewModel.selectWorkflow` as well. Also, in the latter case I'd suggest different naming, like `confirmRenameWorkflowWithDigest`, `selectWorkflowWithDigest`.

erosdome (changes requested):

erosdome (approved):

PR #569, opened 2021-03-17
[EN-623] Fix version number display

Ozonid:


source/templates/_main.slim
```diff
@@ -58,7 +58,7 @@ main[ng-controller="MainController as mainCtrl" ng-init="mainCtrl.init()"]
 		p == data[:strings][:main][:footer][:editor_open_source]
 		== svg("heart")
 		p == replaced_string(data[:strings][:main][:footer][:check_out_source_code], [data[:routes][:other][:bitrise_workflow_editor_source_code]])
-		a.version[ng-if="requestService.mode == 'website'" href="#{ replaced_string(data[:routes][:other][:bitrise_workflow_editor_current_release], [ENV['version']]) }" target="_blank" rel="noreferrer noopener"] v#{ ENV['version'] }
```
gaborszakacs, line 61: Let's check if we use `data[:routes][:other][:bitrise_workflow_editor_current_release]` elsewhere and remove it if we don't.

gaborszakacs (approved):

PR #571, opened 2021-03-23
[EN-778] Post current open tab with app config updates

Ozonid:


source/javascripts/services/_appService.js.erb
```diff
@@ -209,7 +209,7 @@
 				appService.appConfig.trigger_map = updatedTriggerMap;
 			};
 
-			appService.saveAppConfig = function(userSavedYmlToRepositoryPromise, requestConfig) {
+			appService.saveAppConfig = function(userSavedYmlToRepositoryPromise, requestConfig, tabOpenDuringSave) {
```
erosdome, line 212: Please reorder the input args, the `requestConfig` should always be the last one, as it's the least likely defined on calls. The method is not called anywhere else aside from MainController, so there shouldn't be any issues with having a different 2nd arg as before.

source/javascripts/services/_requestService.js.erb
```diff
@@ -308,7 +308,7 @@ import jsyaml from "js-yaml";
 
 			// App config
 
-			requestService.postAppConfig = function(appConfig, requestConfig) {
+			requestService.postAppConfig = function(appConfig, requestConfig, tabOpenDuringSave) {
```
erosdome, line 311: Same - move `requestConfig` to be the last input arg.

erosdome (changes requested):

source/javascripts/services/_appService.js.erb
```diff
@@ -273,7 +273,7 @@
 
 				return validateVariables(variables)
 					.then(function() {
-						return userSavedYmlToRepositoryPromise || requestService.postAppConfig(appService.appConfig, requestConfig);
+						return userSavedYmlToRepositoryPromise || requestService.postAppConfig(appService.appConfig, requestConfig, tabOpenDuringSave);
```
erosdome, line 276:
```suggestion
						return userSavedYmlToRepositoryPromise || requestService.postAppConfig(appService.appConfig, tabOpenDuringSave, requestConfig);
```

erosdome (changes requested):

erosdome (approved):

PR #589, opened 2021-07-02
[PLANG-110] Add custom rum timing to WFE page loads

matenadasdi:


source/javascripts/controllers/_MainController.js.erb
```diff
@@ -255,6 +259,8 @@ import _ from "underscore";
 						}
 
 						viewModel.menuProgress.success();
+
+						datadogRumCustomTiming('wfe', 'topNavigation');
```
erosdome, line 263: After the workflows tab opens, there's still some additional loading of steps' data inside the tab itself. Don't we want to include that as well? (heads up: it might be complicated to do that, because in that case we need that tab's ctrl to be able to distinguish its initial loading from loadings that are done when the tab is revisited)

erosdome (approved):

erosdome (approved):

PR #602, opened 2021-12-15
[PLANG-556] Fix step list filtering

imrekel:
https://bitrise.atlassian.net/browse/PLANG-556

NOTE: the root cause seem to be an Algolia filtering issue: I debugged the code, if a project type filter goes out to Algolia, it doesn't return any steps.

This fix does not fix the root cause but skips the Algolia filtering. Even though this results in larger traffic, this has been already what the users were doing to tackle the issue: clicking the All button to get all steps first and the filter locally. UX wise I don't see a noticable difference in load times and such.

Fixing the Algolia filtering requires more investigation/time and I suggest fixing this ASAP as it's a serious user facing issue.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
-ruby 2.5.7
+ruby 2.6.9
 nodejs 10.20.1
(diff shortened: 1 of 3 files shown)


erosdome (dismissed):

sylank (approved):
LGTM 👍

PR #603, opened 2021-12-22
[PLANG-558] Add new step - search returns no results 

sylank:
When searching for a step the default is Project Type filter which does not always return results

When this happens customers tend to think there are no steps available

This begs the question when we find no results should be show something to the user?

"We didn't find any results for your search in ReactNative steps, would you like to search all steps? <click link to search all steps>"

or 

"We didn't find any results for your search in ReactNative steps, here are some additional results we found that are not React"

Steps to reproduce:
Using a ReactNative app
In the WFE click + to add a new step
Search for "device testing for android"

Actual result:
No results are shown 

Expected result:
Results are shown

image

image

### DIFF
--- a/source/javascripts/components/_AddStep.js.erb
+++ b/source/javascripts/components/_AddStep.js.erb
@@ -175,7 +175,7 @@ import { safeDigest } from "../services/react-compat";
 
 							// NOTE: quick fix for step filtering - it seems there is an issue with the Algolia query or service filetring steps by project type (returns 0 steps)
 							// this was originally stepSourceService.loadLatestOfficialSteps(appService.appDetails.projectTypeID) when filtering is enabled
-							loadPromises.push(stepSourceService.loadLatestOfficialSteps());							
+							loadPromises.push(stepSourceService.loadLatestOfficialSteps());
 						}
 
 						steps = [];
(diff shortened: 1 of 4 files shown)


source/templates/add_step_sidebar.slim
```diff
@@ -39,5 +39,11 @@
 					== svg("workflow/icon-add_step-show_all")
 					span == data[:strings][:workflows][:steps][:add][:show_all_steps]
 			ol.steps[ng-if="!addStepCtrl.isCategoryMode"]
+				div[ng-if="addStepCtrl.shouldFilterByProjectType && addStepCtrl.filteredSteps(category.steps).length == 0"]
+					p We didn't find any results for your search in {{ appService.appDetails.projectTypeID }} steps category, would you like to search all steps?
```
imrekel, line 43: I think you should create a string map for the project types. I'm not sure what's the whole list but for instance:

- ios -> iOS
- android -> Android
- react (or what it's called) -> React Native
... defaulting to the type id if not specified

Another option is to omit the category name from the message if this is too complicated.

imrekel (changes requested):

source/javascripts/components/_AddStep.js.erb
```diff
@@ -324,6 +324,35 @@ import { safeDigest } from "../services/react-compat";
...
+        					return "Cordova";
+        				case "ionic":
+        					return "Ionic";
+        				case "react-native":
+        					return "React Native";
+        				case "flutter":
+        					return "Flutter";
+        				case "other":
+        					return "Other";
+        			}
```
imrekel, line 353: please check if it's syntactically correct :)

```suggestion
                default:
                  return projectType;
        			}
```

imrekel (approved):

PR #610, opened 2022-02-07
introduce go modules

bejczib:


bitrise.yml
```diff
@@ -172,51 +125,35 @@ workflows:
...
-            # go test $GOLIST_WITHOUT_VENDOR
             go test ./...
     - script:
         title: Go Vet
         inputs:
         - content: |-
             #!/bin/bash
             set -ex
-            go vet $GOLIST_WITHOUT_VENDOR
+            go vet
```
gaborszakacs, line 141: I think we need to add `./...` here, to do it for all packages.

```suggestion
            go vet ./...
```

bitrise.yml
```diff
@@ -172,51 +125,35 @@ workflows:
...
     - script:
         title: Err check
         inputs:
         - content: |-
             #!/bin/bash
             set -ex
-            errcheck -asserts=true -blank=true $GOLIST_WITHOUT_VENDOR
+            errcheck -asserts=true -blank=true
```
gaborszakacs, line 148:
```suggestion
            errcheck -asserts=true -blank=true ./...
```

gaborszakacs (changes requested):
Thanks for taking the time!

PR #646, opened 2022-02-09
fix: Rename workflow recipe link ids

zoltanszabo-bitrise:


source/javascripts/_componentRegister.js
```diff
@@ -48,7 +48,7 @@ angular
 		"rWorkflowSelector",
 		register(WorkflowSelector, ["selectedWorkflow", "workflows", "selectWorkflow", "renameWorkflowConfirmed"])
 	)
-	.component("rWorkflowRecipesLink", register(WorkflowRecipesLink, ["id"]))
+	.component("rWorkflowRecipesLink", register(WorkflowRecipesLink, ["linkId"]))
```
simonmarton, line 51: Is this used anywhere from angular?

simonmarton (approved):

source/javascripts/_componentRegister.js
zoltanszabo-bitrise: It might not be, it may be just muscle memory 😄

PR #666, opened 2022-05-13
Fix cropped credit cost when machine type name is long

gsanta:
jira: https://bitrise.atlassian.net/browse/PLANG-1624


source/stylesheets/_tab-stack.scss.erb
```diff
@@ -192,11 +192,15 @@
...
 							width: 100%;
+							flex-wrap: wrap;
+							justify-content: space-between;
 
 							.icon-and-name-and-note {
 								display: flex;
 								align-items: center;
 								margin-right: 12px;
+								min-width: 0;
+								text-owerflow: ellipsis;
```
danieldecsi-bitrise, line 203:
```suggestion
								text-overflow: ellipsis;
```

source/stylesheets/_tab-stack.scss.erb
```diff
@@ -192,11 +192,15 @@
 							display: flex;
```
danieldecsi-bitrise, line 192: This is another issue I just noticed. If the machine types wrap into multiple lines, there are no distance between the rows:

image

I think we could quickly fix this by:
```
.machine-types {
  ...
  gap: 18px;
}
```

and `margin-right: 18px` should be removed from `.machine-type-wrapper` at the same time.

danieldecsi-bitrise (changes requested):

danieldecsi-bitrise (approved):

PR #683, opened 2022-07-22
Update dependencies

doomsayer13:


package.json
```diff
@@ -36,7 +36,7 @@
 	},
 	"homepage": "https://github.com/bitrise-io/bitrise-workflow-editor#readme",
 	"dependencies": {
-		"@bitrise/bitkit": "^9.0.0",
```
erosdome, line 39: Why shouldn't we allow to be updated to e. g. 9.0.1?

package.json
```diff
@@ -9,7 +9,7 @@
 		"test": "npm run build && npm run karma",
 		"test-ci": "npm run lint && npm run karma",
 		"test-watch": "onchange -ik -d 500 'source/javascripts/**/*.js.erb' -- npm test",
-		"build": "webpack --bail -p",
```
erosdome, line 12: Is `-p` deprecated or some other reason why we should drop it?

webpack.config.js
```diff
@@ -149,7 +159,7 @@ module.exports = {
 			},
 
 			{
-				test: /\.scss(\.erb)?$/,
+				test: /\.s[ac]ss(\.erb)?$/i,
```
erosdome, line 162: I think we don't have .sass files, do we need to support those as well?

erosdome (changes requested):

package.json
doomsayer13: Because there are breaking changes in the 9+ versions, didn't checked in which one, but if we want to update bitkit then we should  probably go through all the *.tsx files, there was a lot of issues.

package.json
doomsayer13: Yes, it's deprecated

webpack.config.js
doomsayer13: I just copy pasted it from the documentation, but I think I can change it back to the original one.

package.json
erosdome: Because semantic versioning might sometimes be off, resulting in minor breaking changes within 9.x.x

package.json
erosdome: It's deprecated

webpack.config.js
erosdome: It's the recommendation, we might not need it but it doesn't hurt.

erosdome (approved):

PR #690, opened 2022-08-31
Added Segment library, along with tracking event for Changing Machine…

cdhenley219-bit:
This is part of https://bitrise.atlassian.net/browse/PLG-180


roland-spekker-bitrise (dismissed):

roland-spekker-bitrise:
LGTM but I have zero experience with the workflow editor

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -488,9 +489,21 @@ import datadogRumCustomTiming from "../utils/datadogCustomRumTiming.ts";
...
 									return appService
 										.saveStackAndDockerImage(deferredSavePromise, viewModel.currentMenu.id)
 										.then(function() {
+						
+											if (machineTypeUpdated) {
+												const event = 'Machine Type Changed - Stacks & Machines Tab';
+												window.analytics && window.analytics.track(event, {
+												  machine_type_id: viewModel.dataToSave.meta['bitrise.io'].machine_type_id,
+												  user_slug: viewModel.currentUser.slug,
+												  organization_slug: appService.appDetails.ownerData.type === 'Organization' ? appService.appDetails.ownerData.slug: undefined
```
gsanta, line 503: A space is missing before the `:` in the ternary :)

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -488,9 +489,21 @@ import datadogRumCustomTiming from "../utils/datadogCustomRumTiming.ts";
...
+
 									return appService
 										.saveStackAndDockerImage(deferredSavePromise, viewModel.currentMenu.id)
 										.then(function() {
+						
+											if (machineTypeUpdated) {
+												const event = 'Machine Type Changed - Stacks & Machines Tab';
+												window.analytics && window.analytics.track(event, {
+												  machine_type_id: viewModel.dataToSave.meta['bitrise.io'].machine_type_id,
+												  user_slug: viewModel.currentUser.slug,
```
gsanta, line 502: I think `currentUser` can be undefined if `getCurrentUserData` fails, maybe we should use it like this:`viewModel.currentUser?.slug`

gsanta (changes requested):

package.json
```diff
@@ -40,6 +40,7 @@
 		"@bitrise/steplib-search": "^2.3.0",
 		"@datadog/browser-logs": "^1.7.0",
 		"@datadog/browser-rum": "^2.6.1",
+		"@types/segment-analytics": "0.0.34",
```
gsanta, line 43: Do we need the type definition file? Only js files were modified so we are not using any types.
(But for me it can stay here, just asking)

source/javascripts/controllers/_MainController.js.erb
gsanta: It's completely blank here for me, but if you go to the Files Changed tab it shows the correct line :)

package.json
cdhenley219-bit: I went with the TypeScript version based on the docs I saw on WFE gradually moving to React and TypeScript. In the Monolith, we use Segment in both JS and TS files, so I assumed that may happen here as well.

source/javascripts/controllers/_MainController.js.erb
cdhenley219-bit: This has been updated

source/javascripts/controllers/_MainController.js.erb
cdhenley219-bit: This has been updated

gsanta (approved):

PR #694, opened 2022-09-08
PLG-226 - WFE Product tour

roland-spekker-bitrise:
JIRA Link: https://bitrise.atlassian.net/browse/PLG-226

### DIFF
--- a/source/javascripts/components/ProductTour/Highlighter.scss
+++ b/source/javascripts/components/ProductTour/Highlighter.scss
@@ -1,0 +1,32 @@
+.highlighter {
+    width: 100vw;
+    height: 100vh;
+    position: fixed;
+    top: 0;
+    left: 0;
+    z-index: 100;
+    pointer-events: none;
+
+    > * {
+      pointer-events: all;
+    }
+
+    .highlighter__backdrop-mask {
+      width: 100vw;
+      height: 100vh;
+      background-color: rgba(0, 0, 0, 0.2);
+      position: absolute;
+      top: 0;
+      left: 0;
+      pointer-events: none;
+    }
+
+    .highlighter__selection {
+      position: absolute;
+      background-color: transparent;
+      border-radius: 4px;
+      box-shadow: 0 0 0 3px rgba(0, 123, 255, 1);
+      z-index: 200;
+    }
+}
+
--- a/source/javascripts/components/ProductTour/Highlighter.scss
+++ b/source/javascripts/components/ProductTour/Highlighter.scss
@@ -5,7 +5,6 @@
     top: 0;
     left: 0;
     z-index: 100;
-    pointer-events: none;
 
     > * {
       pointer-events: all;
@@ -18,14 +17,13 @@
       position: absolute;
       top: 0;
       left: 0;
-      pointer-events: none;
     }
 
     .highlighter__selection {
       position: absolute;
       background-color: transparent;
       border-radius: 4px;
-      box-shadow: 0 0 0 3px rgba(0, 123, 255, 1);
+      box-shadow: 0 0 0 3px rgba(151, 71, 255, 1);
       z-index: 200;
     }
 }
--- a/source/javascripts/components/ProductTour/Highlighter.tsx
+++ b/source/javascripts/components/ProductTour/Highlighter.tsx
@@ -14,7 +14,6 @@ interface HighlighterProps {
 
 export const useHighlighter = (selectedId: string) => {
 	const ref = useRef<HTMLElement | null>(null);
-	const animationFrameRef = useRef<number | undefined>();
 	const [highlightedRect, setHighlightedRect] = useState<DOMRect | null>(null);
 
 	const highlightElement = useCallback(id => {
@@ -35,15 +34,8 @@ export const useHighlighter = (selectedId: string) => {
 	useEffect(() => {
 		const onResize = () => {
 			if (ref.current) {
-				if (animationFrameRef.current) {
-					cancelAnimationFrame(animationFrameRef.current);
-				}
-
-				animationFrameRef.current = requestAnimationFrame(() => {
-					const rect = ref.current!.getBoundingClientRect();
-					setHighlightedRect(rect);
-					animationFrameRef.current = undefined;
-				});
+				const rect = ref.current!.getBoundingClientRect();
+				setHighlightedRect(rect);
 			}
 		};
 
--- a/source/javascripts/components/ProductTour/Highlighter.tsx
+++ b/source/javascripts/components/ProductTour/Highlighter.tsx
@@ -12,7 +12,7 @@ interface HighlighterProps {
 	renderSelection?: () => JSX.Element;
 }
 
-export const useHighlighter = (selectedId: string) => {
+export const useHighlighter = (selectedId: string | undefined) => {
 	const ref = useRef<HTMLElement | null>(null);
 	const [highlightedRect, setHighlightedRect] = useState<DOMRect | null>(null);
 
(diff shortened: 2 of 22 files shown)


source/javascripts/components/ProductTour/Highlighter.scss
```diff
@@ -0,0 +1,30 @@
+.highlighter {
```
cdhenley219-bit, line 1: The styles look good here, but shouldn't this be placed in BitKit? I thought that was the approach to take with new styles/UI components.

roland-spekker-bitrise: The bitkit version is pre chakra based. Mate already contacted the platform team if they could upgrade and migrate the components to the latest bitkit version.

When the upgrade happens it will simplify a lot of things, like using the included popover instead of manually creating it and positioning it like in here. I don't know when exactly they can pick up this task, so I went ahead with it this way. 

I just didn't want to move everything into style props and that's why there are css files. My thought process was that these will be refactored when the chakra powered bitkit will arrive.

source/javascripts/components/ProductTour/Highlighter.tsx
```diff
@@ -0,0 +1,110 @@
...
+	}, [selectedId, highlightElement]);
+
+	useEffect(() => {
+		const onResize = () => {
+			if (ref.current) {
+				if (animationFrameRef.current) {
+					cancelAnimationFrame(animationFrameRef.current);
+				}
+
+				animationFrameRef.current = requestAnimationFrame(() => {
```
gsanta, line 42: Why is `requestAnimationFrame` used here?
`setHighlightedRect` will be used only in the next render cycle of react, so does it add anything plus compared to using it without `requestAnimationFrame`?

source/javascripts/components/ProductTour/ProductTour.tsx
```diff
@@ -0,0 +1,97 @@
...
+		const filtered = tips.filter(tip => menuIds.includes(tip.id));
+		setValidTips(filtered);
+		setSelectedId(filtered[0].id);
+	}, []);
+
+	// NOTE: Angular passes the menu items ids it will render,
+	// but we stil need to wait for the templates to render them before we can move on.
+	useWaitForElements(menuIds, onFound);
+
+	const { rect, clipPath, highlighted } = useHighlighter(selectedId);
```
gsanta, line 26: For me it would be more understandable if `useHighlighter` would be called from the `Highlighter` component, the only thing used here is the `highlighted` property. Wouldn't it be better to have a callback on the `Highlighter` component something like `onHighlighted` and then this hook could be encapsulated within the `Highlighter`

source/javascripts/components/ProductTour/tips.ts
```diff
@@ -0,0 +1,47 @@
...
+		id: "secrets",
+		title: "Secrets",
+		description: "Find out how to set up secrets, how to edit them, expose them to pull requests and more.",
+		link: "https://devcenter.bitrise.io/en/builds/secrets.html"
+	},
+	{
+		id: "env-vars",
+		title: "Env Vars",
+		description:
+			"Here you the define environment variables that the steps need to access in a build. " +
```
gsanta, line 28: Typo in: `Here you the define environment`

gsanta (changes requested):

source/javascripts/components/ProductTour/Highlighter.tsx
roland-spekker-bitrise: Completely agree, I was investigating some lagging/frame drops and I think I've left it in there.

source/javascripts/components/ProductTour/ProductTour.tsx
```diff
@@ -0,0 +1,97 @@
...
+		onTrackClose();
+	};
+
+	useEffect(() => {
+		if (isOpen && !!selectedId && highlighted) {
+			onDisplayTooltip();
+		}
+	}, [isOpen, highlighted, onDisplayTooltip, selectedId]);
+
+	const onChange = useCallback(id => {
```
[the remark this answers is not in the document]
roland-spekker-bitrise: It is currently, because of this: 
```
useEffect(() => {
		if (selectedId !== undefined) {
			onChange(selectedId);
		}
	}, [selectedId, onChange]);
```
But I don't like it this way so I'll take another look

source/javascripts/components/ProductTour/ProductTooltip.tsx
```diff
@@ -0,0 +1,185 @@
...
+						finished
+					};
+				}
+			}
+			break;
+		}
+		case "prev": {
+			const { selectedIndex, items } = state;
+			if (selectedIndex !== undefined) {
+				const prevIndex = Math.max(selectedIndex - 1, 0);
```
simonmarton, line 64: I think this can be simplified a bit
```suggestion
			if (selectedIndex) {
				const prevIndex = selectedIndex - 1;
```

source/javascripts/components/ProductTour/ProductTour.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+	const onFound = useCallback(() => {
+		const filtered = tips.filter(tip => menuIds.includes(tip.id));
+		setValidTips(filtered);
+		setSelectedId(filtered[0].id);
+	}, []);
+
+	// NOTE: Angular passes the menu items ids it will render,
+	// but we stil need to wait for the templates to render them before we can move on.
+	useWaitForElements(menuIds, onFound);
```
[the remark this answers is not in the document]
simonmarton: Or `filtered` can check agains  the callback's `foundElements` array

simonmarton (commented):
This is minor, but files have different indentation, please use 2 spaces in any new files

source/javascripts/components/ProductTour/tips.ts
roland-spekker-bitrise: I used the copy from the ticket and asked Hunter if this is what he wanted and the response is: 

"It was actually supposed to be as you can see in the screenshot, i.e. this is where you do this rather than this is where you can do it. I guess it depends on how often env vars are needed for a build to run successfully. If it's something you might or might not do, 'can' would work better. If it's something you will (almost) always do, it's better without 'can'."

So I'll leave the copy as is as this is what was intended.

source/javascripts/components/ProductTour/ProductTour.tsx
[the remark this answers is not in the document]
roland-spekker-bitrise: I've refactored it and lifted up the state to `ProductTour` to not have the same state - who is selected - be replicated in  the 2 places. This solves this super bad syncing useEffect too.

source/javascripts/components/ProductTour/ProductTour.tsx
roland-spekker-bitrise: Yes I could've went with something like this 
```tsx
<Highlighter>{(rect) => {<ProductTooltip />}}</Highlighter>
```
This would mean the `Highlighter` component fully encapsulates the flow of highlighting. 

But I've opted to slice it in a different way as the only thing they need to share technically is the `DOMRect` data. So technically we could have different combinations of components/hooks for visualising the highlight. I think with me moving out the hook from `Highlighter` makes it more obvious.

Wdyt? Is this acceptable?

source/javascripts/components/ProductTour/ProductTour.tsx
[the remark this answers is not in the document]
roland-spekker-bitrise: I've refactored it to use the found elements array

simonmarton (approved):

PR #695, opened 2022-09-13
feat: use dedicated product tour check endpoint - PLG-226

roland-spekker-bitrise:


source/javascripts/controllers/_MainController.js.erb
```diff
@@ -664,8 +664,11 @@ import datadogRumCustomTiming from "../utils/datadogCustomRumTiming.ts";
 
 				appService.getAppConfig();
 
-				requestService.getCurrentUserData().then(function(user) {
-					viewModel.currentUser = user;
+				Promise.all([
```
simonmarton, line 667: I couldn't find the difference, I guess this is mostly just to match prev implementations 🤷 
```suggestion
				$q.all([
```

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -664,8 +664,11 @@ import datadogRumCustomTiming from "../utils/datadogCustomRumTiming.ts";
...
+					requestService.getCurrentUserData(),
+					requestService.getProductTourShown(),
```
avivenzio-bitrise, line 669: Assuming the product tour is a non critical feature. Does it make sense to make these requests sequentially?
My concern is that in the event the product tour requests fails this promise will reject and never set the `viewModel.currentUser` which I'm assuming is used throughout the page.

The upside by making the requests sequentially is that if the product tour request fails, the user should still be able to have the same experience they do today.

roland-spekker-bitrise: I was kind of on the fence about putting it here, but I didn't want to change the component API too for this. I didn't want to delay the release but I guess you are right. I'll decouple the two completely.

simonmarton (approved):

PR #696, opened 2022-09-26
add beta warning for m1 max elite xl machines

avivenzio-bitrise:
Jira: https://bitrise.atlassian.net/browse/PLG-295

Add beta warning for M1 Max Elite XL machines.

### DIFF
--- a/rails/strings.rb
+++ b/rails/strings.rb
@@ -696,7 +696,7 @@ def strings
                     gen1: "The tried and tested machines that have been serving our customers for many years now",
                     gen2: "These machines run on a new hardware generation, offering improved performance compared to Gen1 hardware"
                 },
-                beta_machine_warning: "<machineName> machines are in beta. Some features (e.g. the Android emulator) are not supported yet. <a href='<url>' target='_blank'>Learn more</a>"
+                apple_silicon_beta_warning: "<machineName> machines are in beta. Some features (e.g. the Android emulator) are not supported yet. <a href='<url>' target='_blank'>Learn more</a>"
             },
             machine_type: {
                 workflow_specific_settings: {
--- a/rails/strings.rb
+++ b/rails/strings.rb
@@ -696,8 +696,7 @@ def strings
                     gen1: "The tried and tested machines that have been serving our customers for many years now",
                     gen2: "These machines run on a new hardware generation, offering improved performance compared to Gen1 hardware"
                 },
-                m1_elite_xl_beta_warning: "M1 Elite XL machines are in beta. Some features (e.g. the Android emulator) are not supported yet. <a href='<url>' target='_blank'>Learn more</a>",
-                m1_max_elite_xl_beta_warning: "M1 Max Elite XL machines are in beta. Some features (e.g. the Android emulator) are not supported yet. <a href='<url>' target='_blank'>Learn more</a>"
+                beta_machine_warning: "<machineName> machines are in beta. Some features (e.g. the Android emulator) are not supported yet. <a href='<url>' target='_blank'>Learn more</a>"
             },
             machine_type: {
                 workflow_specific_settings: {
--- a/rails/strings.rb
+++ b/rails/strings.rb
@@ -696,7 +696,8 @@ def strings
                     gen1: "The tried and tested machines that have been serving our customers for many years now",
                     gen2: "These machines run on a new hardware generation, offering improved performance compared to Gen1 hardware"
                 },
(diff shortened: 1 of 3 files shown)


rails/strings.rb
```diff
@@ -696,7 +696,8 @@ def strings
                     gen1: "The tried and tested machines that have been serving our customers for many years now",
                     gen2: "These machines run on a new hardware generation, offering improved performance compared to Gen1 hardware"
                 },
-                m1_elite_xl_beta_warning: "M1 Elite XL machines are in beta. Some features (e.g. the Android emulator) are not supported yet. <a href='<url>' target='_blank'>Learn more</a>"
+                m1_elite_xl_beta_warning: "M1 Elite XL machines are in beta. Some features (e.g. the Android emulator) are not supported yet. <a href='<url>' target='_blank'>Learn more</a>",
+                m1_max_elite_xl_beta_warning: "M1 Max Elite XL machines are in beta. Some features (e.g. the Android emulator) are not supported yet. <a href='<url>' target='_blank'>Learn more</a>"
```
jakerobb, line 700: Is there a way we could do something like:

```
apple_silicon_beta_warning: "%{machine_type_name} machines are in beta. Some features (e.g. the Android emulator) are not supported yet. <a href='<url>' target='_blank'>Learn more</a>",
apple_silicon_machine_types: [g2-m1.8core, g2-m1-max.5core],
```
I'm saying "apple silicon" because M2 is out and it's only a matter of time before we have an M2 offering. It's likely that we'll have a growing variety of these over time, and there's no saying how long this beta period will last, so I'm thinking we could generalize and avoid some code duplication.

Then, in the Angular code, provide the machine type name as a parameter to the actual warning message, and display it if the machine type is on the list -- replacing both existing invocations of this warning.

The desired end result: the next time we need to add this beta warning to another machine type, you just add the ID to the array and you're done! :D

source/templates/stack.slim
```diff
@@ -108,3 +109,4 @@
 							/ a.machine-type-info-link[ng-href="{{ cpuDescriptionURL }}" target="_blank"] == data[:strings][:machine_type][:learn_more_about]
 					.machine-type-warning-wrapper
 						p.notification.warning[ng-if="stackCtrl.workflowMachineType(workflow).id === 'g2-m1.8core'"] == replaced_string(data[:strings][:stack][:m1_elite_xl_beta_warning], [data[:routes][:devcenter][:m1_stacks]])
+						p.notification.warning[ng-if="stackCtrl.workflowMachineType(workflow).id === 'g2-m1-max.5core'"] == replaced_string(data[:strings][:stack][:m1_max_elite_xl_beta_warning], [data[:routes][:devcenter][:m1_stacks]])
```
jakerobb, line 112: Unrelated to your PR, it would be nice to factor the common bits out from the default and workflow-specific settings. :)

jakerobb (commented):
These changes look good to me. I left a suggestion, but it's not something we really need to do right now; if it's hard, feel free to ignore. If we add the same warning for a third machine type, I'll feel more strongly about it.

I'll leave actual approval to someone with more FE expertise.

rails/strings.rb
avivenzio-bitrise: I also thought about something like this and didn't move forward because it was 2 and not 3. But I think your point regarding M2 is valid, let me take a look this afternoon

source/javascripts/controllers/_StackController.js.erb
```diff
@@ -26,6 +26,7 @@
 				"gen1": "<%= data[:strings][:stack][:machine_type_notes][:gen1] %>",
 				"gen2": "<%= data[:strings][:stack][:machine_type_notes][:gen2] %>"
 			};
+			viewModel.betaMachineTypeIds = ['g2-m1.8core', 'g2-m1-max.5core'];
```
jakerobb, line 29: The beta warning in question is specific to Apple Silicon. I'd rename a few things here -- `appleSiliconMachineTypeIds`, `isAppleSiliconMachineType`, `apple_silicon_beta_warning`, etc.

jakerobb (commented):
Looks good! Just one suggestion re: naming.

source/javascripts/controllers/_StackController.js.erb
avivenzio-bitrise: @jakerobb updated

cdhenley219-bit:
This looks good to me, but will defer to @gsanta for final approval.

cdhenley219-bit (approved):

roland-spekker-bitrise (approved):

gsanta (approved):

PR #698, opened 2022-10-05
Fix stack description url [SWAT-716]

doomsayer13:
JIRA ticket: https://bitrise.atlassian.net/browse/SWAT-716

### DIFF
--- a/source/javascripts/controllers/_StackController.js.erb
+++ b/source/javascripts/controllers/_StackController.js.erb
@@ -164,9 +164,7 @@
 					return stack.descriptionURLGen2;
 				}
 
-				if (stack.descriptionURL) {
-					return stack.descriptionURL;
-				}
+				return stack.descriptionURL;
 			};
 
 			viewModel.hasDescriptionURLForStack = function(stack) {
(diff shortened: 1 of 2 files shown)


source/javascripts/controllers/_StackController.js.erb
```diff
@@ -163,9 +163,16 @@
 				if (stack.descriptionURLGen2 !== undefined) {
 					return stack.descriptionURLGen2;
 				}
-				return stack.descriptionURL;
+
+				if (stack.descriptionURL) {
+					return stack.descriptionURL;
+				}
```
erosdome, line 169: If `stack.descriptionURL` is falsy (null, undefined, "") then the `viewModel.descriptionURLForStack` should be the same too. However, this solution always returns undefined, I think we shouldn't want that.

source/javascripts/controllers/_StackController.js.erb
```diff
@@ -163,9 +163,16 @@
...
 			};
 
+			viewModel.hasDescriptionURLForStack = function(stack) {
+				return stack.descriptionURLGen2 || stack.descriptionURL;
```
erosdome, line 173: `"a" || "b"` would return `"a"`, whereas a function called `hasDescriptionURLForStack` should return a boolean. Cast the result to that:
```suggestion
				return !!(stack.descriptionURLGen2 || stack.descriptionURL);
```

erosdome (changes requested):

erosdome (approved):

PR #713, opened 2022-10-19
Made changes to Workflow Editor Product Tour based on latest design

cdhenley219-bit:
Changes are for story https://bitrise.atlassian.net/browse/PLG-318


source/javascripts/components/ProductTour/Highlighter.tsx
```diff
@@ -48,15 +48,17 @@ export const Highlighter = ({
...
+							borderRadius="8"
+							border="10px solid white"
 							boxShadow="0 0 0 3px rgba(151, 71, 255, 1)"
 							zIndex="200"
 							style={{
-								left: `${rect?.x}px`,
-								top: `${rect?.y}px`,
-								width: `${rect?.width}px`,
-								height: `${rect?.height}px`
+								left: `${rect?.x - 10}px`,
```
avivenzio-bitrise, line 56: why do you need these offset values (`-10`) instead of passing in the desired value as it was in the previous implementation? Is it the addition of 10px border?

source/javascripts/components/ProductTour/Highlighter.tsx
```diff
@@ -48,15 +48,17 @@ export const Highlighter = ({
...
+								top: `${rect?.y - 10}px`,
+								width: `${rect?.width + 20}px`,
+								height: `${rect?.height + 20}px`
```
avivenzio-bitrise, line 59: same question as above

source/javascripts/components/ProductTour/ProductTooltip.tsx
```diff
@@ -38,17 +40,20 @@ export const ProductTooltip = ({
...
-			top={(rect?.y ?? 0) + (rect?.height ?? 0) + 20}
+			left={tip.position === "right" ? rect!.x + rect!.width + 37: rect!.width/2 - 210}
+			top={tip.position === "right" ? rect!.bottom/2  - 22: (rect?.y ?? 0) + (rect?.height ?? 0) + 40}
 			className="product-tooltip"
+			position="relative"
 		>
 			<Box display="flex" flexGrow="1">
-				<div className="arrow-up" style={{ left: rect!.width / 2 - 5 }} />
+				{ tip.position === "right" ? 
+					<div className="arrow-left" /> :
```
avivenzio-bitrise, line 50: kind of a small thing. wonder if its easier to read if the conditional is for the class name instead of changing the element like:
```tsx
<div className={`arrow-${tip.position === "right" ? "left" : "up"}`} />
```

source/javascripts/components/ProductTour/ProductTooltip.tsx
```diff
@@ -38,17 +40,20 @@ export const ProductTooltip = ({
 
 	return (
 		<Box
-			left={rect?.x}
-			top={(rect?.y ?? 0) + (rect?.height ?? 0) + 20}
+			left={tip.position === "right" ? rect!.x + rect!.width + 37: rect!.width/2 - 210}
+			top={tip.position === "right" ? rect!.bottom/2  - 22: (rect?.y ?? 0) + (rect?.height ?? 0) + 40}
```
avivenzio-bitrise, line 44: how does this behave on different screen sizes/zoom levels with these magic numbers?

Another thought is maybe pulling each out into a variable so when reading it might be easier to understand why the math is needed and what each number is based on.

avivenzio-bitrise (commented):
The CI is failing because the e2e tests need the following update if you want to put iit in, not sure which pr will get merged first
https://github.com/bitrise-io/bitrise-workflow-editor/pull/712/files#diff-8038de0d16c18ffbd0581a0138debffc2a8ea7b2c3062c8fdc649e66a1ec738eR166

source/javascripts/components/ProductTour/Highlighter.tsx
cdhenley219-bit: Yes, exactly. It helps to add the thick white border around each highlighted item.

source/javascripts/components/ProductTour/Highlighter.tsx
cdhenley219-bit: Yes, this is all related to the same thing.

cdhenley219-bit:
> The CI is failing because the e2e tests need the following update if you want to put iit in, not sure which pr will get merged first https://github.com/bitrise-io/bitrise-workflow-editor/pull/712/files#diff-8038de0d16c18ffbd0581a0138debffc2a8ea7b2c3062c8fdc649e66a1ec738eR166

Thanks for letting me know. If it's already implemented on your end, then I can wait until your PR is merged first. I'll put this one into Draft status until then.

source/javascripts/components/ProductTour/ProductTooltip.tsx
cdhenley219-bit: Good catch! This has been updated.

source/javascripts/components/ProductTour/ProductTooltip.tsx
cdhenley219-bit: Honestly, I hadn't thought much through it, as the page is already not super responsive. However, what I can do is have it default to the "bottom" position when in mobile view.

source/javascripts/components/ProductTour/ProductTooltip.tsx
```diff
@@ -36,19 +38,25 @@ export const ProductTooltip = ({
...
 			className="product-tooltip"
+			position="relative"
 		>
 			<Box display="flex" flexGrow="1">
-				<div className="arrow-up" style={{ left: rect!.width / 2 - 5 }} />
+				<div className={`arrow-${tip.position === "right" ? "left" : "up" }`} />
```
roland-spekker-bitrise, line 54: Now that we have a proper bitkit version, could we switch to the `Popover` component to fix a lot of manual stuff we do here? Render something for the `PopoverTrigger` and we could just delete most of the tooltip positioning related stuff. 

I would do it in a follow up PR if it's not trivial to do, to not interfere with this development.

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -96,6 +96,8 @@ import datadogRumCustomTiming from "../utils/datadogCustomRumTiming.ts";
 			viewModel.menuIds = viewModel.menus.map(function(menu) {
```
roland-spekker-bitrise, line 96: I think the `viewModel.menuIds` were only used by the tour so we can delete them if different ids will be used

source/javascripts/components/ProductTour/ProductTooltip.tsx
cdhenley219-bit: This has been taken out and replaced with the BitKit popover

source/javascripts/components/ProductTour/ProductTooltip.tsx
```diff
@@ -36,19 +38,25 @@ export const ProductTooltip = ({
...
+	const onStartAgain = (): void => {
+		onButtonClick("start again");
+		onRestart();
+	};
+
 	return (
 		<Box
-			left={rect?.x}
-			top={(rect?.y ?? 0) + (rect?.height ?? 0) + 20}
+			left={tip.position === "right" ? rect!.x + rect!.width + 37: rect!.width/2 - 210}
```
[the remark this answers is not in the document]
cdhenley219-bit: This has been taken out and replaced with the BitKit popover

source/javascripts/components/ProductTour/ProductTooltip.tsx
cdhenley219-bit: This has been updated. Please see my latest changes

source/javascripts/controllers/_MainController.js.erb
cdhenley219-bit: These have been taken out. For some reason, when I tried doing it before, it resulted in unexpected behavior. However, it may have been due to other items during development, as it's now working perfectly without the old `menuId`s

avivenzio-bitrise (commented):
looks good to me! just a few very small comments

roland-spekker-bitrise (changes requested):

source/javascripts/components/ProductTour/HighlighterProductTooltip.tsx
```diff
@@ -0,0 +1,135 @@
...
+	selectedIndex,
+	total,
+	onNext,
+	onPrev,
+	onRestart,
+	rect,
+	onClose,
+	onButtonClick,
+}:HighlighterProductTooltipProps):(JSX.Element|null) => {
+    if (!tip) {
```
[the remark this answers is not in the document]
avivenzio-bitrise: good catch, we should add some linting rules around this.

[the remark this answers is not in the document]
cdhenley219-bit: This was logic previously there that I reused, but I'll update it to follow this convention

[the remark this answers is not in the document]
roland-spekker-bitrise: yep the logic was there but without a hook

source/javascripts/components/ProductTour/ProductTour.tsx
```diff
@@ -78,19 +77,20 @@ export const ProductTourContent = ({ menuIds, currentUser, onDismiss }: ProductT
...
 				onNext={onNext}
 				onPrev={onPrev}
+				onRestart={onRestart}
 				selectedIndex={selectedIndex!}
 				total={items.length}
 				rect={rect}
 				tip={tip}
-			/>
-		</Highlighter>
+			/>
```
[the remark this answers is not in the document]
cdhenley219-bit: These have been removed. It'd be nice if WFE had linting rules to check for things like this. Something to think about and maybe add in the near future.

source/javascripts/components/ProductTour/useHighlightedArea.ts
```diff
@@ -27,7 +27,7 @@ export const useHighlightedArea = (selectedId: string | undefined): DOMRect | nu
 		const element = document.getElementById(id);
 		if (element) {
 			ref.current = element;
-			const rect = element.getBoundingClientRect();
+			const rect = element.getBoundingClientRect();
```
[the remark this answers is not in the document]
cdhenley219-bit: Removed

source/javascripts/components/ProductTour/useHighlightedArea.ts
```diff
@@ -41,7 +41,7 @@ export const useHighlightedArea = (selectedId: string | undefined): DOMRect | nu
 	useEffect(() => {
 		const onResize = (): void => {
 			if (ref.current) {
-				const rect = ref.current!.getBoundingClientRect();
+				const rect = ref.current.getBoundingClientRect();
```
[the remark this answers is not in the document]
cdhenley219-bit: I didn't receive any errors about it, likely because line 43 already has a check in place for the existence of the ref. Initially I made the update because I thought it was the root cause of the bug I was talking about in standup, but now that I know it isn't, I've reverted it back to the prior value.

source/javascripts/components/ProductTour/ProductTour.tsx
```diff
@@ -1,14 +1,14 @@
...
+import { HighlighterOverlay } from "./HighlighterOverlay";
+import { HighlighterProductTooltip } from "./HighlighterProductTooltip";
 import { ProductTourProps, Tips } from "./types";
 import { useTrackingFunction } from "../../hooks/utils/useTrackingFunction";
 import { tips } from "./tips";
 import { useWaitForElements } from "./useWaitForElement";
 import { useProductTour } from "./useProductTour";
 import { useHighlightedArea, getClipPathFromRect } from "./useHighlightedArea";
-
+
```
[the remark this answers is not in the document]
cdhenley219-bit: This has been removed

source/javascripts/components/ProductTour/HighlighterProductTooltip.tsx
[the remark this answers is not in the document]
cdhenley219-bit: Looking at it again, I'm not sure of what to change. `HighlighterProductTooltip` is a component, and I'm not making any calls to a hook within a conditional statement. I have moved the check for the null tip to where the other rendering logic is located to make it more readable, but if anything else is needed, please let me know.

avivenzio-bitrise (approved):

roland-spekker-bitrise (approved):

PR #714, opened 2022-10-21
[PLG-377] update guide build icon based on build status

avivenzio-bitrise:
https://bitrise.atlassian.net/browse/PLG-377

- change build icon on guide based on build status

### Screenshots
Screen Shot 2022-10-21 at 3 04 19 PM
Screen Shot 2022-10-21 at 3 03 56 PM

### DIFF
--- a/source/javascripts/components/GuidedOnboarding/types.ts
+++ b/source/javascripts/components/GuidedOnboarding/types.ts
@@ -1,6 +1,7 @@
 import { ReactNode } from "react";
 
 export interface AppStep {
+    id: string;
     title: string;
     isSuccessful: boolean;
     subSteps: SubStep[];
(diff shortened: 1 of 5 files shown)


cdhenley219-bit (dismissed):

source/javascripts/components/GuidedOnboarding/types.ts
```diff
@@ -12,3 +13,11 @@ export interface SubStep {
     content: ReactNode;
     href?: string;
 }
+
+export enum BuildStatus {
```
roland-spekker-bitrise, line 17: This is absolutely just a preference, but I'm curious what others think about it. I personally prefer objects with `as const` vs enums. 

Screenshot 2022-10-24 at 10 19 05

Both the build output and the typing make more sense to me.

roland-spekker-bitrise (dismissed):

source/javascripts/components/GuidedOnboarding/types.ts
avivenzio-bitrise: I like this pattern, updated

roland-spekker-bitrise (approved):

PR #716, opened 2022-11-01
Added functionality for rendering proper icon

cdhenley219-bit:


source/javascripts/_componentRegister.js
```diff
@@ -74,4 +74,4 @@ angular
 	)
 	.component("rWorkflowRecipesInfoBanner", register(WorkflowRecipesInfoBanner, []))
 	.component("rProductTour", register(ProductTour, ["menuIds", "currentUser", "productTourShown", "onDismiss"]))
-	.component("rGuidedOnboarding", register(GuidedOnboarding, ["isEnabled", "isOpen", "onTurnOff", "buildStatus"]));
+	.component("rGuidedOnboarding", register(GuidedOnboarding, ["isEnabled", "isOpen", "onTurnOff", "buildStatus", "lastWorkflowEditedApp", "lastWorkflowEditedDate", "lastRunningBuildFinishDate"]));
```
jakerobb, line 77: It doesn't seem like we're using `lastWorkflowEditedApp` here anywhere. Am I missing something?

avivenzio-bitrise: +1

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -216,6 +217,7 @@ import datadogRumCustomTiming from "../utils/datadogCustomRumTiming.ts";
 				}).then(
 					function() {
 						viewModel.loadDataProgress.success();
+						//test
```
avivenzio-bitrise, line 220:
```suggestion
```

source/javascripts/components/GuidedOnboarding/GuidedOnboarding.tsx
```diff
@@ -85,12 +87,16 @@ export const GuidedOnboarding = ({
...
+    lastWorkflowEditedDate,
+    lastRunningBuildFinishDate,
 }: GuidedOnboardingProps): JSX.Element | null => {
   const { isMobile } = useResponsive();
 
   const stepSuccessful = {
-    run_a_build: buildStatus !== BuildStatus.Running
+    run_a_build: buildStatus !== BuildStatus.Running,
+    config_workflows: (lastRunningBuildFinishDate && lastWorkflowEditedDate) 
+                        && (lastRunningBuildFinishDate > lastWorkflowEditedDate) || false,
```
avivenzio-bitrise, line 99: does this work the way one would expect with date objects? should something like `getTime()` be used before comparison?

source/javascripts/services/_requestService.js.erb
```diff
@@ -316,7 +316,8 @@ import jsyaml from "js-yaml";
 											name: response.data.owner_type == "User" ? response.data.owner_username : response.data.owner_name
 										},
 										buildData: {
-											lastRunningBuildOrLastBuildStatus: response.data.last_running_build_or_last_build_build_status
+											lastRunningBuildOrLastBuildStatus: response.data.last_running_build_or_last_build_build_status,
+											lastRunningBuildFinishDate: response.data.last_running_build_or_last_build_finished_at
```
avivenzio-bitrise, line 320: Is this parsed into a `Date()` before this step or is it a datetime formatted string?

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -674,11 +691,12 @@ import datadogRumCustomTiming from "../utils/datadogCustomRumTiming.ts";
...
-						const metadataReq = requestService.getCurrentUserMetadata('show_getting_started_guide');
+						const guideMetadataReq = requestService.getCurrentUserMetadata('show_getting_started_guide');
+						const lastWFEDateMetadataReq = requestService.getCurrentUserMetadata('last_workflow_edited_date');
 						const tourReq = requestService.getProductTourShown();
-
-						Promise.all([metadataReq, tourReq]).then(([metadataResp, tourResp]) => {
-							viewModel.guidedOnboardingEnabledForUser = metadataResp;
+						Promise.all([guideMetadataReq, lastWFEDateMetadataReq, tourReq]).then(([guideMetadataResp,lastWFEDateMetadataResp, tourResp]) => {
+							viewModel.guidedOnboardingEnabledForUser = guideMetadataResp;
+							viewModel.lastWorkflowEditedDate = lastWFEADateMetadataResp;
```
avivenzio-bitrise, line 699:
```suggestion
							viewModel.lastWorkflowEditedDate = lastWFEDateMetadataResp;
```

source/javascripts/components/GuidedOnboarding/GuidedOnboarding.tsx
jakerobb: If they're proper JavaScript `Date` objects, [it works](https://stackoverflow.com/a/493018/535401).

avivenzio-bitrise: mind blown

source/javascripts/controllers/_MainController.js.erb
avivenzio-bitrise: same thought as https://github.com/bitrise-io/bitrise-workflow-editor/pull/716/files#r1010892542
Can we confirm the response is parsed into a `Date` object?

source/javascripts/components/GuidedOnboarding/GuidedOnboarding.tsx
roland-spekker-bitrise: Yep it is kind of funny. 

```
So, when + receives an object (including array) for either operand, it first calls the ToPrimitive abstract operation (section 9.1) on the value, which then calls the [[DefaultValue]] algorithm (section 8.12.8) with a context hint of number.
```

This is from a [book](https://github.com/getify/You-Dont-Know-JS) I highly recommend btw.

Bonus fun fact, you can actually check this with [Date.prototype[@@toPrimitive]](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/@@toPrimitive). 

```
const a = new Date()
a[Symbol.toPrimitive]('number')
```

As the MDN entry states: 
JavaScript calls the [@@toPrimitive]() method to convert an object to a primitive value. You rarely need to invoke the [@@toPrimitive]() method yourself; JavaScript automatically invokes it when encountering an object where a primitive value is expected.

source/javascripts/_componentRegister.js
jakerobb: Just to be clear: the approach described on the Jira issue calls for the use of `lastWorkflowEditedApp`, and I don't think you can meet this acceptance criteria without it:

> When the user has saved a workflow and then a build **for the same app** has completed successfully, the icon is a fully filled circle.

I wanted you to use it, not take it out! :)

source/javascripts/components/GuidedOnboarding/GuidedOnboarding.tsx
```diff
@@ -85,12 +87,16 @@ export const GuidedOnboarding = ({
...
+    lastWorkflowEditedDate,
+    lastRunningBuildFinishDate,
 }: GuidedOnboardingProps): JSX.Element | null => {
   const { isMobile } = useResponsive();
 
   const stepSuccessful = {
-    run_a_build: buildStatus !== BuildStatus.Running
+    run_a_build: buildStatus !== BuildStatus.Running,
+    config_workflows: (!!lastRunningBuildFinishDate && !!lastWorkflowEditedDate) 
+                        && (new Date(lastRunningBuildFinishDate) > new Date(lastWorkflowEditedDate)),
```
jakerobb, line 99: As with the website PR, I don't see where you're checking that the build was successful.

source/javascripts/controllers/_MainController.js.erb
avivenzio-bitrise: looks like the above suggestion still holds true. I think you will always get `viewModel.lastWorkflowEditedDate` as `undefined` if not a run time error

source/javascripts/services/_requestService.js.erb
avivenzio-bitrise: resolving as prop api has changed to strings

source/javascripts/controllers/_MainController.js.erb
avivenzio-bitrise: bumping this one up as well

source/javascripts/components/GuidedOnboarding/GuidedOnboarding.tsx
cdhenley219-bit: This has been updated.

source/javascripts/controllers/_MainController.js.erb
cdhenley219-bit: this has been removed

source/javascripts/controllers/_MainController.js.erb
cdhenley219-bit: Sorry about the confusion, I didn't properly respond to your feedback here. The response is actually a string, and my last changes reflect this confirmation.

avivenzio-bitrise, line 699: `lastWFEADateMetadataResp` is not defined, I think you have a typo and it should be `lastWFEDateMetadataResp`

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -454,6 +455,19 @@ import datadogRumCustomTiming from "../utils/datadogCustomRumTiming.ts";
 				});
 			}
 
+			function updateLastWorkflowMetadata(){
+				const appSlug = appService.appDetails.slug;
+				const dateToday = new Date().toString();
+			
+				requestService.updateCurrentUserMetadata({
+					'last_workflow_edited_date': dateToday,
+					'last_workflow_edited_app': appSlug
```
avivenzio-bitrise, line 464: The api accepts setting two metadata values at once? This function might allow it but I would confirm that the actual http api is behaving as expected

cdhenley219-bit: It only accepts one metadata key per call. This is due to the way it's processed on the backend (there is no controller or model functionality to accept multiple metadata keys). I had created hook called `useUserAllMetadata` to do this on the frontend, but it still would have only been chained requests using `Promise.all`. It works fine with the separate calls, but if more than two metadata items are needed in the future, a rewrite of how this is handled in the backend would be a good idea.

source/javascripts/controllers/_MainController.js.erb
cdhenley219-bit: This has been updated

jakerobb (approved):
LGTM!

PR #724, opened 2023-01-12
CI-608 update the WFE to use the new yml formatting service

roland-spekker-bitrise:


source/javascripts/hooks/api/useMonolithApiCallback.ts
```diff
@@ -5,12 +5,20 @@ export interface MonolithError {
...
+	return useFetchCallback<T, MonolithError>(
+		url,
+		{
+			...init,
+			headers: {
+				"X-CSRF-TOKEN": getCookie("CSRF-TOKEN"),
+				...init?.headers
+			}
+		},
+		parser as any
```
pentzzsolt, line 22: I'm a little concerned about this. `useFetchCallback` passes the body to this function, so without any type safety, I can essentially pass something to a string or something. Do you think this is worth spending some extra effort on?

source/javascripts/components/common/RepoYmlStorageActions.tsx
```diff
@@ -1,18 +1,59 @@
...
+const useFormattedYml = (appConfig: AppConfig): { yml: string; failed: boolean; loading: boolean } => {
+	const [yml, setYml] = useState(typeof appConfig === "string" ? appConfig : "");
+	const formatAppConfigRef = useRef<() => void>();
+	const { loading, failed, result, call } = useMonolithApiCallback<string>(
+		"/api/cli/format",
+		{
+			method: "POST",
+			body: JSON.stringify(appConfig)
+		},
+		(value: string) => value
```
pentzzsolt, line 21: I'm wondering if instead of this, we could pass a parameter to `useMonolithApiCallback`, something like an enum of `text` and `json`, which would then internally call either `text` or `json` on the return value of `fetch`? Feels a bit awkward to pass an identity function here just to override the default value.

pentzzsolt (changes requested):

source/javascripts/components/common/RepoYmlStorageActions.tsx
```diff
@@ -1,18 +1,59 @@
...
+
+		if (result && !failed) {
+			setYml(result);
+		}
+	}, [result, failed, appConfig]);
+
+	return {
+		yml,
+		failed: !!failed,
+		loading
```
gsanta, line 48: `failed` and `loading` are not used anywhere

gsanta (changes requested):

source/javascripts/hooks/api/useMonolithApiCallback.ts
roland-spekker-bitrise: Sure, I'll fix the typing

source/javascripts/components/common/RepoYmlStorageActions.tsx
roland-spekker-bitrise: I'd prefer the flexibility of the parser function over something like just text/json. This way you can do additional processing on it. It's kind of analogous to a select function in react-query, at least that is my mental model about it.

roland-spekker-bitrise: And on a side note, I would avoid enums in general.

source/javascripts/components/common/RepoYmlStorageActions.tsx
roland-spekker-bitrise: Removed them, now just returning the yml

gsanta (approved):

PR #725, opened 2023-01-13
feat: provide yml to formatter endpoint instead of json - CI-608

roland-spekker-bitrise:


source/javascripts/hooks/api/useFetchCallback.ts
```diff
@@ -25,55 +25,59 @@ function useFetchCallback<T, E>(
...
+						throwError(new Error("Url is required"));
+						return;
 					}
-				});
 
-				setStatusCode(result.status);
+					const result = await fetch(url, {
+						signal: abortController.current?.signal,
+						...init,
+						...options,
```
roland-spekker-bitrise, line 47: FYI This options line is the only relevant change, all the other stuff is just formatting. Allowing additional options at this point means we only need to do the yml formatting when we actually want to call the endpoint. 

`useFetchCallback` previously only had the option to do this with the init object. The previous implementation did the json stringify on each render.

Screenshot 2023-01-13 at 13 56 10

This was kind of passable but not ideal but for the yml transform it would be even worse.

source/javascripts/components/common/RepoYmlStorageActions.tsx
```diff
@@ -9,40 +9,43 @@ type RepoYmlStorageActionsProps = {
...
-		{
-			method: "POST",
-			body: JSON.stringify(appConfig)
-		},
-		identity
-	);
+	const formatAppConfigRef = useRef<(options?: RequestInit) => void>();
+	const { failed, result, call } = useMonolithApiCallback<string>("/api/cli/format", fetchOptions, identityParser);
 
 	// NOTE: call function isn't referentially stable
```
gsanta, line 26: Is this still referentially unstable now that fetchOptions are moved out of `useFormattedYml `?

gsanta (changes requested):

source/javascripts/components/common/RepoYmlStorageActions.tsx
roland-spekker-bitrise: Yes, as `useMonolithApiCallback` is still recreating the init object to add the csrf cookie.

gsanta (approved):

PR #735, opened 2023-01-25
Add border to update step button

gsanta:


source/stylesheets/_tab-workflows.scss.erb
```diff
@@ -494,6 +494,8 @@ section.manage-workflows {
 									background-color: $def-danger;
 									@include transition(background-color $def-transition_duration);
 
+									box-shadow: 0px 0px 0px 2px white;
```
roland-spekker-bitrise, line 497: No variable for the bg color? It's fine by me, but if there is one, we could use that.

gsanta: No variable I checked and it was just 'white'.

roland-spekker-bitrise (approved):

PR #750, opened 2023-02-28
fix: both radio buttons could be selected at once

roland-spekker-bitrise:


roland-spekker-bitrise:
It turns out that it does need a `RadioGroup` wrapper. I've hooked it up as controlled and sync it to the `copyRepositoryYmlToWebsite` value.

gsanta (approved):

gsanta (approved):

PR #751, opened 2023-02-28
[PLG-1289] group machine types by chip

avivenzio-bitrise:
https://bitrise.atlassian.net/browse/PLG-1289

### Screenshots

Screen Shot 2023-02-28 at 11 29 54 AM

### DIFF
--- a/source/javascripts/controllers/_StackController.js.erb
+++ b/source/javascripts/controllers/_StackController.js.erb
@@ -41,9 +41,9 @@
 				return result;
 			}, {});
 
-			viewModel.getStackMachineTypeChipOrder = function (stackId) {
+			viewModel.getStackMachineTypeChipOrder = function(stackId) {
 				if (viewModel.stackMachineMap) {
-					return Object.keys(viewModel.stackMachineMap[stackId] || {}).sort()
+					return Object.keys(viewModel.stackMachineMap[stackId] || {}).sort();
 				}
 				return [];
       		}
--- a/source/javascripts/controllers/_StackController.js.erb
+++ b/source/javascripts/controllers/_StackController.js.erb
@@ -36,10 +36,17 @@
 					}
 					return machineType.stackType === stack.type;
 				});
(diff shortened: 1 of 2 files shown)


source/javascripts/controllers/_StackController.js.erb
```diff
@@ -36,11 +36,18 @@
 					}
 					return machineType.stackType === stack.type;
 				});
-				result[stack.id] = _.groupBy(stackMachineTypes, 'tag');
```
jakerobb, line 39: Oh, I didn't realize we were already grouping! I checked, and currently the only values for `tag` are `gen1` and `gen2`, so I'm confident there's no downside here.

source/javascripts/controllers/_StackController.js.erb
```diff
@@ -36,11 +36,18 @@
...
 				return result;
 			}, {});
 
+			viewModel.getStackMachineTypeChipOrder = function(stackId) {
+				if (viewModel.stackMachineMap) {
+					return Object.keys(viewModel.stackMachineMap[stackId] || {}).sort();
+				}
+				return [];
+			}
+
```
jakerobb, line 50: Here we're basically saying "sort the groups alphabetically by the value of `chip`", right? i.e. relying on the fact that `apple` comes before `intel`.

So, if for example if we started supporting AWS's Graviton chips (and used either `aws` or `graviton` as the value for `chip`), those would insert themselves between Apple and Intel. 

In less of a rush, I'd wish for something more robust -- but don't change it now. 🙂

jakerobb (approved):
LGTM!

source/javascripts/controllers/_StackController.js.erb
avivenzio-bitrise: So I thought about this and decided that doing a plain sort is the simplest option for now that will allow us to group by the value of `chip` without knowing what the possible values are. I think it would be better for us to add a new machine `chip` value and see the machine groups out of order than to not see the new machine. also this protects us in the unlikely event that `apple` is changed to `Apple` or something like that.

PR #752, opened 2023-02-28
chore: update dev start script

avivenzio-bitrise:
Running in local website mode was not working because the `MODE` env var was being overwritten.


roland-spekker-bitrise (approved):
Nice find

PR #753, opened 2023-03-06
[SSW-120] return copy of all stacks for other project type

avivenzio-bitrise:
https://bitrise.atlassian.net/browse/SSW-120


jakerobb (approved):
LGTM! :awyeah:

PR #765, opened 2023-03-29
Update bitrise cli dependency

godrei:
This PR updates bitrise cli go dependency by running `go get github.com/bitrise-io/bitrise` && `go mod tidy` && `go mod vendor`.

When running the WF editor locally bitrise-cli models are used in the `apiserver` package for parsing and validating the given bitrise.yml. Because of the outdated dependency the WF editor was not aware of the latest bitrise.yml format version (12), so opening such bitrise.yml locally failed with:

```
The bitrise.yml has a higher Format Version (12) than the bitrise CLI model's version (11).
```


zoltanszabo-bitrise (changes requested):
Do we really need to push the vendor folder?

morbalint (approved):

PR #774, opened 2023-05-02
Made FE fixes for Product Tour Bubble

cdhenley219-bit:


source/javascripts/components/ProductTour/HighlighterProductTooltip.tsx
```diff
@@ -105,7 +105,7 @@ export const HighlighterProductTooltip = ({
                     </Box>
 
                 {finished ? (
-                    <Box>
+                    <Box display="flex" flexDirection="column" gap="8">
```
avivenzio-bitrise, line 108: Can you add a screenshot or something for this? wondering if `flexDirection="column"` is what we want?

PR #785, opened 2023-05-22
Final version of Machine deprecation warning [ER-446]

doomsayer13:


source/javascripts/services/_dateService.js.erb
```diff
@@ -58,9 +58,11 @@
 			};
 
 			dateService.toLocaleMonthDayDateString = function(string) {
-				var date = new Date(string);
+				const date = new Date(string);
+				const month = date.toLocaleDateString('en-US', { month: 'short' });
+				const day = date.toLocaleDateString('en-US', { day: '2-digit' });
 
-				return date.toLocaleDateString('en-US', { month: 'short', day: '2-digit' });
+				return day + 'th of ' + month;
```
erosdome, line 65: This will not produce 1st, 2nd, 3rd. We should either introduce a lib for this whole functionality (I think we shouldn't do that now just because of this), or ask design whether we can use a simpler format instead (e. g. 2023. 05. 22).

erosdome (dismissed):

erosdome (approved):

PR #801, opened 2023-08-03
fix: small WFE bugs [BIVS-2050]

morbalint:
there was an issue with `undefined !== null` when checking for missing value, the fix is using `!=` instead.

### DIFF
--- a/source/javascripts/services/_appService.js.erb
+++ b/source/javascripts/services/_appService.js.erb
@@ -424,11 +424,11 @@
 					}
 
 					appService.validationWarnings = [];
-					if (validationData.warnings.config !== null) {
+					if (validationData.warnings.config.length > 0) {
 						appService.validationWarnings = appService.validationWarnings.concat(validationData.warnings.config);
 					}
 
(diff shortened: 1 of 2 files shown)


source/javascripts/services/_appService.js.erb
```diff
@@ -424,11 +424,11 @@
 					}
 
 					appService.validationWarnings = [];
-					if (validationData.warnings.config !== null) {
+					if (validationData.warnings.config.length != null) {
```
roland-spekker-bitrise, line 427: The previous version checks if the config isn't null and now there is no null check and we check if the length isn't null? The length of an array won't be null.

roland-spekker-bitrise (changes requested):
Can you elaborate on this fix?

source/javascripts/services/_appService.js.erb
morbalint: nope, fixed

morbalint:
> Can you elaborate on this fix?

added a more detailed description

roland-spekker-bitrise (approved):

PR #804, opened 2023-08-11
add unique step limit notice to YML tab [ER-657]

nyikos-zoltan:
https://bitrise.atlassian.net/browse/ER-657


bejczib:
The company appreciates your service. Kepp up!

PR #818, opened 2023-11-22
[RA-2081] Update deps and cleanup

gaborpongracz:


miklosboros (approved):
This should be tested

PR #833, opened 2023-12-08
[RA-2155] Rollback for default and workflow specific stacks

gsanta:


rails/strings.rb
```diff
@@ -693,6 +693,14 @@ def strings
                     gen1: "The tried and tested machines that have been serving our customers for many years now",
                     gen2: "These machines run on a new hardware generation, offering improved performance compared to Gen1 hardware"
                 },
+                rollback_version: {
+                    not_available: "Previous version is not available",
+                    use_rollback_version: "Use previous version",
```
gaborpongracz, line 698: `rollback_version.use_rollback_version`? maybe the second `rollback_version` could be left out.

gaborpongracz (approved):

PR #835, opened 2024-01-03
Improve stack description

ofalvai:


gsanta:
Do we want this "Read more about the Stack Update Policy" for workfow-specific stacks too, or enough for the default stack?

Screenshot 2024-01-03 at 11 50 03

ofalvai:
Fixed, good catch @gsanta 🙏

gsanta (approved):

PR #836, opened 2024-01-03
Dependency upgrade [RA-2244]

gsanta:


.tool-versions
```diff
@@ -1,2 +1,2 @@
-ruby 3.2.1
-nodejs 16.17.1
+ruby 3.2.2
+nodejs 18.16.0
```
moczolaszlo, line 2: Is this could be the current LTS? (20.10.0), Is WFE works with that?

spec/integration/cypress/plugins/index.js
```diff
@@ -16,7 +16,7 @@ const cucumber = require("cypress-cucumber-preprocessor").default
 /**
  * @type {Cypress.PluginConfig}
  */
-module.exports = (on, config) => {
+module.exports = (on) => {
   // `on` is used to hook into various events Cypress emits
   // `config` is the resolved Cypress config
```
moczolaszlo, line 21:
```suggestion
```

moczolaszlo (commented):
I only have one question. :)

.tool-versions
gsanta: I will try!
Wanted to make it the same version as the website but if it works with the latest why not.

spec/integration/cypress/plugins/index.js
gsanta: I removed this because with the updated linter it threw an `unused variable` error

moczolaszlo: That ok, just remove the comment too 😄

gsanta: Ah I misunderstood :) good catch

moczolaszlo (approved):

moczolaszlo (approved):

PR #924, opened 2024-04-17
fix: deleting the last workflow

zoltanszabo-bitrise:
CI-2877

image


AndrasEszes (commented):
After I removed all workflows I created a new one... I added a new step, then I removed that workflow again. Config panel still shows the previously added step.

https://github.com/bitrise-io/bitrise-workflow-editor/assets/1190738/d95eef3e-d29a-41c9-91c4-6aa81a63686d

PR #982, opened 2024-05-28
fix: history navigation error when using search params from React

AndrasEszes:


source/javascripts/components/PipelinesPage/providers/SearchParamsProvider.tsx
```diff
@@ -0,0 +1,40 @@
...
+  }, [searchParams]);
+
+  useEffect(() => {
+    const prevSearchParams = new URLSearchParams(window.location.hash.split('?')[1] || '');
+    const nextSearchParams = new URLSearchParams(searchParams);
+
+    prevSearchParams.sort();
+    nextSearchParams.sort();
+
+    if (prevSearchParams.toString() !== nextSearchParams.toString()) {
```
zoltanszabo-bitrise, line 29: Why do not use lodash's `isEqual` here instead?

source/javascripts/components/PipelinesPage/providers/SearchParamsProvider.tsx
```diff
@@ -0,0 +1,40 @@
...
+  setSearchParams: Dispatch<SetStateAction<URLSearchParams>>;
+};
+
+export const SearchParamsContext = createContext<Context>({
+  searchParams: new URLSearchParams(),
+  setSearchParams: () => undefined,
+});
+
+const SearchParamsProvider = ({ children }: Props) => {
+  const [searchParams, setSearchParams] = useState(new URLSearchParams(window.location.hash.split('?')[1] || ''));
```
zoltanszabo-bitrise, line 16: Creating a helper function for this one couldn't hurt, you already write this twice.
`new URLSearchParams(window.location.hash.split('?')[1] || '')`

source/javascripts/components/PipelinesPage/providers/SearchParamsProvider.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  const value = useMemo(() => {
+    return { searchParams, setSearchParams };
+  }, [searchParams]);
+
+  useEffect(() => {
+    const prevSearchParams = new URLSearchParams(window.location.hash.split('?')[1] || '');
+    const nextSearchParams = new URLSearchParams(searchParams);
```
zoltanszabo-bitrise, line 24: Isn't `searchParams` already an `URLSearchParams` type?

source/javascripts/components/PipelinesPage/hooks/useSearchParams.ts
```diff
@@ -1,19 +1,8 @@
...
-    const nextSearchParams = setter(searchParams);
-    const hasNextSearchParams = nextSearchParams.size > 0;
-
-    const url = new URL(window.location.href);
-    url.hash = `#!/${url.hash.replace('#!/', '').split('?')[0]}${hasNextSearchParams ? `?${nextSearchParams}` : ''}`;
-
-    history.replace(url.href.replace(url.origin, ''));
-  };
-
+  const { searchParams, setSearchParams } = useContext(SearchParamsContext);
```
zoltanszabo-bitrise, line 5: You could return an array from the context right away

source/javascripts/components/PipelinesPage/providers/SearchParamsProvider.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+    prevSearchParams.sort();
+    nextSearchParams.sort();
+
+    if (prevSearchParams.toString() !== nextSearchParams.toString()) {
+      const url = new URL(window.location.href);
+      const hasNextSearchParams = nextSearchParams.size > 0;
+      url.hash = `#!/${url.hash.replace('#!/', '').split('?')[0]}${hasNextSearchParams ? `?${nextSearchParams}` : ''}`;
```
zoltanszabo-bitrise, line 32: Something like this instead?
```
const hashBase = url.hash.split('?')[0];
url.hash = nextSearchParams.size > 0 `${hashBase}?${nextSearchParams}` : hashBase;

zoltanszabo-bitrise (changes requested):

source/javascripts/components/PipelinesPage/providers/SearchParamsProvider.tsx
AndrasEszes: I don't think it would work because isEqual returns false when two objects contain methods with different references.

zoltanszabo-bitrise (approved):

PR #995, opened 2024-06-05
Use new endpoints for secrets [ER-1394]

nyikos-zoltan:
https://bitrise.atlassian.net/browse/ER-1394


matenadasdi (approved):
It’s ok for now but I would revisit these files as we progress with the upcoming steps amd decrease complexity and increase readibility.

PR #997, opened 2024-06-05
Fix hide workflows and pipelines link [RA-2935]

gsanta:


AndrasEszes (changes requested):
I'd rather fix this in Bitkit here: https://github.com/bitrise-io/bitkit/blob/master/src/Components/Breadcrumb/Breadcrumb.tsx#L27

We should check the `child` element is a valid react element or not:
```typescript
if (!React.isValidElement(child)) {
  return null
}

return ( ... )
```

OR

```typescript
const items = Children.filter(React.isValidElement).map((child) => {
  ...
})
```

AndrasEszes (approved):

PR #1012, opened 2024-06-18
fix: is_expand default checked on new wf config

AndrasEszes:


source/javascripts/components/WorkflowConfigPanel/components/EnvVarsCard.tsx
```diff
@@ -62,6 +62,9 @@ const EnvVarCard = ({ id, index, onRemove }: { id: string; index: number; onRemo
     transform: CSS.Translate.toString(transform),
   };
 
+  // NOTE: Default value isn't apply because the ref connected to the FormControl instead the Checkbox component in Bitkit
```
gaborszakacs, line 65: typo
```suggestion
  // NOTE: Default value doesn't apply because the ref connected to the FormControl instead the Checkbox component in Bitkit
```

gaborszakacs (dismissed):

source/javascripts/components/WorkflowConfigPanel/components/EnvVarsCard.tsx
AndrasEszes: thx 😅

PR #1062, opened 2024-07-10
feat: ChainWorkflowDrawer

zoltanszabo-bitrise:


source/javascripts/controllers/_WorkflowsController.js.erb
```diff
@@ -114,7 +117,8 @@ import {extractInputNames, extractReleaseNotesUrl} from "../components/StepConfi
           }
         });
 
-        viewModel.enableReactStepDrawer = launchDarklyService.variation("enable-new-wfe-step-drawer", false);
+        viewModel.enableReactStepDrawer = launchDarklyService.variation("enable-new-wfe-step-drawer", true);
+        viewModel.enableReactChainWorkflowDrawer = launchDarklyService.variation("enable-new-chain-workflow-drawer", true);
```
AndrasEszes, line 121: Are you sure to set these `true` by default?

zoltanszabo-bitrise: Why not? 😛

PR #1082, opened 2024-07-19
[FRONTEND-463] replace monaco editor with react version

aorcsik:


source/javascripts/controllers/_YMLController.js
```diff
@@ -83,38 +46,26 @@ import {configureMonacoYaml} from 'monaco-yaml';
...
-          scrollBeyondLastLine: false,
-          readOnly: false,
-          theme: "vs-dark"
+        var unwatchYMLChange = $scope.$watch(() => {
+          return appService.appConfigYML;
+        }, (value) => {
+          if (value !== undefined) {
+            viewModel.yml = value;
+            unwatchYMLChange();
+          };
```
zoltanszabo-bitrise, line 63: Is it for the first initialization, so it's a one-time watcher?
It feels a bit hacky, I would like to better understand what's the problem here?
Are we waiting for the appConfigYML to be set?

source/javascripts/controllers/_YMLController.js
```diff
@@ -83,38 +46,26 @@ import {configureMonacoYaml} from 'monaco-yaml';
...
-            width: codeContainer.clientWidth,
-            height: codeContainer.clientHeight,
-          });
-        }
-
-        $timeout(resizeEditor, 0);
-        window.addEventListener('resize', resizeEditor);
       }
 
+      init();
```
zoltanszabo-bitrise, line 67: Why not call this `init` from the `slim` with `ng-init`?

zoltanszabo-bitrise (dismissed):

source/javascripts/controllers/_YMLController.js
moczolaszlo: I dont know, but I changed to `ng-init`.

source/javascripts/controllers/_YMLController.js
moczolaszlo: I can't answer for it. Maybe @AndrasEszes or @aorcsik .

aorcsik: Yes, it is a "one-time watcher" that allows the editor to be rerendered only until the YML is loaded, but then it prevents rerendering on content change, which causes a loss of focus

You may call this hacky, but putting React into Angular is already a hack.

source/javascripts/controllers/_YMLController.js
aorcsik: We kept it there because we didn't want to change anything that didn't need to be changed.

source/javascripts/controllers/_YMLController.js
zoltanszabo-bitrise: I'm just wondering whether this one-time watcher is enough. If a user switches to another tab and makes changes, would the YML tab reflect those changes?

AndrasEszes: Good question! I think yes, because every time when you switch to the YML tab (as I know) the controller will be re-instantiated, so the one-time watcher will run as expected. But to be sure, I need to try this use-case.

AndrasEszes: @zoltanszabo-bitrise It works as expected!


https://github.com/user-attachments/assets/00661051-9152-4d6d-b987-f36155ab1c10

AndrasEszes (approved):

PR #1091, opened 2024-07-26
[FRONTEND-470] replace old segment code with new tracking logic + cleanup

aorcsik:
- add segmentTrack function
- cleanup legacy segment key and uses
- https://bitrise.atlassian.net/browse/BIVS-2703


.env.dev.website
```diff
@@ -0,0 +1,12 @@
...
+#/       [how it works](https://dotenvx.com/encryption)     /
+#/----------------------------------------------------------/
+DOTENV_PUBLIC_KEY_DEV_WEBSITE="024c22e369c8109468a063a1c5316c4b5aea1a32f179000f618bdd1e1a3a5e4e1c"
+
+# .env.dev.website
+DEV=true
+NODE_ENV=development
+MODE=WEBSITE
+PUBLIC_URL_ROOT=http://localhost:4000
+SEGMENT_KEY="encrypted:BKKIpSvgyU/Tx5CXGVo5BRFe2A5NNFQXUaazZwDu8UbGMAwnnCKKsnIZwamqDHsIsUa2u4BkacKJPQhFGc+MEYYEwPJcOJXkJdPEDK/gcz1maEPofG1E3GENUNCdsRyKSbOFf7y2mE4kWaW03tEZVQa/AbDn7U1bo5HPSRGFLGaM"
```
zoltanszabo-bitrise, line 12: I have a problem with the encrypted value here; which segment key is behind the encrypted value? It's just an extra hoop I don't like.

moczolaszlo: What do you mean? Its the `Segment dev-monolith-frontend`, but ideally, you, as a developer dont have to know that :)
Or do you against the encryption?

moczolaszlo (approved):

PR #1114, opened 2024-08-06
feat: add eslint-import-resolver-alias package

AndrasEszes:
Now we have alias configuration in the `tsconfig.json`, which can potentially bypass the `import/no-cycle` ESLint rule when using aliases in the codebase. To address this, I have configured the `eslint-import-resolver-alias` package for the project. This change ensures that ESLint can detect import cycles even when we use import aliases.


source/javascripts/core/BitriseYml.ts
```diff
@@ -12,48 +9,4 @@ type Meta = Required<BitriseYml>['meta'] & {
   };
 };
 
-function deleteWorkflow(yml: BitriseYml, workflowId: string): BitriseYml {
```
gsanta, line 15: Why is this deleted? Was it unused?

gsanta (changes requested):

source/javascripts/core/BitriseYml.ts
AndrasEszes: Yes, it is unused. We're using the deleteWorkflow function in models/BitriseYmlService.ts. It hasn't been decided how we'll organize things yet, but everything will be in the right place (hopefully) before we release the new React-only Workflows page. 🙂

AndrasEszes: Additionally, this file had an import cycle before the deletion...

gsanta (approved):

PR #1122, opened 2024-08-16
fix: wrong LD client ID for staging environment

AndrasEszes:


source/javascripts/services/_launchDarklyService.js.erb
```diff
@@ -1,11 +1,13 @@
-(function() {
+(function () {
     "use strict";
 
-    const CLIENT_SIDE_ID = '<%= data[:constants][:ld][:client_id] %>';
+    const CLIENT_SIDE_ID = window.parent.location.host === 'app.bitrise.io'
+        ? '5e70774c8a726707851d2fff'
+        : '5e70774c8a726707851d2ffe';
```
erosdome, line 6: Move these to constants, like
`data[:constants][:ld][:client_id_production]`
`data[:constants][:ld][:client_id_staging]`

erosdome (changes requested):

source/javascripts/services/_launchDarklyService.js.erb
AndrasEszes: Thank you, good point! But at the end, when the WFE is no longer using AngularJS, we want to remove all Ruby and Slim templates as well. That is why I wouldn't be putting that into the strings.rb file.

AndrasEszes: I need this because I created a workspace in the staging environment where the new Workflows page is enabled, and I realized (again) that the WFE is loading the production LD flags instead of the staging ones.

erosdome: Good point, though I'd still suggest using the strings.rb for now, and then when we'll eliminate it completely, we'll think of a new replacement solution which will similarly have all constants gathered, as opposed to having them scattered throughout the code.

erosdome (approved):

PR #1135, opened 2024-09-03
Add tests to setChainedWorkflows

zsolt-vicze:


source/javascripts/core/models/BitriseYmlService.ts
```diff
@@ -137,6 +137,7 @@ function setChainedWorkflows(
     return copy;
   }
 
+  // Do we need this? Typescript should prevent this.
```
AndrasEszes, line 140: We can fool the TypeScript easily, but I have no strong opinion with this...
```typescript
BitriseYmlService.setChaniedWorkflow('wf1', 'something' as Placement, [], yml)
```

AndrasEszes (approved):

PR #1142, opened 2024-09-05
refactor: SecretsPage

zoltanszabo-bitrise:
Use the new SecretApi with React Query


source/javascripts/controllers/_MainController.js.erb
```diff
@@ -927,6 +923,68 @@ import {segmentTrack} from "@/utils/segmentTracking";
...
+                "": ""
+              },
+              Variable.defaultVariableConfig()
+            )
+
+            newVariable.key(secret.key);
+            newVariable.value(secret.value);
+            newVariable.isProtected(secret.isProtected);
+            newVariable.isExpose(secret.isExpose);
+            newVariable.isExpand(secret.isExpand);
```
zsolt-vicze, line 979: We don't need `isKeyChangeable` here?

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -927,6 +923,68 @@ import {segmentTrack} from "@/utils/segmentTracking";
...
+
+          // remove the ones that are not in the react list
+          return secretsFromReact.filter(s => !s.isShared).find((s) => s.key === secret.key())
+        });
+
+        // edit existing ones
+        appService.secrets.forEach(secret => {
+          if (secret.scope() === 'workspace') return;
+
+          const s = secretsFromReact.find((s) => s.key === secret.key() && !s.isShared);
```
zsolt-vicze, line 944: We could have a more descriptive name for `s`, e.g. `secretFromReact`?

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -927,6 +923,68 @@ import {segmentTrack} from "@/utils/segmentTracking";
...
+
+          return secret;
+        });
+
+        // add new ones
+        secretsFromReact.forEach(secret => {
+          if (secret.isShared) return;
+
+          const existingSecret = appService.secrets
+            .filter(s => s.scope() !== 'workspace')
```
zsolt-vicze, line 962: If I understand correctly `isShared` and `scope() == 'workspace'` is the same, right? We should use the named function in that case everywhere.

source/javascripts/controllers/_MainController.js.erb
```diff
@@ -927,6 +923,68 @@ import {segmentTrack} from "@/utils/segmentTracking";
...
+            newVariable.value(secret.value);
+            newVariable.isProtected(secret.isProtected);
+            newVariable.isExpose(secret.isExpose);
+            newVariable.isExpand(secret.isExpand);
+
+            appService.secrets.push(newVariable);
+          }
+        });
+
+        appService.savedSecrets = appService.secrets;
```
zsolt-vicze, line 985: Just for my benefit: how/when the save will actually happen?

source/javascripts/core/api/SecretApi.ts
```diff
@@ -164,28 +137,20 @@ async function getSecretValue({
...
-
     const response = await Client.get<SecretsMonolithResponse[number]>(getSecretItemPath(params), {
       signal,
     });
     return response.value;
   }
 
   // CLI mode
-  return Promise.reject(new Error('Getting secret environment value is only available in website mode'));
+  return '';
```
zsolt-vicze, line 150: Why don't we need the error message anymore?

source/javascripts/services/_appService.js.erb
```diff
@@ -339,19 +339,19 @@
...
 
         const appLevelSecrets = appService.secrets.filter((s) => s.scope() !== 'workspace');
 
         return validateVariables(appLevelSecrets, true, true)
-          .then(function () {
-            if (!skipRequest) {
-              return requestService.postSecrets(appLevelSecrets);
-            }
-          })
+          //          .then(function () {
```
zsolt-vicze, line 350: Why do we need this commented out? Can we just delete it?

source/javascripts/core/api/SecretApi.ts
zoltanszabo-bitrise: Because the CLI secrets response already has values.

source/javascripts/controllers/_MainController.js.erb
zoltanszabo-bitrise: Technically yes, but the Angular secret and React secret objects differ a bit, and the Angular one only has the `scope(newScope)` getter/setter function.

AndrasEszes (approved):

PR #1145, opened 2024-09-09
fix: stacks & machines endpoint response processing

AndrasEszes:
The available stack items has no `available_machines` field in the `/app/:appSlug/stacks_and_machines` endpoint response.


AndrasEszes:
I close this PR because @zsolt-vicze will implement this mapping on the backend side.

PR #1255, opened 2024-11-05
Ci 3806 pin only the major version

zsolt-vicze:


source/javascripts/components/unified-editor/StepSelectorDrawer/StepSelectorDrawer.stories.tsx
```diff
@@ -15,6 +17,7 @@ export default {
   args: {
     defaultIsOpen: true,
   },
+  decorators: [(Story) => withBitriseYml(MockYml, Story)],
```
AndrasEszes, line 20: I prefer to use the `TEST_BITRISE_YML` global variable in stories instead of `MockYml`. I would like to remove `MockYml` in the future.

More context: `TEST_BITRISE_YML` contains the `test_bitrise.yml` file in JSON object format, which is used in the local development mode of the WFE.

AndrasEszes (approved):

PR #1260, opened 2024-11-06
fix: load steps from github

AndrasEszes:
https://bitrise.atlassian.net/browse/SSW-1937

GitHub has CORS restrictions on certain headers that the browser sends. The X-Csrf-Token, application/json content-type, and Accept headers are not allowed.


source/javascripts/core/api/StepApi.ts
```diff
@@ -256,7 +256,13 @@ async function getDirectGitStepByCvs(cvs: string, defaultStepLibrary: string): P
       const result = await Client.get<{ content: string }>(url);
       stepYml = parse(atob(result.content)) as StepYmlObject;
     } else {
-      const result = await Client.text(url);
+      const result = await Client.text(url, {
```
zsolt-vicze, line 259: If this affect only GitHub, should we add this only for GitHub?

zoltanszabo-bitrise (approved):

PR #1266, opened 2024-11-08
feat: graph pipeline placeholder workflow

AndrasEszes:


source/javascripts/pages/PipelinesPage/components/PipelineCanvas/GraphPipelineCanvas/components/WorkflowNode/Handles.tsx
```diff
@@ -26,6 +39,20 @@ const defaultHandleButtonStyle: CSSProperties = {
   backgroundColor: 'transparent',
 };
 
+const placeholderNode = (dependsOn?: string | null): Node<PipelineWorkflow> => ({
```
zsolt-vicze, line 42: Maybe `createPlaceholderNode` would be a better name? Same below `createPlaceholderEdge`?

zsolt-vicze (commented):
Maybe this would be a good time to add a new entry to the storybook with a graph pipeline with workflows? I think now just empty states are there.

Also there is a merge conflict right now.

PR #1268, opened 2024-11-08
feat: implement health check test

AndrasEszes:


package.json
```diff
@@ -37,6 +37,7 @@
     "storybook": "storybook dev -p 6006",
     "test": "karma start spec/karma.conf.js",
     "test:jest": "jest",
+    "test:smoke": "playwright test",
```
gaborszakacs, line 40: At some point it would be include something about this in the Readme, how to set up/run these tests locally (if we intend to make them runnable locally) or how it's wired in the CI.

tests/smoke.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { test, expect } from '@playwright/test';
+
+const APP_ID = process.env.SMOKE_TEST_APP_ID || '';
+const VERSION = process.env.NPM_PACKAGE_VERSION || '';
```
gaborszakacs, line 4: Is this an NPM package version? Seems like it's the WFE version 🤔

AndrasEszes: We're using this environment variable in the CI, which contains the WFE version. It's derived from the `package.json` file, so I think this is why the name was chosen.

package.json
AndrasEszes: At this stage, I don’t find it important to make this runnable locally; It doesn’t make sense in its current state. Of course, I can document how to run it in a CI environment if necessary.

tests/smoke.spec.ts
gaborszakacs: interesting, thx for the context!

gaborszakacs (approved):

gsanta (approved):

PR #1309, opened 2024-11-16
fix: workflow renaming

AndrasEszes:


source/javascripts/controllers/_MainController.js.erb
AndrasEszes, line 1: Actually, the changes in this file solve the problem, but by the time I got here, I had already fixed/simplified other parts of the codebase as well.

zoltanszabo-bitrise (approved):

PR #1341, opened 2024-11-27
Unify create dialogs

moczolaszlo:
https://bitrise.atlassian.net/browse/FRONTEND-494

### DIFF
--- a/source/javascripts/components/unified-editor/CreateEntityDialog/CreateEntityDialog.tsx
+++ b/source/javascripts/components/unified-editor/CreateEntityDialog/CreateEntityDialog.tsx
@@ -1,0 +1,101 @@
+import { Button, Dialog, DialogBody, DialogFooter, DialogProps, Input, Select } from '@bitrise/bitkit';
+import { useForm } from 'react-hook-form';
+
+type FormValues = {
+  entityId: string;
+  baseEntityId: string;
+};
+
+type Props = Omit<DialogProps, 'onCloseComplete' | 'title'> & {
+  baseEntityIds: string[];
+  entityName: string;
+  onCloseComplete: (entityId: string) => void;
+  onCreateEntity: (entityId: string, baseEntityId?: string) => void;
+  sanitizer: (value: string) => string;
+  validator: (value: string) => string | boolean;
+};
+
+const CreateEntityDialog = ({
+  baseEntityIds,
+  entityName,
+  onClose,
+  onCloseComplete,
+  onCreateEntity,
+  sanitizer,
+  validator,
+  ...props
+}: Props) => {
+  const {
+    reset,
+    register,
+    getValues,
+    handleSubmit,
+    setValue,
+    formState: { errors },
+  } = useForm<FormValues>({
+    defaultValues: {
+      entityId: '',
+      baseEntityId: '',
+    },
+  });
+
+  const handleNameChange = (event: React.ChangeEvent<HTMLInputElement>) => {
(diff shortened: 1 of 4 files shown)


source/javascripts/components/unified-editor/CreateEntityDialog/CreateEntityDialog.tsx
```diff
@@ -0,0 +1,120 @@
...
+        setSelectedPipeline(entityId);
+      }
+    }
+    reset();
+    onCloseComplete?.();
+  };
+
+  return (
+    <Dialog
+      title={`Create ${ENTITY_NAMES[entityType]}`}
```
zoltanszabo-bitrise, line 72: I think the name of the Dialog should alse come from the specific implementation.

You should think about the generic dialog as a blueprint. Anything dynamic should be come from as a prop, and supplied by the specific implementation.

Making a generic dialog doesnt mean that a single dialog will handle all the use-cases. It will simply implement part of the dialog which would be the same for all. But the differences between the specific dialogs should be abstracted away, and the responsibility of handling those should rely on the specific implementation.

What it means the you will still have separate dialogs for Workflow and Pipeline entities, they will know their own rules, and encapsulate it inside the specific dialog. And they would pass down these implementation details to the generic dialog.

Something like this:
```js
function CreateWorkflowDialog (){
    const baseIds = ...;
    const validateName = ...;
    const onCreate = ...;

    return (
        <CreateEntity title="Create Workflow" baseIds={baseIds} validate={validateName} onCreate={onCreate} />
    );
}
```

source/javascripts/components/unified-editor/CreateEntityDialog/CreateEntityDialog.tsx
```diff
@@ -0,0 +1,120 @@
...
+            inputRef={(ref) => ref?.setAttribute('data-1p-ignore', '')}
+            errorText={errors.entityId?.message}
+            {...register('entityId', {
+              onChange: handleNameChange,
+              validate: (v) => {
+                if (entityType === 'workflow') {
+                  return WorkflowService.validateName(v, baseIds);
+                }
+                return PipelineService.validateName(v, pipelineIds);
+              },
```
zoltanszabo-bitrise, line 93: You probably know what is the problem here based on my previous comments

source/javascripts/components/unified-editor/DeleteEntityDialog/DeleteEntityDialog.tsx
```diff
@@ -1,30 +1,33 @@
...
-  onDeleteWorkflow?: (workflowId: string) => void;
+  entityId: string;
+  entityType: EntityType;
+  onDelete?: (entityId: string) => void;
 };
 
-const DeleteWorkflowDialog = ({ workflowId, onDeleteWorkflow, onClose, ...dialogProps }: Props) => {
-  const deleteWorkflow = useBitriseYmlStore((s) => s.deleteWorkflow);
+const DeleteEntityDialog = ({ entityId, entityType, onDelete, onClose, ...dialogProps }: Props) => {
+  const deleteEntity = useBitriseYmlStore((s) => (entityType === 'pipeline' ? s.deletePipeline : s.deleteWorkflow));
```
zoltanszabo-bitrise, line 13: Nope, it should come from outside

zoltanszabo-bitrise (dismissed):
I dis not go through the full PR but based ony inputs on the Create dialog, you should be able to figure out the rest.

AndrasEszes (approved):

PR #1388, opened 2025-01-17
feat: connect ConfigMergeDialog

AndrasEszes:
https://bitrise.atlassian.net/browse/CI-4154
https://bitrise.atlassian.net/browse/CI-4163


source/javascripts/services/_appService.js.erb
```diff
@@ -643,6 +627,51 @@
...
+
+      appService.detectRemoteAppConfigYMLConflict = function (requestConfig) {
+        return requestService.getAppConfigYML(requestConfig).then(function (remoteYaml) {
+          appService.remoteAppConfigYML = remoteYaml;
+
+          return {
+            yourYaml: appService.appConfigYML,
+            baseYaml: appService.savedAppConfigYML,
+            remoteYaml: remoteYaml,
+            hasConflict: remoteYaml !== appService.savedAppConfigYML,
```
zsolt-vicze, line 654: Do we plan to make this smarter? In this form, is not really conflict detection but simply change detection. We could have more suitable name for this like `hasChange` since the `hasConflict` name is slightly misleading I think.

zsolt-vicze (approved):

PR #1393, opened 2025-01-21
feat: update workflow variant uses on workflow rename

AndrasEszes:
https://bitrise.atlassian.net/browse/CI-4167

### DIFF
--- a/source/javascripts/core/models/BitriseYmlService.ts
+++ b/source/javascripts/core/models/BitriseYmlService.ts
@@ -1165,7 +1165,11 @@ function renameWorkflowInUses(
 ): PipelineWorkflows {
   return mapValues(workflows, (workflow) => {
     const workflowCopy = deepCloneSimpleObject(workflow);
-    workflowCopy.uses = workflowCopy.uses === workflowId ? newWorkflowId : workflowCopy.uses;
+
+    if (workflowCopy.uses === workflowId) {
+      workflowCopy.uses = newWorkflowId;
+    }
+
     return workflowCopy;
   });
 }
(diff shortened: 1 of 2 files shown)


source/javascripts/core/models/BitriseYmlService.ts
```diff
@@ -1158,6 +1158,24 @@ function renameWorkflowInDependsOn(
...
+  workflowId: string,
+  newWorkflowId: string,
+  workflows: PipelineWorkflows,
+): PipelineWorkflows {
+  return mapValues(workflows, (workflow) => {
+    const workflowCopy = deepCloneSimpleObject(workflow);
+
+    workflowCopy.uses = workflowCopy.uses === workflowId ? newWorkflowId : workflowCopy.uses;
+
+    if (shouldRemoveField(workflowCopy.uses, workflow.uses)) {
```
zsolt-vicze, line 1171: Why do we need this? Is it possible the updated value will be empty? I don't think I can rename a worklfow to "".

AndrasEszes: Because if it is missing and the current workflow has no `uses` field, it is set to `undefined` in the previous lines. As a result, the updated YML object ends up with multiple `uses` fields, all with a value of `undefined`, which is different from having no field at all.

zsolt-vicze: Can we just write this?

```
    if (isEmpty(workflowCopy.uses)) {
      delete workflowCopy.uses;
    }
```

zsolt-vicze (approved):

PR #1398, opened 2025-01-22
fix: state inconsistency

zoltanszabo-bitrise:
https://bitrise.atlassian.net/browse/CI-4177

### DIFF
--- a/source/javascripts/controllers/_EnvVarsController.js.erb
+++ b/source/javascripts/controllers/_EnvVarsController.js.erb
@@ -1,312 +1,312 @@
-(function() {
-	"use strict";
+(function () {
+  "use strict";
 
-	angular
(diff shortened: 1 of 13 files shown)


AndrasEszes (changes requested):
It is not ideal because this approach reloads the entire page component after saving. This means that if a dialog is open and someone clicks the 'Save Changes' button, the dialog will close, as all states on the page will reset (except those stored in the URL).

I suggest creating two separate events, e.g., `MainController::reloadAppConfig` and `MainController::reloadAppConfigYML`, which should be triggered based on the currently open page when the user clicks the 'Save Changes' button. For example, if the CI configuration is saved on the Configuration YAML page, you should reload other pages and vice versa.

AndrasEszes (approved):

PR #1422, opened 2025-02-05
feat: send X-Bitrise-Config-Version header on save

AndrasEszes:
https://bitrise.atlassian.net/browse/CI-4287

### DIFF
--- a/source/javascripts/services/request-service.ts
+++ b/source/javascripts/services/request-service.ts
@@ -27,6 +27,8 @@ class RequestService {
 
   public appSlug = '';
 
+  public appConfigVersionHeaderName = 'X-Bitrise-Config-Version';
+
   private logger: Logger | undefined;
 
   constructor(logger: Logger) {
@@ -90,7 +92,7 @@ class RequestService {
     }
 
     const defaultError = window.strings.request_service.load_app_config.default_error;
-    const version = response.headers.get('X-Bitrise-Config-Version');
+    const version = response.headers.get(this.appConfigVersionHeaderName);
     const content = await this.convertResponseToText(response, defaultError);
 
     return { version, content };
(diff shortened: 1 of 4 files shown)


source/javascripts/services/request-service.ts
```diff
@@ -27,6 +27,8 @@ class RequestService {
 
   public appSlug = '';
 
+  public appConfigVersionHeaderName = 'X-Bitrise-Config-Version';
```
zsolt-vicze, line 30: Why do we need to declare this in two places?

spec/javascripts/requestService.spec.js
```diff
@@ -64,7 +64,7 @@ describe("RequestService", () => {
 				it("resolves with app config and version", (done) => {
 					jasmine.Ajax.stubRequest("/api/app/my-app-slug/config.yml").andReturn({
 						responseText: "my-app-config",
-						responseHeaders: { "X-Bitrise-Config-Version": "7a6s5dfvas6df" },
+						responseHeaders: { [RequestService.appConfigVersionHeaderName]: "7a6s5dfvas6df" },
```
zsolt-vicze, line 67: Nitpick/opinion: It's a matter of taste but I usually don't mind hard coded values in tests. Yes, when the constant is updated the test needs to be updated, but usually it is a useful reminder to me.

source/javascripts/services/request-service.ts
AndrasEszes: Good question! Yeah, it's weird... I have no idea why we have two request services in the codebase (a classic AngularJS version and a plain JavaScript class version), but I can link the RequestService field in the old requestService so that it is defined in one place.

spec/javascripts/requestService.spec.js
AndrasEszes: Good point. I'll change that.

zsolt-vicze (approved):

PR #1428, opened 2025-02-06
feat: split the save logic based on feature flag

AndrasEszes:


apiserver/service/bitrise_config.go
```diff
@@ -37,6 +59,12 @@ func GetBitriseYMLHandler(w http.ResponseWriter, r *http.Request) {
 
 // PostBitriseYMLHandler ...
 func PostBitriseYMLHandler(w http.ResponseWriter, r *http.Request) {
+	contStr, err := fileutil.ReadStringFromFile(config.BitriseYMLPath)
```
gaborszakacs, line 62: We're swallowing an error here, is that intentional? Is there a valid scenario when the file cannot be read (e.g. it's not there)? Same question for the other handler.

apiserver/service/bitrise_config.go
```diff
@@ -37,6 +59,12 @@ func GetBitriseYMLHandler(w http.ResponseWriter, r *http.Request) {
...
+	if err == nil && HasConfigVersionConflict(r, contStr) {
+		w.WriteHeader(http.StatusConflict)
```
gaborszakacs, line 64: Maybe it's a bit more human friendly if we respond with an error message as well, we could create a shared method for that and use it in both handlers, seems like we have an "established" error message format already.

gaborszakacs (approved):

PR #1434, opened 2025-02-09
feat: webpack SWC loader

AndrasEszes:


webpack.config.js
```diff
@@ -172,52 +173,42 @@ module.exports = {
...
-            compilerOptions: {
-              sourceMap: !isProd,
-            },
-          },
-        },
-        exclude: /node_modules/,
-      },
-      {
-        test: /\.tsx?$/,
+        test: /\.tsx?$/i,
```
gsanta, line 184: What about source map?

AndrasEszes: Good point! We need to generate source maps for the production build as well, because they should be uploaded to DataDog.

gsanta (approved):

PR #1506, opened 2025-03-04
BIVS-3025

moczolaszlo:
https://bitrise.atlassian.net/browse/BIVS-3025

### DIFF
--- a/source/javascripts/components/unified-editor/StepBundleConfig/StepBundleConfig.context.tsx
+++ b/source/javascripts/components/unified-editor/StepBundleConfig/StepBundleConfig.context.tsx
@@ -1,18 +1,26 @@
-import { createContext, PropsWithChildren, useContext } from 'react';
+import { createContext, PropsWithChildren, useContext, useMemo } from 'react';
 import useStepBundle from '@/hooks/useStepBundle';
 import { StepBundle } from '@/core/models/Step';
 import useStep from '@/hooks/useStep';
 
-type State = StepBundle | undefined;
-const Context = createContext<State>(undefined);
-
 type Props = PropsWithChildren<{ id?: string; stepBundleId?: string; stepIndex: number; workflowId?: string }>;
-const StepBundleConfigProvider = ({ id, stepBundleId, children, stepIndex, workflowId }: Props) => {
+
+type State = { stepBundle: StepBundle | undefined } & Props;
+const Context = createContext<State>({
+  stepBundle: undefined,
+  stepIndex: -1,
+});
+
+const StepBundleConfigProvider = (props: Props) => {
+  const { children, id, stepBundleId, stepIndex, workflowId } = props;
+
   const stepLike = useStep({ stepBundleId, stepIndex, workflowId });
 
   const stepBundle = useStepBundle(id || stepLike?.data?.id || '');
 
-  return <Context.Provider value={stepBundle as StepBundle}>{children}</Context.Provider>;
+  const contextValue = useMemo(() => ({ stepBundle, ...props }), [stepBundle, props]);
+
+  return <Context.Provider value={contextValue}>{children}</Context.Provider>;
 };
 
 export const useStepBundleConfigContext = () => {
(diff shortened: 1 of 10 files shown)


source/javascripts/components/unified-editor/StepBundleConfig/StepBundleConfig.context.tsx
```diff
@@ -1,18 +1,26 @@
...
-const StepBundleConfigProvider = ({ id, stepBundleId, children, stepIndex, workflowId }: Props) => {
+
+type State = { stepBundle: StepBundle | undefined } & Props;
+const Context = createContext<State>({
+  stepBundle: undefined,
+  stepIndex: -1,
+});
+
+const StepBundleConfigProvider = (props: Props) => {
+  const { children, id, stepBundleId, stepIndex, workflowId } = props;
```
AndrasEszes, line 15: What is the difference between `id` and `stepBundleId`? 🤔

source/javascripts/components/unified-editor/StepBundleConfig/StepBundleConfigHeader.tsx
```diff
@@ -10,7 +10,8 @@ type HeaderProps = {
 
 const StepBundleConfigHeader = (props: HeaderProps) => {
   const { variant } = props;
-  const { cvs, id, userValues } = useStepBundleConfigContext() ?? {};
+  const { stepBundle } = useStepBundleConfigContext() ?? {};
+  const { cvs, id, userValues } = stepBundle || {};
```
AndrasEszes, line 14:
```suggestion
  const { cvs, id, userValues } = stepBundle ?? {};
```

source/javascripts/core/services/BitriseYmlService.spec.ts
AndrasEszes, line 1: What happens when your input key is `opts`?
When you try to update an input what is not defined in the step bundle?
When you try to update an input in a non-existent workflow?
When you try to update an input in a non-existent step bundle?
When you try to update an input with a wrong step index?
What about the nested step bundles? Is it working when you try to update an input inside a step bundle what is inside a step bundle what is inside the step bundle?

These are just a few things in my mind what i'm interested, but I'm sure it has lot more edge cases.

source/javascripts/components/unified-editor/StepBundleConfig/utils/StepBundle.utils.ts
AndrasEszes, line 1: Is it almost the same as `EnvVarService.parseYmlEnvVar`. Please add a TODO or something to this about we should re-think the EnvironmentItemModel parsing, because it's probably a very-common model in the YAML.

source/javascripts/components/unified-editor/StepBundleConfig/StepBundleConfig.context.tsx
moczolaszlo: The `id` is the own id, the `stepBundleId` is the parent's id.

source/javascripts/core/services/BitriseYmlService.spec.ts
moczolaszlo: `What happens when your input key is opts?`
Good question, probably the CLI will throw error. But I will add a check for the UI.

`When you try to update an input what is not defined in the step bundle?`
Its not possible.

```When you try to update an input in a non-existent workflow?
When you try to update an input in a non-existent step bundle?```
The function will return the original YML.

`When you try to update an input with a wrong step index?`
There is nothing to change the indexes, after add or delete the list will rerender with new indexes.

```What about the nested step bundles? Is it working when you try to update an input inside a step bundle what is inside a step bundle what is inside the step bundle?```
It's works as the same like in a workflow: stepBundleID + index.

source/javascripts/components/unified-editor/StepBundleConfig/utils/StepBundle.utils.ts
moczolaszlo: True dat.

AndrasEszes (approved):

PR #1514, opened 2025-03-12
Create step bundle based on utility workflow [BIVS-3040]

moczolaszlo:
https://bitrise.atlassian.net/browse/BIVS-3040

### DIFF
--- a/source/javascripts/core/services/BitriseYmlService.ts
+++ b/source/javascripts/core/services/BitriseYmlService.ts
@@ -236,13 +236,44 @@ function cloneStepInStepBundle(stepBundleId: string, stepIndex: number, yml: Bit
   return copy;
 }
 
-function createStepBundle(stepBundleId: string, yml: BitriseYml, baseStepBundleId?: string): BitriseYml {
+function createStepBundle(
+  stepBundleId: string,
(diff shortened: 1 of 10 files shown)


source/javascripts/core/services/BitriseYmlService.ts
AndrasEszes, line 1: Where are the tests for these changes? 🥲

source/javascripts/pages/StepBundlesPage/components/Drawers.tsx
```diff
@@ -34,14 +34,24 @@ const Drawers = () => {
     }
   };
 
+  const handleCreateStepBundle = (newId: string, baseEntityId?: string) => {
+    const [type, baseId] = baseEntityId?.split('#') || [];
+    if (!type || type === 'step_bundles') {
```
AndrasEszes, line 39: Can we use an `enum` for the `type` instead of plain string? It's no a blocker, just hard to remember the possible values of these.

AndrasEszes (changes requested):
I had some comments before... 🙇🏻

AndrasEszes (approved):

PR #1575, opened 2025-05-12
Handle dedicated vs legacy dedicated separately [NOJIRA]

erosdome:


source/javascripts/core/api/StacksAndMachinesApi.mswMocks.ts
```diff
@@ -3,7 +3,8 @@ import { delay, http, HttpResponse } from 'msw';
 import StacksAndMachinesApi, { StacksAndMachinesResponse } from './StacksAndMachinesApi';
 
 type Options = {
-  hasDedicatedMachine?: boolean;
+  isDedicated?: boolean;
+  isLegacyDedicated?: boolean;
```
zoltanszabo-bitrise, line 7: What do you think about merging these into a single prop, like: `privateCloud?: 'no-machines', 'machine-overrides'`

zoltanszabo-bitrise (approved):

PR #1581, opened 2025-05-30
feat: don't guess stack OS type, use the new API field

ofalvai:
### Context

Upcoming Linux stacks will have a more meaningful stack ID convention and won't start with `linux-`. This would break the string parsing thing in this repo, therefore a new `os` field was added to the stack API response.

### Changes

Make the `Stack` model aware of the OS with a typed field (not just a string). I think it's important to make this model as clear as possible, and "warp" everything else, such as the mapping of `osx` to `macos` (`osx` is something I want to purge from all codebases anyway, this is a nice opportunity to make the workflow editor future-proof).

### DIFF
--- a/source/javascripts/core/api/StacksAndMachinesApi.ts
+++ b/source/javascripts/core/api/StacksAndMachinesApi.ts
@@ -19,6 +19,7 @@ type StacksAndMachinesResponse = {
       rollback_version?: {
         [machineTypeId: string]: { free?: string; paying?: string };
       };
+      os?: string;
     };
   };
   available_machines: {
@@ -69,9 +70,20 @@ async function getStacksAndMachines({ appSlug, signal }: { appSlug: string; sign
   const promotedMachineTypes: MachineType[] = [];
   const defaultMachineTypeIdOfOSs: { [key: string]: string } = {};
 
+  const mapOSValues = (os?: string) => {
+    switch (os) {
+      case 'macos':
+        return 'macos';
+      case 'linux':
+        return 'linux';
+      default:
+        return 'unknown';
+    }
+  };
+
   mapValues(
     response.available_stacks,
-    ({ title, description = '', available_machines = [], rollback_version, ...rest }, id) => {
+    ({ title, description = '', available_machines = [], rollback_version, os, ...rest }, id) => {
       availableStacks.push({
         id: String(id),
         name: title,
@@ -80,6 +92,7 @@ async function getStacksAndMachines({ appSlug, signal }: { appSlug: string; sign
           rest['description-link-gen2-applesilicon'] || rest['description-link-gen2'] || rest['description-link'],
         machineTypes: available_machines,
         rollbackVersion: rollback_version,
+        os: mapOSValues(os),
       });
     },
   );
--- a/source/javascripts/core/services/StackAndMachineService.ts
+++ b/source/javascripts/core/services/StackAndMachineService.ts
@@ -44,7 +44,18 @@ function isSelfHostedStack(stack: Stack) {
 }
 
 function getOsOfStack(stack: Stack): string {
-  return stack.id.split('-')[0];
+  switch (stack.os) {
+    case 'macos':
+      // This is a workaround to keep it compatible with other data structures returned by the API.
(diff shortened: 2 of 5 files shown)


source/javascripts/core/services/StackAndMachineService.ts
```diff
@@ -44,7 +44,16 @@ function isSelfHostedStack(stack: Stack) {
...
-  return stack.id.split('-')[0];
+  switch (stack.os) {
+    case 'macos':
+      // This is a workaround to keep it compatible with other data structures returned by the API.
+      // It should be cleaned up once the API response uses `macos` everywhere.
+      return 'osx';
+    case 'linux':
+      return 'linux';
+    default:
+      return 'unknown';
```
AndrasEszes: Don't we need to keep the `stack.id.split('-')[0]` here (as fallback) to avoid compatibility issues?

source/javascripts/core/api/StacksAndMachinesApi.ts
```diff
@@ -19,6 +19,7 @@ type StacksAndMachinesResponse = {
       rollback_version?: {
         [machineTypeId: string]: { free?: string; paying?: string };
       };
+      os?: string;
```
zoltanszabo-bitrise, line 22: Why is it optional? Aren't we committing to providing it in all cases? Or is it because of the self-hosted stack? Can the BE also provide the `unknown`  value in this case?

source/javascripts/core/api/StacksAndMachinesApi.ts
```diff
@@ -69,9 +70,20 @@ async function getStacksAndMachines({ appSlug, signal }: { appSlug: string; sign
   const promotedMachineTypes: MachineType[] = [];
   const defaultMachineTypeIdOfOSs: { [key: string]: string } = {};
 
+  const mapOSValues = (os?: string) => {
```
zoltanszabo-bitrise, line 73: It's unnecessary, can be simplified to `os ?? 'unknown'` or `os || 'unknown'` if we anticipate it being an empty string.

source/javascripts/core/api/StacksAndMachinesApi.ts
ofalvai: It's not always defined, unfortunately. The "self-hosted agent" concept was hacked into the stack list unfortunately, so the backend returns some "stacks" that are self-hosted agents and have no `os` fields.

source/javascripts/core/services/StackAndMachineService.ts
```diff
@@ -44,7 +44,16 @@ function isSelfHostedStack(stack: Stack) {
 }
 
 function getOsOfStack(stack: Stack): string {
-  return stack.id.split('-')[0];
+  switch (stack.os) {
+    case 'macos':
+      // This is a workaround to keep it compatible with other data structures returned by the API.
```
zoltanszabo-bitrise, line 49: Can you be more specific about this?

source/javascripts/core/services/StackAndMachineService.ts
ofalvai: Good idea

source/javascripts/core/api/StacksAndMachinesApi.ts
ofalvai: I'm not a TS expert, but I don't think it's possible to satisfy both TS typechecking and runtime correctness with those. `os` can be any arbitrary string, and we need to map unexpected values to 'unknown'. Unless I miss something obvious.

source/javascripts/core/services/StackAndMachineService.ts
ofalvai: Sure, I added more details, is that enough?

source/javascripts/core/api/StacksAndMachinesApi.ts
zoltanszabo-bitrise: Yes, but the OS won't be some random string, because we define what the OS can be from the source side, but we can keep this just to be on the safe side.

AndrasEszes (approved):

PR #1591, opened 2025-06-11
fix: dont convert floats with trailing zeros

zoltanszabo-bitrise:


viktorbenei:
# ci-pipeline Pipeline Summary

Pipeline status: **aborted**! [View the Pipeline](https://app.bitrise.io/app/1686da85b5935fd6/pipelines/24cd4882-c0c6-49c1-9304-4cc25cec4629) 
 [*run-ci-checks*](https://app.bitrise.io/build/78cbed76-c067-4bbb-9c72-a3d963fda9da)  -  <br>[*run-deploy-storybook*](https://app.bitrise.io/build/ea379f0f-231d-426a-81ae-3939cf882978)  -  <br>

AndrasEszes:
I have two questions regarding these:

1. I notice there are many "DO NOT CONVERT" conditions in the toTypedValue function. Are all of these really necessary? It feels a bit unusual that we rely on multiple regex tests to exclude specific cases instead of explicitly handling only the cases we want to convert.

2. There are quite a few tests for the toTypedValue function, which is impressive. However, in real life, we usually use this function to set a value in the YAML Document, which is then stringified using yaml.stringify. Wouldn't it be beneficial to have some integration tests to verify how these values are actually stringified by the yaml package?

zoltanszabo-bitrise:
> I have two questions regarding these:
> 
> 1. I notice there are many "DO NOT CONVERT" conditions in the toTypedValue function. Are all of these really necessary? It feels a bit unusual that we rely on multiple regex tests to exclude specific cases instead of explicitly handling only the cases we want to convert.
> 2. There are quite a few tests for the toTypedValue function, which is impressive. However, in real life, we usually use this function to set a value in the YAML Document, which is then stringified using yaml.stringify. Wouldn't it be beneficial to have some integration tests to verify how these values are actually stringified by the yaml package?

1. Yes, some are not necessary. I was also considering simplifying it.
2. The yaml package will use the type information to decide how to stringify it.
(Generally we shouldn't be concerned about it, we can look up it's tests).

PR #1610, opened 2025-07-04
[BIVS-3278] Show disabled state for triggers on Triggers page

k-olly:
JIRA-card: https://bitrise.atlassian.net/browse/BIVS-3278

### DIFF
--- a/source/javascripts/pages/TriggersPage/components/TargetBasedTriggers/TargetBasedTriggers.tsx
+++ b/source/javascripts/pages/TriggersPage/components/TargetBasedTriggers/TargetBasedTriggers.tsx
@@ -22,8 +22,11 @@ import TriggerConditions from '@/components/unified-editor/Triggers/TriggerCondi
 import { trackEditTrigger, trackTriggerEnabledToggled } from '@/core/analytics/TriggerAnalytics';
 import { TargetBasedTrigger, TriggerSource, TYPE_MAP } from '@/core/models/Trigger';
 import TriggerService from '@/core/services/TriggerService';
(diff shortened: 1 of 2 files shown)


source/javascripts/pages/TriggersPage/components/TargetBasedTriggers/TargetBasedTriggers.tsx
```diff
@@ -158,6 +165,11 @@ const TargetBasedTriggers = () => {
                           isDraftPr={trigger.isDraftPr}
                           priority={trigger.priority}
                           triggerType={trigger.triggerType}
+                          triggerDisabled={
+                            !trigger.isActive ||
+                            (source === 'workflows' && workflowTriggersEnabled[sourceId] === false) ||
+                            (source === 'pipelines' && pipelineTriggersEnabled[sourceId] === false)
```
moczolaszlo, line 171: Please, use `useTargetBasedTriggers`.
`useTargetBasedTriggers(source, sourceId).enabled`

moczolaszlo (changes requested):

source/javascripts/pages/TriggersPage/components/TargetBasedTriggers/TargetBasedTriggers.tsx
k-olly: We can not use useTargetBasedTriggers, which returns enabled field for 1 specific trigger. We need here all triggers enabled fields from the workflows and pipelines, if you iterate on them with a map or foreach, you can not use a hook inside.

PR #1636, opened 2025-08-26
Show invalid per-app machine in workflow machine selector when unset, instead of OS default [SSW-2366]

erosdome:


source/javascripts/core/services/StackAndMachineService.ts
```diff
@@ -293,9 +293,10 @@ function prepareStackAndMachineSelectionData(props: SelectStackAndMachineProps):
   const defaultMachineTypeOfOS = getMachineById(selectableDefaultMachines, osMachineId);
 
   const isInvalidMachineType = isInvalidStack || (!!selectedMachineTypeId && !selectedMachineType);
+  const isInvalidDefaultMachineType = isInvalidStack || (!!projectMachineTypeId && !defaultMachineType);
 
   // Machine type options
-  if (!withoutDefaultOptions) {
+  if (!withoutDefaultOptions && !isInvalidDefaultMachineType) {
```
erosdome, line 306: I saw that we have this dedicated block for adding the default machine type options, and we also have a block below for handling the invalid machine type. The current special case is both (default & invalid), so it could be handled in either block - I chose the latter, but if you prefer I can do it in this one instead.

zoltanszabo-bitrise (approved):

PR #1756, opened 2026-04-29
feat: [BE-2051] Declarative tools setup `tools:` block data layer

marcell-vida:
Tool versions for declarative tool setup can only be configured in the yaml file, which is blocking many users from trying it. This PR implements the data layer for a UI-based version configurator for the declarative tool setup.

Adding data layer is the first iteration of implementing a UI based configurator.

Check [BE-2051](https://bitrise.atlassian.net/browse/BE-2051) for more information.

[BE-2051]: https://bitrise.atlassian.net/browse/BE-2051?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


source/javascripts/core/models/Tools.ts
```diff
@@ -0,0 +1,8 @@
+export type VersionStrategy = 'absolute-latest' | 'latest-of' | 'preinstalled' | 'exact';
```
ofalvai: What about following the naming convention of the CLI codebase? For example, `latest-released` vs `latest-installed`?
It's not a hard preference, just curious why this enum differs from the CLI enum values.

source/javascripts/core/services/ToolsService.ts
```diff
@@ -0,0 +1,119 @@
+import { ParsedToolVersion } from '../models/Tools';
+import { bitriseYmlStore, updateBitriseYmlDocument } from '../stores/BitriseYmlStore';
+import YmlUtils from '../utils/YmlUtils';
+import WorkflowService from './WorkflowService';
+
+enum ToolsSource {
```
ofalvai: nit: `ToolScope`? Sounds a bit more accurate.

source/javascripts/core/services/ToolsService.ts
```diff
@@ -0,0 +1,119 @@
...
+      return `${parsed.prefix}:latest`;
+    case 'preinstalled':
+      return `${parsed.prefix}:installed`;
+    case 'exact':
+      return parsed.version;
+  }
+}
+
+function validateSourceId(source: ToolsSource, workflowId?: string, doc = bitriseYmlStore.getState().ymlDocument) {
+  if (source === ToolsSource.Workflow && !workflowId) {
```
ofalvai: nit: I'm not a TypeCcript magician myself, but I have a feeling this runtime check could be converted into a type system check

source/javascripts/hooks/useTools.ts
```diff
@@ -0,0 +1,10 @@
+import { Tools } from '@/core/models/BitriseYml';
+import useBitriseYmlStore from '@/hooks/useBitriseYmlStore';
+
+export function useTools(): Tools {
+  return useBitriseYmlStore(({ yml }) => yml.tools ?? {});
+}
+
+export function useWorkflowTools(workflowId: string): Tools {
+  return useBitriseYmlStore(({ yml }) => yml.workflows?.[workflowId]?.tools ?? {});
```
ofalvai: nit: is it a good idea to silently swallow errors and return an empty tool result, as if there were no tools set by the user?

Definitely a nitpick comment and I don't want to make the final call, I'm not that familiar with the codebase.

source/javascripts/core/services/ToolsService.spec.ts
```diff
@@ -0,0 +1,285 @@
...
+    it('parses "latest" as latest-released without prefix', () => {
+      expect(ToolsService.parseToolVersion('latest')).toEqual({ strategy: 'latest-released' });
+    });
+
+    it('parses "<prefix>:latest" as latest-released with prefix', () => {
+      expect(ToolsService.parseToolVersion('22:latest')).toEqual({ strategy: 'latest-released', prefix: '22' });
+      expect(ToolsService.parseToolVersion('3.3:latest')).toEqual({ strategy: 'latest-released', prefix: '3.3' });
+    });
+
+    it('parses "<prefix>:installed" as latest-installed', () => {
```
zoltanszabo-bitrise, line 16: Do we support `installed` (without a prefix) similar to `latest` for resolving to the `latest-installed` version?

source/javascripts/core/services/ToolsService.spec.ts
```diff
@@ -0,0 +1,285 @@
...
+      expect(ToolsService.parseToolVersion('3.3:installed')).toEqual({ strategy: 'latest-installed', prefix: '3.3' });
+    });
+
+    it('parses a bare semver as exact', () => {
```
zoltanszabo-bitrise: How are partial or normalized versions resolved?
`3`/ `3.3` / `3.3.x` / `3.x.x`?

Should they be treated as `3:installed` and `3.3:installed`?

source/javascripts/core/services/ToolsService.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+  return [];
+}
+
+function validateToolId(id: string, existingIds: string[] = []) {
+  if (!id.trim()) {
+    return 'Tool ID is required';
+  }
+
+  if (existingIds.includes(id)) {
```
zoltanszabo-bitrise: It might return the error if you change the tool id, and than change it back to the original. Most service also use the initialX as a param, so you can check whether we are back at square 1.

source/javascripts/core/services/ToolsService.ts
```diff
@@ -0,0 +1,110 @@
...
+    return 'Tool ID must be unique';
+  }
+
+  return true;
+}
+
```
zoltanszabo-bitrise, line 82: I also feel that the tool version could have it's on `validateToolVersion` function, because there will be cases when the user just types `22:`, which could mean either `22:installed` or `22:latest`, and additional edge cases. The setToolVersion should be resilient as well, but having a validation for tool version could be a first line of defense here.

source/javascripts/core/services/ToolsService.spec.ts
```diff
@@ -0,0 +1,285 @@
...
+            node: 22:latest
+        `);
+
+        ToolsService.deleteTool('node', { type: 'root' });
+
+        expect(getYmlString()).toEqual(yaml`format_version: '13'`);
+      });
+    });
+
+    describe('workflow-level', () => {
```
zoltanszabo-bitrise, line 316: Add a test for "workflow does not exists" use-case

source/javascripts/core/services/ToolsService.ts
```diff
@@ -0,0 +1,149 @@
...
+const PARTIAL_VERSION_REGEX = /^\d+(\.\d+)*(\.x)*$/;
+const COMPLETE_SEMVER_REGEX = /^\d+\.\d+\.\d+$/;
+
+function parseToolVersion(raw: string): ParsedToolVersion {
+  if (raw === 'unset') {
+    return { kind: 'unset' };
+  }
+
+  if (raw === 'latest') {
```
benbitrise: Should we make this case insensitive? Perhaps someone types `nodejs: Latest`.

source/javascripts/core/models/Tools.ts
marcell-vida: Good call, switched to match the CLI.

source/javascripts/core/services/ToolsService.ts
marcell-vida: Good idea, it was replaced with type level check.

source/javascripts/hooks/useTools.ts
marcell-vida: Yes, that's reasonable, it is modified to return undefined now, in case the workflow does not exist.

source/javascripts/core/services/ToolsService.spec.ts
marcell-vida: Yes, now it is supported.

source/javascripts/core/services/ToolsService.spec.ts
marcell-vida: Yes, they should, now they are handled as `latest-installed`.

source/javascripts/core/services/ToolsService.ts
marcell-vida: Good catch, added a validation so reject malformed inputs.

source/javascripts/core/services/ToolsService.ts
marcell-vida: Good idea, changed it to check only if ID does not equal to the previous one.

source/javascripts/core/services/ToolsService.spec.ts
marcell-vida: `setTool` and `deleteTool` both test it now.

source/javascripts/core/services/ToolsService.ts
marcell-vida: Yes, now it's case insesitive.

AndrasEszes (approved):

PR #1772, opened 2026-05-19
First iteration of Tool Setup UI

ofalvai:
> **Review tip:** This PR is structured into logical commits that can be reviewed in order, commit by commit.
>
> 1. `refactor: ToolsService.setTool accepts strategy and inputValue instead of versionString`
> 2. `refactor: replace useWorkflowTools with useToolsForScope`
> 3. `feat: add ToolRow component`
> 4. `feat: implement ToolVersions UI with tool list and add button`
> 5. `feat: wire ToolVersions into StackAndMachine`

---

Implements the Tool Setup UI, completing the data layer work from #1756. The component lets users add, edit, and remove tools with per-strategy version inputs, scoped to either the root `tools:` block or a workflow-level override.

**Out of scope for now:**

- Final copy and links (Learn more, section heading)
- Fetching valid tool IDs and available versions — requires API work currently in progress
- Icons for popular tools — coming alongside the API work

### DIFF
--- a/source/javascripts/components/ToolVersions/ToolRow.tsx
+++ b/source/javascripts/components/ToolVersions/ToolRow.tsx
@@ -54,48 +54,47 @@ const ToolRow = ({ toolId, versionString, existingToolIds, scope, autoFocus, onR
 
   return (
     <Box display="flex" alignItems="flex-start" gap="12">
-      <Box width="160px" flexShrink={0}>
-        <BitkitTextInput
-          size="md"
-          placeholder="Tool (ruby, node, etc.)"
-          errorText={idError}
-          inputProps={{ defaultValue: toolId, autoFocus, onBlur: handleIdBlur }}
-        />
-      </Box>
+      <BitkitTextInput
+        size="md"
+        width="160px"
+        flexShrink="0"
+        placeholder="Tool (ruby, node, etc.)"
+        errorText={idError}
+        inputProps={{ defaultValue: toolId, autoFocus, onBlur: handleIdBlur }}
+      />
 
       <Box display="flex" flex="1" alignItems="flex-start" gap="8">
-        <Box flex="1">
-          <BitkitNativeSelect
-            size="md"
-            value={parsed.strategy}
-            onChange={(e) => {
-              const newStrategy = e.target.value as VersionStrategy;
-              const isPrefix = (s: VersionStrategy) => s === 'latest-released' || s === 'latest-installed';
-              const newInputValue = isPrefix(parsed.strategy) && isPrefix(newStrategy) ? inputValue : '';
-              ToolsService.setTool(toolId, newStrategy, newInputValue, scope);
-            }}
-          >
-            {Object.entries(STRATEGY_LABELS)
-              .filter(([value]) => !(value === 'unset' && scope.type === 'root'))
-              .map(([value, label]) => (
-                <option key={value} value={value}>
-                  {label}
-                </option>
-              ))}
-          </BitkitNativeSelect>
-        </Box>
+        <BitkitNativeSelect
+          flex="1"
+          size="md"
+          value={parsed.strategy}
+          onChange={(e) => {
+            const newStrategy = e.target.value as VersionStrategy;
+            const isPrefix = (s: VersionStrategy) => s === 'latest-released' || s === 'latest-installed';
+            const newInputValue = isPrefix(parsed.strategy) && isPrefix(newStrategy) ? inputValue : '';
+            ToolsService.setTool(toolId, newStrategy, newInputValue, scope);
+          }}
+        >
+          {Object.entries(STRATEGY_LABELS)
+            .filter(([value]) => !(value === 'unset' && scope.type === 'root'))
+            .map(([value, label]) => (
+              <option key={value} value={value}>
+                {label}
+              </option>
+            ))}
+        </BitkitNativeSelect>
 
         {parsed.strategy !== 'unset' && (
-          <Box width="160px" flexShrink={0}>
-            <BitkitTextInput
-              size="md"
-              placeholder={parsed.strategy === 'exact' ? 'e.g. 24.7.0' : 'prefix, e.g. 22'}
-              inputProps={{
-                value: inputValue,
-                onChange: (e) => ToolsService.setTool(toolId, parsed.strategy, e.target.value, scope),
-              }}
-            />
-          </Box>
+          <BitkitTextInput
+            size="md"
+            width="160px"
+            flexShrink="0"
+            placeholder={parsed.strategy === 'exact' ? 'e.g. 24.7.0' : 'prefix, e.g. 22'}
+            inputProps={{
+              value: inputValue,
+              onChange: (e) => ToolsService.setTool(toolId, parsed.strategy, e.target.value, scope),
+            }}
+          />
         )}
       </Box>
 
--- a/source/javascripts/components/ToolVersions/ToolRow.tsx
+++ b/source/javascripts/components/ToolVersions/ToolRow.tsx
@@ -1,0 +1,107 @@
+import { BitkitIconButton, BitkitNativeSelect, BitkitTextInput, IconMinusCircle } from '@bitrise/bitkit-v2';
+import { Box } from '@chakra-ui/react/box';
+import { FocusEventHandler, useState } from 'react';
+
+import { VersionStrategy } from '@/core/models/Tools';
+import ToolsService, { ToolScope } from '@/core/services/ToolsService';
+
+const STRATEGY_LABELS: Record<VersionStrategy, string> = {
+  'latest-released': 'Latest released version',
+  'latest-installed': 'Latest preinstalled version',
+  exact: 'Exact version',
+  unset: 'Do nothing (unset global setting)',
+};
+
+type ToolRowProps = {
+  toolId: string;
+  versionString: string;
+  existingToolIds: string[];
+  scope: ToolScope;
+  autoFocus?: boolean;
+  onRemove: () => void;
+};
+
+const ToolRow = ({ toolId, versionString, existingToolIds, scope, autoFocus, onRemove }: ToolRowProps) => {
+  const [idError, setIdError] = useState<string | undefined>();
+  const parsed = ToolsService.parseToolVersion(versionString);
+  let inputValue: string;
+  switch (parsed.strategy) {
+    case 'exact':
+      inputValue = parsed.version;
+      break;
+    case 'unset':
+      inputValue = '';
+      break;
+    default:
+      inputValue = parsed.prefix ?? '';
+  }
+
+  const handleIdBlur: FocusEventHandler<HTMLInputElement> = (e) => {
+    const newId = e.target.value.trim();
+    if (newId === toolId) {
+      setIdError(undefined);
+      return;
+    }
+    const validation = ToolsService.validateToolId(newId, toolId, existingToolIds);
+    if (validation !== true) {
+      setIdError(validation);
+      return;
+    }
+    setIdError(undefined);
+    ToolsService.deleteTool(toolId, scope);
+    ToolsService.setTool(newId, parsed.strategy, inputValue, scope);
+  };
+
+  return (
+    <Box display="flex" alignItems="flex-start" gap="12">
+      <Box width="160px" flexShrink={0}>
+        <BitkitTextInput
+          size="md"
+          placeholder="Tool (ruby, node, etc.)"
+          errorText={idError}
+          inputProps={{ defaultValue: toolId, autoFocus, onBlur: handleIdBlur }}
+        />
+      </Box>
+
+      <Box display="flex" flex="1" alignItems="flex-start" gap="8">
+        <Box flex="1">
+          <BitkitNativeSelect
+            size="md"
+            value={parsed.strategy}
+            onChange={(e) => {
+              const newStrategy = e.target.value as VersionStrategy;
+              const isPrefix = (s: VersionStrategy) => s === 'latest-released' || s === 'latest-installed';
+              const newInputValue = isPrefix(parsed.strategy) && isPrefix(newStrategy) ? inputValue : '';
+              ToolsService.setTool(toolId, newStrategy, newInputValue, scope);
+            }}
+          >
+            {Object.entries(STRATEGY_LABELS)
+              .filter(([value]) => !(value === 'unset' && scope.type === 'root'))
+              .map(([value, label]) => (
+                <option key={value} value={value}>
+                  {label}
+                </option>
+              ))}
+          </BitkitNativeSelect>
+        </Box>
+
+        {parsed.strategy !== 'unset' && (
+          <Box width="160px" flexShrink={0}>
+            <BitkitTextInput
+              size="md"
+              placeholder={parsed.strategy === 'exact' ? 'e.g. 24.7.0' : 'prefix, e.g. 22'}
+              inputProps={{
+                value: inputValue,
+                onChange: (e) => ToolsService.setTool(toolId, parsed.strategy, e.target.value, scope),
+              }}
+            />
+          </Box>
+        )}
+      </Box>
+
+      <BitkitIconButton variant="tertiary" icon={IconMinusCircle} label="Remove tool" onClick={onRemove} />
+    </Box>
+  );
+};
+
+export default ToolRow;
--- a/source/javascripts/components/ToolVersions/ToolRow.tsx
+++ b/source/javascripts/components/ToolVersions/ToolRow.tsx
@@ -3,7 +3,7 @@ import { Box } from '@chakra-ui/react/box';
 import { FocusEventHandler, useState } from 'react';
 
 import { VersionStrategy } from '@/core/models/Tools';
-import ToolsService, { ToolScope } from '@/core/services/ToolsService';
+import ToolsService from '@/core/services/ToolsService';
 
 const STRATEGY_LABELS: Record<VersionStrategy, string> = {
   'latest-released': 'Latest released version',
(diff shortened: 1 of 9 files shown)


source/javascripts/core/services/ToolsService.spec.ts
```diff
@@ -62,45 +62,6 @@ describe('ToolsService', () => {
     });
   });
 
-  describe('serializeToolVersion', () => {
```
ofalvai, line 65: `serializeToolVersion` is no longer need to be exported, the component doesn't use it directly. These tests also almost perfectly overlap with `setTool()` tests, so I moved the missing test cases there.

.claude/scheduled_tasks.lock
marcell-vida, line 1: I think we should ignore this file.

source/javascripts/hooks/useTools.ts
```diff
@@ -1,14 +1,15 @@
...
-  });
+export function useToolsForScope(scope: ToolScope): Tools {
+  return useBitriseYmlStore(
+    useShallow(({ yml }) => {
+      if (scope.type === 'workflow') {
+        return yml.workflows?.[scope.workflowId]?.tools ?? {};
+      }
+      return yml.tools ?? {};
+    }),
+  );
```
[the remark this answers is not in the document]
marcell-vida: I avoided `useShallow` because of this exact same warning previously.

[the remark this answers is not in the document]
ofalvai: Fixed

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -1,14 +1,52 @@
...
+              existingToolIds={Object.keys(tools)}
+              scope={scope}
+              autoFocus={toolId === ''}
+              onRemove={() => ToolsService.deleteTool(toolId, scope)}
+            />
+          ))}
+        </Box>
+      )}
+
+      <Box>
```
moczolaszlo: Unnecessary wrapper — the Box has no props at all. The `BitkitButton` can sit directly in the parent flex column (it already has `gap="24"`, so the button just lands in the next gap slot).

```diff
-      <Box>
-        <BitkitButton
-          variant="secondary"
-          size="md"
-          state={'' in tools ? 'disabled' : undefined}
-          onClick={() => ToolsService.setTool('', 'latest-released', '', scope)}
-        >
-          Add new
-        </BitkitButton>
-      </Box>
+      <BitkitButton
+        variant="secondary"
+        size="md"
+        state={'' in tools ? 'disabled' : undefined}
+        onClick={() => ToolsService.setTool('', 'latest-released', '', scope)}
+      >
+        Add new
+      </BitkitButton>
```

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -0,0 +1,107 @@
...
+      return;
+    }
+    setIdError(undefined);
+    ToolsService.deleteTool(toolId, scope);
+    ToolsService.setTool(newId, parsed.strategy, inputValue, scope);
+  };
+
+  return (
+    <Box display="flex" alignItems="flex-start" gap="12">
+      <Box width="160px" flexShrink={0}>
```
moczolaszlo: `BitkitTextInput` extends `BitkitFieldProps` (see `lib/components/BitkitTextInput/BitkitTextInput.tsx:9`), which itself extends `Omit<Field.RootProps, ...>` — so the entire Chakra `Field.Root` prop set flows through. `width` and `flexShrink` work directly on `BitkitTextInput`:

```diff
-      <Box width="160px" flexShrink={0}>
-        <BitkitTextInput
-          size="md"
-          placeholder="Tool (ruby, node, etc.)"
-          errorText={idError}
-          inputProps={{ defaultValue: toolId, autoFocus, onBlur: handleIdBlur }}
-        />
-      </Box>
+      <BitkitTextInput
+        size="md"
+        width="160px"
+        flexShrink="0"
+        placeholder="Tool (ruby, node, etc.)"
+        errorText={idError}
+        inputProps={{ defaultValue: toolId, autoFocus, onBlur: handleIdBlur }}
+      />
```

(Aside: `flexShrink={0}` is a number — Bitkit's convention is that size-style props are strings. `flexShrink="0"`.)

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -0,0 +1,107 @@
...
+        <BitkitTextInput
+          size="md"
+          placeholder="Tool (ruby, node, etc.)"
+          errorText={idError}
+          inputProps={{ defaultValue: toolId, autoFocus, onBlur: handleIdBlur }}
+        />
+      </Box>
+
+      <Box display="flex" flex="1" alignItems="flex-start" gap="8">
+        <Box flex="1">
```
moczolaszlo: `BitkitNativeSelect` also extends `BitkitFieldProps` (`lib/components/BitkitNativeSelect/BitkitNativeSelect.tsx:12-15`); only the 5 select-specific props (`defaultValue`, `name`, `onChange`, `placeholder`, `value`) are split out via `splitProps` — `flex` works directly on `BitkitNativeSelect`:

```diff
-      <Box display="flex" flex="1" alignItems="flex-start" gap="8">
-        <Box flex="1">
-          <BitkitNativeSelect ... />
-        </Box>
+      <Box display="flex" flex="1" alignItems="flex-start" gap="8">
+        <BitkitNativeSelect flex="1" ... />
```

(The outer `<Box display="flex" flex="1" alignItems="flex-start" gap="8">` is needed — flex container layout; only the inner `<Box flex="1">` is redundant.)

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -0,0 +1,107 @@
...
+              .map(([value, label]) => (
+                <option key={value} value={value}>
+                  {label}
+                </option>
+              ))}
+          </BitkitNativeSelect>
+        </Box>
+
+        {parsed.strategy !== 'unset' && (
+          <Box width="160px" flexShrink={0}>
```
moczolaszlo: Same as the tool name `BitkitTextInput` above — `width` + `flexShrink` work directly on the component:

```diff
-          <Box width="160px" flexShrink={0}>
-            <BitkitTextInput
-              size="md"
-              placeholder={parsed.strategy === 'exact' ? 'e.g. 24.7.0' : 'prefix, e.g. 22'}
-              inputProps={{
-                value: inputValue,
-                onChange: (e) => ToolsService.setTool(toolId, parsed.strategy, e.target.value, scope),
-              }}
-            />
-          </Box>
+          <BitkitTextInput
+            size="md"
+            width="160px"
+            flexShrink="0"
+            placeholder={parsed.strategy === 'exact' ? 'e.g. 24.7.0' : 'prefix, e.g. 22'}
+            inputProps={{
+              value: inputValue,
+              onChange: (e) => ToolsService.setTool(toolId, parsed.strategy, e.target.value, scope),
+            }}
+          />
```

moczolaszlo (commented):
## Unnecessary `<Box>` wrappers around Bitkit components

A few `<Box>` wrappers here look like they hit the pattern bitkit-v2's AGENTS.md explicitly discourages.

The rule is already loaded into this project's context — `CLAUDE.md` line 5 imports it:

```
@./node_modules/@bitrise/bitkit-v2/AGENTS.md
```

In the imported AGENTS.md under Coding rules:

> **Minimal wrappers** — only use `<Box>` (from `@chakra-ui/react/box`) when truly needed for layout. Use empty fragments (`<>...</>`) otherwise.

I verified the 4 candidates (see inline comments). Both form components (`BitkitTextInput`, `BitkitNativeSelect`) spread props through `BitkitField` into `Field.Root`, which is a Chakra factory component — so `width`, `flex`, `flexShrink` work directly on them. `BitkitButton` also accepts Chakra style props. So all four Box wrappers can be removed.

Not really this PR's fault — the agent probably just glossed over the rule (AGENTS.md is long, this is one bullet of 14). Might be worth strengthening the enforcement in bitkit-v2 down the line (an ESLint rule for prop-less `<Box>` wrappers, or surfacing the rules earlier in AGENTS.md).

Nice work otherwise! :)

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -0,0 +1,106 @@
...
+  versionString: string;
+  existingToolIds: string[];
+  scope: ToolScope;
+  autoFocus?: boolean;
+  onRemove: () => void;
+};
+
+const ToolRow = ({ toolId, versionString, existingToolIds, scope, autoFocus, onRemove }: ToolRowProps) => {
+  const [idError, setIdError] = useState<string | undefined>();
+  const parsed = ToolsService.parseToolVersion(versionString);
```
moczolaszlo: The strategy is re-derived from the serialized YAML string on every render, but partial inputs round-trip ambiguously. With strategy=`exact`, typing `3.13.4` goes through intermediate values `3`, `3.1`, `3.13` — `parseToolVersion("3")` and `parseToolVersion("3.13")` both return `latest-installed` (they match `PARTIAL_VERSION_REGEX` and fail `COMPLETE_SEMVER_REGEX`). The select flips to "Latest preinstalled" mid-typing and back, which is going to be very confusing.

Suggestion: hold the strategy in local state, initialize it from `versionString` on mount, and only update it on explicit user action (dropdown change). The text input then drives only its own slice without round-tripping through the parser.

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+        {parsed.strategy !== 'unset' && (
+          <BitkitTextInput
+            size="md"
+            width="160px"
+            flexShrink="0"
+            placeholder={parsed.strategy === 'exact' ? 'e.g. 24.7.0' : 'prefix, e.g. 22'}
+            inputProps={{
+              value: inputValue,
+              onChange: (e) => ToolsService.setTool(toolId, parsed.strategy, e.target.value, scope),
```
moczolaszlo: No version validation on the UI side — `validateToolVersion` exists in the service but isn't called here. Clearing the field in `exact` mode writes `tool: ''` to the YAML (invalid state); same for malformed prefixes (e.g. a stray colon).

Suggestion: mirror the `idError` pattern — a `versionError` state + `errorText`, call `ToolsService.validateToolVersion(...)` on blur, and skip the `setTool` write when invalid.

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -0,0 +1,106 @@
...
+          value={parsed.strategy}
+          onChange={(e) => {
+            const newStrategy = e.target.value as VersionStrategy;
+            const isPrefix = (s: VersionStrategy) => s === 'latest-released' || s === 'latest-installed';
+            const newInputValue = isPrefix(parsed.strategy) && isPrefix(newStrategy) ? inputValue : '';
+            ToolsService.setTool(toolId, newStrategy, newInputValue, scope);
+          }}
+        >
+          {Object.entries(STRATEGY_LABELS)
+            .filter(([value]) => !(value === 'unset' && scope.type === 'root'))
```
moczolaszlo: Filtering out `unset` at root scope hides existing data: if the YAML already contains `node: unset` at root (edited externally or via YAML view), `parsed.strategy === 'unset'` but the option isn't rendered, so `<select>` points at a non-existent value. The user sees a blank select and the only recovery is deleting the row.

Suggestion: render the option but `disabled` at root scope, so existing state is at least visible. Or normalize `unset` → `latest-released` when read at root.

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -1,14 +1,51 @@
...
+          ))}
+        </Box>
+      )}
+
+      <BitkitButton
+        variant="secondary"
+        size="md"
+        alignSelf="flex-start"
+        state={'' in tools ? 'disabled' : undefined}
+        onClick={() => ToolsService.setTool('', 'latest-released', '', scope)}
```
moczolaszlo: Persists an empty-keyed entry (`'': latest`) into the YAML the moment "Add new" is clicked. If the user clicks the button and navigates away without filling in the ID, that placeholder stays in the YAML. The `'' in tools` disabled-guard only prevents adding a *second* empty row — it doesn't clean up the first.

Suggestion: keep the draft row in component state (e.g. a `pendingRow` flag or a local draft entry) and only commit to the store on blur once the ID is non-empty and valid. As a bonus this avoids marking the YAML dirty just because the button was clicked.

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -1,14 +1,51 @@
...
+          Configure what tools and versions are required for this workflow to work. Tool setup runs before the first
+          step. <BitkitLink href="#">Learn more</BitkitLink>
+        </Text>
+      </Box>
+
+      {Object.keys(tools).length > 0 && (
+        <Box display="flex" flexDirection="column" gap="16">
+          {Object.entries(tools).map(([toolId, versionString]) => (
+            <ToolRow
+              key={toolId}
```
moczolaszlo, line 44: `key={toolId}` combined with the delete-then-set rename in `handleIdBlur` has two side effects:
1. The row unmounts and remounts on every rename — focus is lost.
2. The renamed entry is appended at the end of the YAML map, so the row visually jumps to the bottom on each rename.

Suggestion: add an in-place `renameTool(oldId, newId, scope)` helper in `ToolsService` that preserves YAML key order, and either keep `key={toolId}` (now stable) or switch to a `useId`-based stable React key.

source/javascripts/hooks/useTools.ts
```diff
@@ -1,14 +1,12 @@
...
-}
-
-export function useWorkflowTools(workflowId: string): Tools | undefined {
+export function useToolsForScope(scope: ToolScope): Tools {
   return useBitriseYmlStore(({ yml }) => {
-    const workflow = yml.workflows?.[workflowId];
-    if (!workflow) return undefined;
-    return workflow.tools ?? {};
+    if (scope.type === 'workflow') {
+      return yml.workflows?.[scope.workflowId]?.tools ?? {};
```
moczolaszlo, line 8: Both branches return a fresh `{}` literal when tools is missing, so Zustand's default equality check sees a new reference every render and re-renders all subscribers even when the YAML didn't change. The old `useTools()` had the same shape, but this rewrite is a good moment to fix it.

Suggestion:
```ts
const EMPTY_TOOLS: Tools = {};
...
return yml.workflows?.[scope.workflowId]?.tools ?? EMPTY_TOOLS;
```

AndrasEszes: I think it is not a real problem here, because the `useBitriseYmlStore` uses the `useShallow` hook inside. So in this case every `{} === {}` comparison should be true.

ofalvai: My Claude also flagged this initially, but then @marcell-vida pointed out the same :) `useShallow()` is applied down there

source/javascripts/components/ToolVersions/ToolRow.tsx
ofalvai: We also noticed this today, and just discussed it within the team. We decided to remove those regexes that try to parse partial version string. This is all just temporary until we have the API that return concrete version suggestions to choose from.

Is it okay to merge this PR as-is, or a.) remove the regexes now, b.) do your suggestion instead?

source/javascripts/components/ToolVersions/ToolRow.tsx
ofalvai: This is technically true, but even if an `unset` entry is set on the global level, it has no effect in practice, it's invalid syntax (even though anyone can write such a YML file). I'd rather not introduce complexity for this use-case, but tell me if it's worth it.

source/javascripts/components/ToolVersions/ToolVersions.tsx
ofalvai: I'm not sure how common this rename pattern is in practice, I don't expect users to rename tools (after all, you need to define an exact tool ID). Also, this will probably be replaced with a dropdown or search suggestions once the API side is ready (which returns both tool IDs and versions for an ID).

source/javascripts/components/ToolVersions/ToolRow.tsx
ofalvai: Update: decided to remove the magic regex as part of this PR: https://github.com/bitrise-io/bitrise-workflow-editor/pull/1772/changes/1dea84e428ee7ef760c5faa818de9973a1ed749f

source/javascripts/components/ToolVersions/ToolVersions.tsx
ofalvai: I gave it a try, let me know what you think: https://github.com/bitrise-io/bitrise-workflow-editor/pull/1772/changes/08f6b2ed7cc44857d41909310155dc3ea008cd16

It increases complexity slightly, so tell me if it's not worth it. Also, I made `ToolRow` much dumber in this commit than before, most of the business logic is in `ToolVersions`.

source/javascripts/components/ToolVersions/ToolRow.tsx
ofalvai: This is intentionally left out of this PR, we'll have a list of tool ID and version values soon, coming from an API. That will reduce the chance of errors like empty tool IDs. Validation will also look different than what we have as `validateToolVersion()` now, which takes a complete string of a version prefix and a strategy.

moczolaszlo (dismissed):
Thanks for the detailed responses and follow-up commits! 🚀

AndrasEszes (approved):

PR #1778, opened 2026-06-02
[BIVS-3678] Visual / YAML switch in WFE

k-olly:
JIRA: https://bitrise.atlassian.net/browse/BIVS-3673

Summary
Adds a Visual / YAML segmented control to the header, replacing the old "Configuration YAML" navigation item in the sidebar. This gives users a cleaner, more prominent way to switch between the visual editor and the raw YAML view.

Changes
Header: Added a BitkitSegmentedControl (Visual / YAML) that navigates between the visual pages and the YML page. Switching back from YAML returns the user to the last visited visual page. If the YAML is invalid, switching to Visual is blocked with a toast.
Navigation: Removed the "Configuration YAML" sidebar item and its divider — the header control replaces it entirely.
MainLayout: The sidebar is hidden while on the YAML page, since the full-width editor doesn't need it.
ConfigSettingsBar: Made layout context-aware — on the YAML page it uses flex-start alignment (no spacer needed), and drops the bottom margin to avoid double-spacing with the editor.
YmlPage: Renders ConfigSettingsBar in website mode so the branch/storage controls are still accessible on the YAML page.
Dependency: Bumped @bitrise/bitkit-v2 to 0.3.237 (picks up the BitkitSegmentedControl fixes used here).

### DIFF
--- a/source/javascripts/components/Header.tsx
+++ b/source/javascripts/components/Header.tsx
@@ -8,7 +8,7 @@ import {
   useResponsive,
   useToast,
 } from '@bitrise/bitkit';
-import { BitkitSegmentedControl } from '@bitrise/bitkit-v2';
+import { BitkitSegmentedControl, IconCode, IconWebUi } from '@bitrise/bitkit-v2';
 import { useCallback, useRef, useState } from 'react';
 import { useEventListener } from 'usehooks-ts';
 
@@ -293,8 +293,12 @@ const Header = () => {
         value={editorView}
         onValueChange={(details) => handleEditorViewChange(details.value)}
       >
-        <BitkitSegmentedControl.Item value="visual">Visual</BitkitSegmentedControl.Item>
-        <BitkitSegmentedControl.Item value="yaml">YAML</BitkitSegmentedControl.Item>
+        <BitkitSegmentedControl.Item icon={IconWebUi} value="visual">
+          Visual
+        </BitkitSegmentedControl.Item>
+        <BitkitSegmentedControl.Item icon={IconCode} value="yaml">
+          YAML
+        </BitkitSegmentedControl.Item>
       </BitkitSegmentedControl>
 
       <Box
--- a/source/javascripts/components/Header.tsx
+++ b/source/javascripts/components/Header.tsx
@@ -68,10 +68,10 @@ const Header = () => {
   const editorView = isYmlPage ? 'yaml' : 'visual';
 
   // Remember the last visual page so switching back from YAML returns to it.
-  const lastVisualPathRef = useRef<string>(paths.workflows);
+  const lastVisualPathnameRef = useRef<string>(paths.workflows);
   useEffect(() => {
     if (!isYmlPage) {
-      lastVisualPathRef.current = path;
+      lastVisualPathnameRef.current = path.split('?')[0];
     }
   }, [isYmlPage, path]);
 
@@ -95,7 +95,10 @@ const Header = () => {
           return;
         }
 
-        navigate(lastVisualPathRef.current);
+        const searchParamsString = new URLSearchParams(searchParams).toString();
+        navigate(
+          searchParamsString ? `${lastVisualPathnameRef.current}?${searchParamsString}` : lastVisualPathnameRef.current,
+        );
       }
     },
     [searchParams, navigate, ymlStatus, toast],
@@ -293,6 +296,7 @@ const Header = () => {
       <BitkitSegmentedControl
         size="sm"
         value={editorView}
+        aria-label="Editor view"
         onValueChange={(details) => handleEditorViewChange(details.value)}
       >
         <BitkitSegmentedControl.Item icon={IconWebUi} value="visual">
--- a/source/javascripts/components/Header.tsx
+++ b/source/javascripts/components/Header.tsx
@@ -8,7 +8,8 @@ import {
   useResponsive,
   useToast,
 } from '@bitrise/bitkit';
-import { useCallback, useState } from 'react';
+import { BitkitSegmentedControl } from '@bitrise/bitkit-v2';
+import { useCallback, useRef, useState } from 'react';
 import { useEventListener } from 'usehooks-ts';
 
 import {
@@ -32,12 +33,15 @@ import { useCiConfigSettings } from '@/hooks/useCiConfigSettings';
 import { closeAIDrawer } from '@/hooks/useCloseAIDrawer';
 import useCurrentPage from '@/hooks/useCurrentPage';
 import useFeatureFlag from '@/hooks/useFeatureFlag';
+import useHashLocation from '@/hooks/useHashLocation';
 import usePushBranch, { PushBranchPayload } from '@/hooks/usePushBranch';
+import useSearchParams from '@/hooks/useSearchParams';
 import useYmlHasChanges from '@/hooks/useYmlHasChanges';
 import useYmlValidationStatus from '@/hooks/useYmlValidationStatus';
 import { usePipelinesPageStore } from '@/pages/PipelinesPage/PipelinesPage.store';
 import { useStepBundlesPageStore } from '@/pages/StepBundlesPage/StepBundlesPage.store';
 import { useWorkflowsPageStore } from '@/pages/WorkflowsPage/WorkflowsPage.store';
+import { paths } from '@/routes';
 
 import ConfigMergeDialog from './ConfigMergeDialog/ConfigMergeDialog';
 import DiffEditorDialog from './DiffEditor/DiffEditorDialog';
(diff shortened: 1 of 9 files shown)


source/javascripts/hooks/useCurrentPage.ts
```diff
@@ -1,6 +1,6 @@
-import useHashLocation from '@/hooks/useHashLocation';
+import { useLocation } from 'wouter';
```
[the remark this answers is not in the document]
k-olly: True, reverted the last commit

source/javascripts/hooks/useCurrentPage.ts
```diff
@@ -1,6 +1,6 @@
...
 
 export default function useCurrentPage() {
-  const [pathWithSearchParams] = useHashLocation();
+  const [pathWithSearchParams] = useLocation();
```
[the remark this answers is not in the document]
k-olly: True, reverted the last commit

source/javascripts/components/unified-editor/ConfigSettingsBar/ConfigSettingsBar.tsx
```diff
@@ -58,84 +57,71 @@ const ConfigSettingsBar = () => {
...
-            <Text as="h5" textStyle="body/md/semibold" color="text/primary">
-              {data?.usesRepositoryYml ? 'in repository' : 'on bitrise.io'}
-            </Text>
-          )}
-        </Box>
+      <Box minWidth={0}>
+        <BitkitList variant="inline" textColor="body" size="md">
+          <BitkitList.Item>bitrise.yml</BitkitList.Item>
+          <BitkitList.Item>
+            <Skeleton loading={isPending}>{data?.usesRepositoryYml ? 'in repository' : 'on bitrise.io'}</Skeleton>
```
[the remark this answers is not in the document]
k-olly: In the official Chakra documentation, Skeleton has loading prop - so not true.

source/javascripts/layouts/MainLayout.tsx
```diff
@@ -25,11 +25,14 @@ const InvalidYmlRedirect = () => {
...
 const MainLayout = () => {
+  const [currentPath] = useHashLocation();
+  const isYmlPage = currentPath.startsWith(paths.yml);
+
   return (
     <Box h="100dvh" display="flex" flexDirection="column">
       <Header />
       <Box display="flex" flex="1" alignItems="stretch" minH={0}>
-        <Navigation borderRight="1px solid" borderColor="border/regular" />
+        {!isYmlPage && <Navigation borderRight="1px solid" borderColor="border/regular" />}
```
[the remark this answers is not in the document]
k-olly: That was true! 
Fixed

source/javascripts/components/Header.tsx
```diff
@@ -57,6 +61,46 @@ const Header = () => {
...
+            status: 'error',
+            title: 'Invalid YAML',
+            description: 'Please fix the errors in your YAML configuration before navigating.',
+            duration: null,
+            isClosable: true,
+          });
+          return;
+        }
+
+        navigate(lastVisualPathRef.current);
```
moczolaszlo: **Blocking — branch / search-param changes made on the YAML page are lost when switching back to Visual.**

The two directions of `handleEditorViewChange` are asymmetric:
- Visual → YAML (line 82) uses the **live** `searchParams`.
- YAML → Visual (this line) navigates to the **frozen** `lastVisualPathRef.current`, which the effect only updates `if (!isYmlPage)` — so anything changed while on the YAML page is ignored.

This matters because the PR now renders `ConfigSettingsBar` on the YAML page, so the user can switch branch there, and `loadConfigFromBranch` writes `branch` into the URL search params (`hooks/useCiConfig.ts`). Repro:
1. `/workflows?branch=main` → ref = `/workflows?branch=main`
2. switch to YAML → `/yml?branch=main`
3. switch branch to `feature-x` on the YAML page → `/yml?branch=feature-x`
4. switch back to Visual → navigates to `/workflows?branch=main` — the `feature-x` switch is lost and the URL diverges from the loaded config.

Suggested fix — store only the pathname and re-attach the live `searchParams` on the way back, symmetric with the YAML direction:
```tsx
const lastVisualPathnameRef = useRef<string>(paths.workflows);
useEffect(() => {
  if (!isYmlPage) lastVisualPathnameRef.current = path.split('?')[0];
}, [isYmlPage, path]);

// visual branch:
const qs = new URLSearchParams(searchParams).toString();
navigate(qs ? `${lastVisualPathnameRef.current}?${qs}` : lastVisualPathnameRef.current);
```

source/javascripts/layouts/MainLayout.tsx
```diff
@@ -25,11 +25,18 @@ const InvalidYmlRedirect = () => {
...
 const MainLayout = () => {
+  const [currentPath] = useHashLocation();
+  const isYmlPage = currentPath.startsWith(paths.yml);
+
   return (
     <Box h="100dvh" display="flex" flexDirection="column">
       <Header />
       <Box display="flex" flex="1" alignItems="stretch" minH={0}>
-        <Navigation borderRight="1px solid" borderColor="border/regular" />
+        <Navigation
```
moczolaszlo, line 54: `ConfigSettingsBar` is now rendered in two places — inside `Navigation` (`components/Navigation.tsx:122`) and in `YmlPage` (`pages/YmlPage/YmlPage.tsx:15`). On the YAML page **both are mounted at once** (since `Navigation` stays mounted via `display:none`), so two live instances run with duplicate dialog/query state.

Suggestion (fine to do in this PR): render it **once** at layout level with CSS grid, keeping the current design exactly — sidebar-width above the submenu on visual pages, full-width bar on the YAML page:
```tsx
<Box
  display="grid"
  flex="1"
  minH={0}
  gridTemplateColumns={isYmlPage ? '1fr' : '256px 1fr'}
  gridTemplateAreas={isYmlPage ? `"bar" "content"` : `"bar content" "nav content"`}
>
  {isWebsiteMode && <ConfigSettingsBar gridArea="bar" />}
  <Navigation gridArea="nav" style={{ display: isYmlPage ? 'none' : undefined }} />
  <Box gridArea="content" overflow="auto">{/* router */}</Box>
</Box>
```
This drops the two existing renders and the `isYmlPage` style toggles inside `ConfigSettingsBar`.

⚠️ Keep `Navigation` **mounted** (the `display:none`, not `{!isYmlPage && <Navigation/>}`) — it hosts the `CI_CONFIG_RECEIVED` and `REQUEST_AI_DRAWER_OPEN` parent-message listeners, which would stop working on the YAML page if it were unmounted. See the note on the next line.

source/javascripts/layouts/MainLayout.tsx
```diff
@@ -25,11 +25,18 @@ const InvalidYmlRedirect = () => {
...
+          borderRight="1px solid"
+          borderColor="border/regular"
+          style={{ display: isYmlPage ? 'none' : undefined }}
```
moczolaszlo, line 58: The inline `style={{ display: 'none' }}` is correct and shouldn't be changed here — `display` can't be passed as a prop because the legacy `Sidebar` type omits it (`Omit<BoxProps, 'display' | ...>`), so the raw `style` attribute is the only escape hatch.

Two notes:
1. Please add a short code comment that `display:none` (rather than conditionally unmounting `Navigation`) is **deliberate**, so the `CI_CONFIG_RECEIVED` / `REQUEST_AI_DRAWER_OPEN` listeners inside `Navigation` stay alive on the YAML page — otherwise this is likely to get "tidied" into a conditional unmount that silently breaks them.
2. Root-cause cleanup is a follow-up (not this PR): lifting those listeners out of `Navigation` into a layout-level hook/provider would make the sidebar purely presentational and freely unmountable, removing this workaround entirely. The `display` omission should **not** be fixed in legacy (v1) Bitkit — but it should be addressed when this `Sidebar` is migrated to `@bitrise/bitkit-v2`, so the inline-style workaround isn't carried over.

source/javascripts/components/Header.tsx
```diff
@@ -246,6 +290,19 @@ const Header = () => {
         {(!isWebsiteMode || !isMobile) && <BreadcrumbLink isCurrentPage>Workflow Editor</BreadcrumbLink>}
       </Breadcrumb>
 
+      <BitkitSegmentedControl
```
moczolaszlo, line 296: The segmented control replaces a previously labeled nav link. Please make sure it exposes an accessible name (e.g. an `aria-label` like "Editor view") so screen-reader users get context beyond the two item labels.

moczolaszlo (dismissed):
Thanks — clean, well-scoped change and the UX is a clear improvement. Requesting changes on one correctness issue; everything else is a suggestion or nit we can settle in discussion.

**Must fix**
- Branch / search-param changes made on the YAML page are lost when switching back to Visual (inline comment in `Header.tsx`). User-visible, since this PR enables branch switching on the YAML page.

**Suggestions**
- De-duplicate `ConfigSettingsBar` (currently rendered and mounted twice); a single layout-level render via CSS grid preserves the current design — inline in `MainLayout.tsx`.
- Add a comment explaining the deliberate `display:none` on `Navigation` (parent-message listeners), and track the listener lift-out + the Bitkit v2 `Sidebar` `display` fix as follow-ups.
- Give the Visual/YAML segmented control an accessible label.

**Tests**
- No tests cover the new switch logic (invalid-YAML block, last-visual-path restore, search-param carry-over). A small unit test on `handleEditorViewChange` would be worth it; please also verify `Header.stories` / `Navigation.stories` still render with the new control.

Happy to re-review and approve once the param-loss issue is addressed (or if it turns out to be a non-issue).

source/javascripts/components/Header.tsx
k-olly: Fixed, thanks for catching this!

source/javascripts/layouts/MainLayout.tsx
k-olly: Added the comment 👍

source/javascripts/layouts/MainLayout.tsx
k-olly: Done

moczolaszlo (approved):
Thanks for the quick turnaround — all points addressed and verified against the diff:

- **Param-loss bug (blocking)**: fixed exactly as discussed — pathname-only ref + live `searchParams` re-attached on the way back, so branch changes made on the YAML page now survive the round-trip. 
- **`ConfigSettingsBar` duplication**: nicely done — single layout-level render via CSS grid, removed from both `Navigation` and `YmlPage`, design preserved.
- **`display:none` rationale**: comment added; `Navigation` stays mounted so the parent-message listeners keep working on the YAML page.
- **Accessibility**: `aria-label` added to the segmented control.
- Stories updated for `Navigation` / `YmlPage`.

Approving. One non-blocking nice-to-have for a follow-up: a small unit test on `handleEditorViewChange` (invalid-YAML block + last-visual-path / search-param round-trip) would lock in the fix. The Bitkit v2 `Sidebar` `display` cleanup also remains a follow-up at migration time, as noted.

PR #1786, opened 2026-06-11
BIVS-3677 [1/7] Prep: standalone fixes touching shared code

trapacska:
Part **1/7** of the modular YAML editor stack (BIVS-3677). This is the only PR in the stack that intentionally changes behavior for current users — keep it small, QA it hard. PRs 2–7 are dormant until the activation PR's feature flag.

## ⚠️ Affects prod — every change below ships to all users immediately

- **Step lists**: deterministic sortable `uniqueId`s — random ids remounted every card whenever lists recomputed; `useShallow` selectors; `StepList` state seeded from props (no empty first frame)
- **useStep**: `gcTime: Infinity` so evicted steps don't re-skeleton when scrolled back into view
- **useSelectedWorkflow / useSelectedStepBundle / usePipelineSelector**: validate against the live location hash instead of the React snapshot (prevents the self-correcting effect from clobbering synchronous navigation)
- **Push-to-branch**: `usePushBranch` now owns the post-push reload; `Header`'s `onSuccess` only closes the dialog (signature simplified to `() => void`)

## ⚠️ QA focus (prod flows, no feature flag involved)

- Drag-and-drop reorder of steps and chained workflows
- Deep links with `workflow_id` / `step_bundle_id` / `pipeline` params
- Push-to-branch end to end (push → toast → editor reloads pushed branch)

### DIFF
--- a/source/javascripts/components/unified-editor/WorkflowCard/components/ChainedWorkflowList.tsx
+++ b/source/javascripts/components/unified-editor/WorkflowCard/components/ChainedWorkflowList.tsx
@@ -30,7 +30,9 @@ function getSortableItem(placement: Placement, parentWorkflowId: string) {
     return {
       id,
       index,
-      uniqueId: crypto.randomUUID(),
+      // Deterministic id: cards are keyed by it, so a random id would remount every
+      // chained workflow card (visible rebuild) whenever the chain list recomputes.
+      uniqueId: `${parentWorkflowId}/${placement}/${id}/${index}`,
       placement,
       parentWorkflowId,
     };
--- a/source/javascripts/hooks/usePushBranch.ts
+++ b/source/javascripts/hooks/usePushBranch.ts
@@ -11,7 +11,7 @@ import {
 import BitriseYmlApi from '@/core/api/BitriseYmlApi';
 import BranchesApi from '@/core/api/BranchesApi';
 import { ClientError } from '@/core/api/client';
-import { getYmlString } from '@/core/stores/BitriseYmlStore';
+import { getYmlString, initializeBitriseYmlDocument } from '@/core/stores/BitriseYmlStore';
 import PageProps from '@/core/utils/PageProps';
 import useBitriseYmlStore from '@/hooks/useBitriseYmlStore';
 
@@ -21,7 +21,8 @@ export type PushBranchPayload = {
 };
 
 type UsePushBranchOptions = {
-  onSuccess?: (response: { ymlString: string; version: string; branch?: string; commitSha?: string }) => void;
+  /** Called after a successful push + store refresh (e.g. to close the dialog). */
+  onSuccess?: () => void;
   onMergeConflict?: (branch: string) => void;
 };
 
@@ -67,11 +68,10 @@ function usePushBranch({ onSuccess, onMergeConflict }: UsePushBranchOptions = {}
             }
           : undefined,
       });
-      const newConfig = await BitriseYmlApi.getCiConfig({
-        projectSlug: PageProps.appSlug(),
-        branch,
-      });
-      onSuccess?.(newConfig);
+      // Reload the pushed branch so the editor reflects the saved state.
+      const newConfig = await BitriseYmlApi.getCiConfig({ projectSlug: appSlug, branch });
+      initializeBitriseYmlDocument(newConfig);
+      onSuccess?.();
     },
     onError: (error, { branch }) => {
       if (error instanceof ClientError && error.status === 409) {
--- a/source/javascripts/hooks/useSelectedWorkflow.ts
+++ b/source/javascripts/hooks/useSelectedWorkflow.ts
@@ -3,7 +3,7 @@ import { useCallback, useEffect, useMemo } from 'react';
 
 import { useWorkflows } from '@/hooks/useWorkflows';
 
-import useSearchParams from './useSearchParams';
+import useSearchParams, { getSearchParamsFromLocationHash } from './useSearchParams';
 
 const GENERATED_WORKFLOW_ID_REGEX = /_[\d]+$/g;
 
@@ -32,8 +32,14 @@ type UseSelectedWorkflowResult = [
 
 const useSelectedWorkflow = (): UseSelectedWorkflowResult => {
   const workflowIds = useWorkflows((s) => Object.keys(s));
-  const [searchParams, setSearchParams] = useSearchParams();
-  const selectedWorkflowId = selectValidWorkflowId(workflowIds, searchParams.workflow_id);
+  // Subscribe to hash changes for re-renders, but validate against the LIVE hash
+  // (not the snapshot): during a synchronous jump-to-definition the active file
+  // swaps before `hashchange` fires, so the snapshot lags and would yield a
+  // fallback id — which the effect below then pins into the URL, clobbering the
+  // jump target.
+  const [, setSearchParams] = useSearchParams();
+  const requestedWorkflowId = getSearchParamsFromLocationHash().workflow_id;
+  const selectedWorkflowId = selectValidWorkflowId(workflowIds, requestedWorkflowId);
 
   const setSelectedWorkflow = useCallback(
     (workflowId?: string | null) => {
@@ -49,10 +55,10 @@ const useSelectedWorkflow = (): UseSelectedWorkflowResult => {
   );
 
   useEffect(() => {
-    if (searchParams.workflow_id !== selectedWorkflowId) {
+    if (requestedWorkflowId !== selectedWorkflowId) {
       setSelectedWorkflow(selectedWorkflowId);
     }
-  }, [searchParams.workflow_id, selectedWorkflowId, setSelectedWorkflow]);
+  }, [requestedWorkflowId, selectedWorkflowId, setSelectedWorkflow]);
 
   return useMemo(() => [selectedWorkflowId, setSelectedWorkflow], [selectedWorkflowId, setSelectedWorkflow]);
 };
--- a/source/javascripts/hooks/useStep.ts
+++ b/source/javascripts/hooks/useStep.ts
@@ -77,6 +77,8 @@ function useStepFromApi(cvs = ''): ApiStepResult {
       cvs && !StepService.isStepBundle(cvs, defaultStepLibrary) && !StepService.isWithGroup(cvs, defaultStepLibrary),
     ),
     staleTime: Infinity,
+    // Keep evicted-from-view steps cached so switching back (e.g. between file tabs) never re-skeletons.
+    gcTime: Infinity,
     retry: false,
   });
 
--- a/jest.config.cjs
+++ b/jest.config.cjs
@@ -8,8 +8,12 @@ module.exports = {
   moduleNameMapper: {
     '\\.(css|less|svg)$': 'identity-obj-proxy',
     '^@bitrise/bitkit-v2$': '<rootDir>/node_modules/@bitrise/bitkit-v2/dist/main.js',
+    // In the jsdom environment the `yaml` package resolves to its untransformed ESM
+    // browser build; pin it to the CJS dist used in the node environment.
+    '^yaml$': '<rootDir>/node_modules/yaml/dist/index.js',
     '@/(.*)': '<rootDir>/source/javascripts/$1',
   },
+  setupFiles: ['<rootDir>/spec/set-node-env.ts'],
   setupFilesAfterEnv: ['<rootDir>/spec/setup-jest.ts'],
   moduleDirectories: ['node_modules', '<rootDir>/spec/__mocks__'],
   watchPlugins: ['jest-watch-typeahead/filename', 'jest-watch-typeahead/testname'],
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,6 +54,8 @@
         "@storybook/react-vite": "^10.3.5",
         "@swc/core": "^1.15.32",
         "@swc/jest": "^0.2.39",
+        "@testing-library/dom": "^10.4.1",
+        "@testing-library/react": "^16.3.2",
         "@types/jest": "^29.5.14",
         "@types/luxon": "^3.7.1",
         "@types/node": "^24.9.2",
@@ -6764,7 +6766,6 @@
       "integrity": "sha512-o4PXJQidqJl82ckFaXUeoAW+XysPLauYI43Abki5hABd853iMhitooc6znOnczgbTYmEP6U6/y1ZyKAIsvMKGg==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.10.4",
         "@babel/runtime": "^7.12.5",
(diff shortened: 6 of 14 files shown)


source/javascripts/hooks/usePushBranch.ts
```diff
@@ -67,11 +68,10 @@ function usePushBranch({ onSuccess, onMergeConflict }: UsePushBranchOptions = {}
...
       });
-      const newConfig = await BitriseYmlApi.getCiConfig({
-        projectSlug: PageProps.appSlug(),
-        branch,
-      });
-      onSuccess?.(newConfig);
+      // Reload the pushed branch so the editor reflects the saved state.
+      const newConfig = await BitriseYmlApi.getCiConfig({ projectSlug: appSlug, branch });
+      initializeBitriseYmlDocument(newConfig);
+      onSuccess?.();
```
moczolaszlo, line 74: Ordering change worth a QA look: the reload now runs *before* the consumer's `onSuccess` (which closes the push dialog), whereas before the dialog closed first and the reload ran after. Re-initializing the whole YAML document while the push dialog is still mounted could trigger an extra render/remount of the subtree behind the still-open dialog. Almost certainly fine, but please exercise this specifically in the push-to-branch QA pass.

source/javascripts/hooks/useStep.ts
```diff
@@ -77,6 +77,8 @@ function useStepFromApi(cvs = ''): ApiStepResult {
       cvs && !StepService.isStepBundle(cvs, defaultStepLibrary) && !StepService.isWithGroup(cvs, defaultStepLibrary),
     ),
     staleTime: Infinity,
+    // Keep evicted-from-view steps cached so switching back (e.g. between file tabs) never re-skeletons.
+    gcTime: Infinity,
```
moczolaszlo, line 81: `gcTime: Infinity` combined with the existing `staleTime: Infinity` means step query entries are never evicted for the session's lifetime — a permanent, monotonic memory footprint rather than a bounded cache. Intentional and fine for typical configs (step count is bounded), just flagging it's unbounded growth in principle.

source/javascripts/components/unified-editor/WorkflowCard/components/ChainedWorkflowList.tsx
```diff
@@ -157,6 +162,7 @@ const ChainedWorkflowList = ({ placement, parentWorkflowId }: Props) => {
   }, [initialSortableItems]);
 
   const isEmpty = sortableItems.length === 0;
+  const hasContent = !isEmpty;
```
moczolaszlo: Nit: `hasContent` is just `!isEmpty`, and both names end up used side by side in the same JSX block (`isAfterRun && isEmpty` next to `isAfterRun && hasContent`). Having two names for the same boolean reads slightly worse than keeping `isEmpty` / `!isEmpty` throughout. Cosmetic.

source/javascripts/hooks/useSelectedWorkflow.ts
```diff
@@ -32,8 +32,14 @@ type UseSelectedWorkflowResult = [
...
   const workflowIds = useWorkflows((s) => Object.keys(s));
-  const [searchParams, setSearchParams] = useSearchParams();
-  const selectedWorkflowId = selectValidWorkflowId(workflowIds, searchParams.workflow_id);
+  // Subscribe to hash changes for re-renders, but validate against the LIVE hash
+  // (not the snapshot): during a synchronous jump-to-definition the active file
+  // swaps before `hashchange` fires, so the snapshot lags and would yield a
+  // fallback id — which the effect below then pins into the URL, clobbering the
+  // jump target.
+  const [, setSearchParams] = useSearchParams();
+  const requestedWorkflowId = getSearchParamsFromLocationHash().workflow_id;
```
moczolaszlo, line 41: Reading `window.parent.location.hash` during render makes render mildly impure. Under React 18 StrictMode double-render the value stays consistent so there's no practical issue today — just worth keeping in mind if this hook ever moves into a concurrent/Suspense boundary. The pattern itself (subscribe via `useSearchParams` for re-renders, read the live hash for the value) is sound. Same note applies to `useSelectedStepBundle` and `usePipelineSelector`.

moczolaszlo (commented):
A few inline notes below — all minor; logic looks correct overall. The one cross-cutting point: this is the only prod-affecting PR in the stack but ships with no test changes, and none of these hooks (`usePushBranch`, `useSelectedWorkflow`, `useSelectedStepBundle`) have existing specs. Worth locking in at least (a) the `usePushBranch` reload→`onSuccess` ordering and (b) the live-hash validation against the snapshot-lag race the comments describe, since that's exactly the kind of timing bug that regresses silently later in the stack.

source/javascripts/hooks/usePushBranch.ts
trapacska: Intentional — the new contract is that `onSuccess` fires only after the store reflects the pushed state; PR 4's modular path relies on this ordering (in-place tree refresh must complete before the dialog closes). Now locked in by a spec: `usePushBranch.spec.tsx` asserts the store already holds the reloaded version/commitSha when `onSuccess` runs. Also added the dialog-remount scenario to the push-to-branch QA checklist.

source/javascripts/hooks/useStep.ts
trapacska: Yes, intentional trade-off: step count per config is small and bounded in practice, and the cache pays for itself on every file-tab switch later in the stack (steps re-mount constantly there). If a pathological config ever shows up we can swap in a finite gcTime — flagged it in the code comment's intent.

source/javascripts/components/unified-editor/WorkflowCard/components/ChainedWorkflowList.tsx
trapacska: Agreed, it read worse — dropped the `hasContent` alias, the block now uses `isEmpty`/`!isEmpty` throughout.

source/javascripts/hooks/useSelectedWorkflow.ts
trapacska: Agreed the read is mildly impure — under StrictMode's double render the value is consistent, and the subscribe-via-snapshot + read-live split is deliberate for the jump race. Noted the concurrent-rendering caveat for if these hooks ever end up under a Suspense boundary. The race itself is now covered by `useSelectedWorkflow.spec.tsx` (live hash wins over a stale snapshot, no URL clobbering).

trapacska:
@moczolaszlo thanks for the review — addressed everything:

- **Tests**: added the two specs you asked for. `usePushBranch.spec.tsx` locks in the reload→`onSuccess` ordering (asserts the store already holds the reloaded version/commitSha when the callback fires, plus a guard-path test), and `useSelectedWorkflow.spec.tsx` covers the live-hash vs stale-snapshot race (synchronous hash change picked up before `hashchange` fires, no URL clobbering) plus the fallback/parallel-variant logic. This required adding `@testing-library/react` as a devDependency (first hook tests in the repo) and a jest `moduleNameMapper` pin so the `yaml` package resolves to its CJS dist under jsdom.
- **`hasContent` nit**: dropped, back to `isEmpty`/`!isEmpty`.
- Replied inline on the push ordering (intentional, contract PR 4 relies on) and the `gcTime`/impure-render notes.

Branch was force-pushed (commit amended to keep the stack one-commit-per-PR); the rest of the stack is rebased on top.

source/javascripts/hooks/usePushBranch.ts
moczolaszlo: Confirmed — the new `usePushBranch.spec.tsx` locks the contract in (asserts the store already reflects version/commitSha of the pushed branch at the moment `onSuccess` fires). That covers my QA concern. Resolving.

source/javascripts/hooks/useStep.ts
moczolaszlo: Acknowledged — intentional and bounded in practice. Resolving.

source/javascripts/components/unified-editor/WorkflowCard/components/ChainedWorkflowList.tsx
moczolaszlo: Thanks for dropping the alias and the related JSX churn. Resolving.

source/javascripts/hooks/useSelectedWorkflow.ts
moczolaszlo: Agreed it's fine under current usage, and the new `useSelectedWorkflow.spec.tsx` race test covers the live-hash behavior. Resolving.

moczolaszlo (dismissed):
All review feedback addressed in the latest revision:

- `usePushBranch.spec.tsx` locks in the reload→`onSuccess` ordering contract, plus the config-not-loaded guard.
- `useSelectedWorkflow.spec.tsx` covers the live-hash validation against the snapshot-lag race, the fallback + URL correction, and generated-id resolution.
- The `hasContent` alias / JSX churn was dropped.

One small thing to keep an eye on (non-blocking): the global `^yaml$` moduleNameMapper in jest.config.cjs now affects every test and hardcodes the `yaml/dist/index.js` path — worth a full-suite green run and a mental note for future `yaml` upgrades.

Nice work — approving.

trapacska:
CI fix: the test step inherits `NODE_ENV=production` from the workflow env, and Jest only defaults `NODE_ENV` to `test` when it's *unset* — so React resolved to its production build and `act()` threw (only visible now that the repo has its first React hook tests). Added `spec/set-node-env.ts` as a jest `setupFiles` entry forcing `NODE_ENV=test` before any test module loads. Verified locally under both `NODE_ENV=production` and default: 1244/1244 green.

source/javascripts/components/unified-editor/WorkflowCard/components/ChainedWorkflowList.tsx
```diff
@@ -43,17 +46,19 @@ function getSortableItemUniqueIds(sortableItems: SortableWorkflowItem[]) {
 
 const ChainedWorkflowList = ({ placement, parentWorkflowId }: Props) => {
   const { droppableContainers } = useDndContext();
-  const workflowIds = useWorkflows((s) => Object.keys(s));
+  const workflowIds = useWorkflows(useShallow((s) => Object.keys(s)));
```
moczolaszlo: `useBitriseYmlStore` already wraps its selector in our `useShallow` (useBitriseYmlStore.ts), and `useWorkflows` goes through it — so wrapping again here double-wraps into `useShallow(useShallow(selector))`: two refs, two dequal passes, no benefit. Our `useShallow` is a deep (`dequal`) compare too, so `Object.keys(s)` was already stable without it. Could we drop the extra `useShallow` and keep the codebase-standard `useWorkflows((s) => Object.keys(s))`?

Same applies to the `useBitriseYmlStore(useShallow(...))` selectors just below (validChainedWorkflowIds) and in StepBundleStepList.tsx / WorkflowStepList.tsx — those can stay plain selectors.

source/javascripts/components/unified-editor/WorkflowCard/components/WorkflowStepList.tsx
```diff
@@ -10,9 +11,11 @@ type Props = {
 };
 
 const WorkflowStepList = ({ workflowId }: Props) => {
-  const steps = useBitriseYmlStore(({ yml }) => {
-    return (yml.workflows?.[workflowId]?.steps ?? []).map((s) => Object.keys(s)[0]);
-  });
+  const steps = useBitriseYmlStore(
+    useShallow(({ yml }) => {
```
k-olly: Please remove unnecessary useShallow, it's already used in useBitriseYmlStore, so it's duplicated

source/javascripts/components/unified-editor/WorkflowCard/components/StepBundleStepList.tsx
```diff
@@ -10,9 +11,11 @@ type Props = {
 };
 
 const StepBundleStepList = ({ stepBundleId }: Props) => {
-  const steps = useBitriseYmlStore(({ yml }) => {
-    return (yml.step_bundles?.[stepBundleId]?.steps ?? []).map((s) => Object.keys(s)[0]);
-  });
+  const steps = useBitriseYmlStore(
+    useShallow(({ yml }) => {
```
k-olly: Please remove unnecessary useShallow, it's already used in useBitriseYmlStore, so it's duplicated

source/javascripts/components/unified-editor/WorkflowCard/components/ChainedWorkflowList.tsx
k-olly: Please remove unnecessary useShallow, it's already used in useBitriseYmlStore, so it's duplicated

source/javascripts/components/unified-editor/WorkflowCard/components/ChainedWorkflowList.tsx
```diff
@@ -43,17 +46,19 @@ function getSortableItemUniqueIds(sortableItems: SortableWorkflowItem[]) {
...
   const isAfterRun = placement === 'after_run';
   const isBeforeRun = placement === 'before_run';
   const isSortable = Boolean(onChainedWorkflowsUpdate);
 
-  const validChainedWorkflowIds = useBitriseYmlStore(({ yml }) => {
-    const chainedWorkflowIds = yml.workflows?.[parentWorkflowId]?.[placement] ?? [];
-    return chainedWorkflowIds.filter((id) => workflowIds.includes(id));
-  });
+  const validChainedWorkflowIds = useBitriseYmlStore(
+    useShallow(({ yml }) => {
```
k-olly: Please remove unnecessary useShallow, it's already used in useBitriseYmlStore, so it's duplicated

source/javascripts/components/unified-editor/WorkflowCard/components/ChainedWorkflowList.tsx
trapacska: Right — `useBitriseYmlStore` already wraps every selector in our dequal-based `useShallow`, so these were double-wrapping for no benefit. Dropped all four (here, the `validChainedWorkflowIds` selector, `StepBundleStepList`, `WorkflowStepList`) — back to plain codebase-standard selectors.

source/javascripts/components/unified-editor/WorkflowCard/components/ChainedWorkflowList.tsx
trapacska: Done — removed, plain selector now (`useBitriseYmlStore`/`useWorkflows` already shallow-compare internally).

source/javascripts/components/unified-editor/WorkflowCard/components/ChainedWorkflowList.tsx
trapacska: Done — removed, plain selector now (`useBitriseYmlStore`/`useWorkflows` already shallow-compare internally).

source/javascripts/components/unified-editor/WorkflowCard/components/StepBundleStepList.tsx
trapacska: Done — removed, plain selector now (`useBitriseYmlStore`/`useWorkflows` already shallow-compare internally).

source/javascripts/components/unified-editor/WorkflowCard/components/WorkflowStepList.tsx
trapacska: Done — removed, plain selector now (`useBitriseYmlStore`/`useWorkflows` already shallow-compare internally).

k-olly (approved):

PR #1787, opened 2026-06-11
BIVS-3677 [2/7] Core: tree models, services, API client

trapacska:
Part **2/7** of the modular YAML editor stack. Stacked on #1786.

## What
All-new core-layer code for multi-file (modular) bitrise.yml configs:
- `models/Tree`: `TreeNode`, `EntityIndex`, wire/response shapes
- `TreeService`: traversal, serialization, source labels (+ tests)
- `FileTreeService`: grouped/collapsed UI file-tree builder (+ tests)
- `EntityIndexService`: which file defines which workflow/pipeline/step bundle (+ tests)
- `BitriseYmlApi`: `getConfig` / `getMergedConfig` / `pushConfigTree` endpoints + wire mappers (purely additive) and MSW mocks
- `BranchesApi`: optional modular `root` payload alongside the existing `bitriseYml`

## Safety
**Zero runtime impact** — nothing in the app imports any of this yet. Existing API functions are untouched; `BranchesApi.pushBranch` keeps its single-file path when `root` isn't passed (and no caller passes it yet).

### DIFF
--- a/source/javascripts/core/api/BitriseYmlApi.tree.mswMocks.ts
+++ b/source/javascripts/core/api/BitriseYmlApi.tree.mswMocks.ts
@@ -1,0 +1,111 @@
+import { delay, http, HttpResponse } from 'msw';
+
+import BitriseYmlApi from '@/core/api/BitriseYmlApi';
+
+const ROOT_NODE = {
+  node_id: 'n_root',
+  path: 'bitrise.yml',
+  contents: 'format_version: "13"\ninclude:\n  - path: modules/workflows.yml\n',
+  source: null,
+  commit_sha: 'a1b2c3d4e5f6789012345678901234567890abcd',
+  editable: true,
+  includes: [
+    {
+      node_id: 'n_workflows',
+      path: 'modules/workflows.yml',
+      contents: 'workflows:\n  build:\n    steps:\n      - script@1: {}\n',
+      source: { path: 'modules/workflows.yml', repository: null, branch: null, tag: null, commit: null },
+      commit_sha: 'a1b2c3d4e5f6789012345678901234567890abcd',
+      editable: true,
+      includes: [],
+    },
+  ],
+};
+
+const ENTITY_INDEX = {
+  workflows: { build: [{ node_id: 'n_workflows' }] },
+  pipelines: {},
+  step_bundles: {},
+};
+
+const MERGED_YML = 'format_version: "13"\nworkflows:\n  build:\n    steps:\n      - script@1: {}\n';
+
+const MODULAR_RESPONSE = {
+  root: ROOT_NODE,
+  entity_index: ENTITY_INDEX,
+  merged_yml: MERGED_YML,
+  branch: 'main',
+};
+
+const SINGLE_NODE_ROOT = {
+  node_id: 'n_root',
+  path: 'bitrise.yml',
+  contents: 'format_version: "13"\nworkflows:\n  build:\n    steps:\n      - script@1: {}\n',
+  source: null,
+  commit_sha: 'a1b2c3d4e5f6789012345678901234567890abcd',
+  editable: true,
+  includes: [],
+};
+
+const SINGLE_NODE_RESPONSE = {
+  root: SINGLE_NODE_ROOT,
+  entity_index: { workflows: { build: [{ node_id: 'n_root' }] }, pipelines: {}, step_bundles: {} },
+  merged_yml: SINGLE_NODE_ROOT.contents,
+  branch: 'main',
+};
+
+export const getConfigTree = (variant: 'modular' | 'single_node' = 'modular', error?: string) => {
+  return http.get(BitriseYmlApi.configTreePath({ projectSlug: ':slug' }), async () => {
+    await delay();
+
+    if (error) {
+      return HttpResponse.json({ error_msg: error }, { status: 422 });
+    }
+
+    return HttpResponse.json(variant === 'modular' ? MODULAR_RESPONSE : SINGLE_NODE_RESPONSE, { status: 200 });
+  });
+};
+
+export const postMergeConfig = (error?: string) => {
+  return http.post('/api/app/:slug/config/merge', async () => {
+    await delay();
+
+    if (error) {
+      return HttpResponse.json({ error_msg: error }, { status: 422 });
+    }
+
+    return HttpResponse.json({ merged_yml: MERGED_YML }, { status: 200 });
+  });
+};
+
+export const postConfigPush = (result: 'ok' | 'conflict' | 'error' = 'ok') => {
+  return http.post('/api/app/:slug/config/push', async () => {
+    await delay();
+
+    if (result === 'error') {
+      return HttpResponse.json({ error_msg: 'validation_failed' }, { status: 422 });
+    }
+
+    if (result === 'conflict') {
+      return HttpResponse.json(
+        {
+          status: 'conflict',
+          conflicts: [{ path: 'modules/workflows.yml', remote: ROOT_NODE.includes[0] }],
+        },
+        { status: 409 },
+      );
+    }
+
+    return HttpResponse.json(
+      {
+        status: 'ok',
+        warnings: [],
+        root: ROOT_NODE,
+        entity_index: ENTITY_INDEX,
+      },
+      { status: 200 },
+    );
+  });
+};
+
+export const treeMocks = { ROOT_NODE, ENTITY_INDEX, MODULAR_RESPONSE, SINGLE_NODE_RESPONSE };
--- a/source/javascripts/core/api/BitriseYmlApi.tree.mswMocks.ts
+++ b/source/javascripts/core/api/BitriseYmlApi.tree.mswMocks.ts
@@ -1,6 +1,6 @@
 import { delay, http, HttpResponse } from 'msw';
 
-import BitriseYmlApi from '@/core/api/BitriseYmlApi';
+import BitriseYmlApi, { WireEntityIndex, WireGetConfigResponse, WireTreeNode } from '@/core/api/BitriseYmlApi';
 
 const ROOT_NODE = {
   node_id: 'n_root',
@@ -20,13 +20,13 @@ const ROOT_NODE = {
       includes: [],
     },
   ],
-};
+} satisfies WireTreeNode;
 
 const ENTITY_INDEX = {
   workflows: { build: [{ node_id: 'n_workflows' }] },
   pipelines: {},
   step_bundles: {},
-};
+} satisfies WireEntityIndex;
 
 const MERGED_YML = 'format_version: "13"\nworkflows:\n  build:\n    steps:\n      - script@1: {}\n';
 
@@ -35,7 +35,7 @@ const MODULAR_RESPONSE = {
   entity_index: ENTITY_INDEX,
   merged_yml: MERGED_YML,
   branch: 'main',
-};
+} satisfies WireGetConfigResponse;
 
 const SINGLE_NODE_ROOT = {
   node_id: 'n_root',
@@ -45,14 +45,14 @@ const SINGLE_NODE_ROOT = {
   commit_sha: 'a1b2c3d4e5f6789012345678901234567890abcd',
   editable: true,
   includes: [],
-};
+} satisfies WireTreeNode;
 
 const SINGLE_NODE_RESPONSE = {
   root: SINGLE_NODE_ROOT,
   entity_index: { workflows: { build: [{ node_id: 'n_root' }] }, pipelines: {}, step_bundles: {} },
   merged_yml: SINGLE_NODE_ROOT.contents,
   branch: 'main',
-};
+} satisfies WireGetConfigResponse;
 
 export const getConfigTree = (variant: 'modular' | 'single_node' = 'modular', error?: string) => {
   return http.get(BitriseYmlApi.configTreePath({ projectSlug: ':slug' }), async () => {
@@ -67,7 +67,7 @@ export const getConfigTree = (variant: 'modular' | 'single_node' = 'modular', er
 };
 
 export const postMergeConfig = (error?: string) => {
-  return http.post('/api/app/:slug/config/merge', async () => {
+  return http.post(BitriseYmlApi.mergeConfigPath(':slug'), async () => {
     await delay();
 
     if (error) {
--- a/source/javascripts/core/api/BitriseYmlApi.tree.mswMocks.ts
+++ b/source/javascripts/core/api/BitriseYmlApi.tree.mswMocks.ts
@@ -78,34 +78,4 @@ export const postMergeConfig = (error?: string) => {
   });
 };
 
-export const postConfigPush = (result: 'ok' | 'conflict' | 'error' = 'ok') => {
-  return http.post('/api/app/:slug/config/push', async () => {
-    await delay();
-
-    if (result === 'error') {
-      return HttpResponse.json({ error_msg: 'validation_failed' }, { status: 422 });
-    }
-
-    if (result === 'conflict') {
-      return HttpResponse.json(
-        {
-          status: 'conflict',
-          conflicts: [{ path: 'modules/workflows.yml', remote: ROOT_NODE.includes[0] }],
-        },
-        { status: 409 },
-      );
-    }
-
-    return HttpResponse.json(
-      {
-        status: 'ok',
-        warnings: [],
-        root: ROOT_NODE,
-        entity_index: ENTITY_INDEX,
-      },
-      { status: 200 },
-    );
-  });
-};
-
 export const treeMocks = { ROOT_NODE, ENTITY_INDEX, MODULAR_RESPONSE, SINGLE_NODE_RESPONSE };
--- a/source/javascripts/core/api/BitriseYmlApi.ts
+++ b/source/javascripts/core/api/BitriseYmlApi.ts
@@ -1,9 +1,8 @@
 import {
   EntityIndex,
+  EntityIndexEntries,
   GetConfigResponse,
   MergedConfigResult,
-  SaveTreeConflict,
-  SaveTreeResult,
   TreeNode,
   TreeNodeSource,
 } from '@/core/models/Tree';
@@ -103,7 +102,6 @@ async function saveCiConfig({ data, version, tabOpenDuringSave, projectSlug, con
 
 const CONFIG_TREE_PATH = `/api/app/:projectSlug/config/tree`;
 const CONFIG_MERGE_PATH = `/api/app/:projectSlug/config/merge`;
-const CONFIG_PUSH_PATH = `/api/app/:projectSlug/config/push`;
 
 type WireTreeNodeSource = {
   path: string | null;
@@ -139,18 +137,6 @@ type WireGetConfigResponse = {
   branch?: string;
 };
 
-type WireSaveTreeResponse = {
-  status: 'ok';
-  warnings?: string[];
-  root: WireTreeNode;
-  entity_index: WireEntityIndex;
-};
-
-type WireSaveTreeConflict = {
-  status: 'conflict';
-  conflicts: Array<{ path: string; remote: WireTreeNode }>;
-};
-
 function mapSource(source: WireTreeNodeSource | null): TreeNodeSource | null {
   if (!source) {
     return null;
@@ -177,7 +163,15 @@ function wireToTreeNode(node: WireTreeNode): TreeNode {
   };
 }
 
-function treeNodeToWire(node: TreeNode): WireTreeNode {
+function treeNodeToWire(node: TreeNode, seen: Set<TreeNode> = new Set()): WireTreeNode {
+  // Identity-based cycle guard: this builds the save/merge payload, so a cyclic
+  // tree (an FE bug — wire payloads are JSON and can't contain cycles) must fail
+  // loudly instead of silently dropping nodes or blowing the stack.
+  if (seen.has(node)) {
+    throw new Error(`Cycle detected in include tree at ${node.path}`);
+  }
+  seen.add(node);
+
   return {
     node_id: node.nodeId,
     path: node.path,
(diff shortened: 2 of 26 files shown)


source/javascripts/core/api/BitriseYmlApi.ts
```diff
@@ -92,10 +101,231 @@ async function saveCiConfig({ data, version, tabOpenDuringSave, projectSlug, con
...
+  tabOpenDuringSave?: string;
+  conversationId?: string;
+}): Promise<SaveTreeResult> {
+  const path = CONFIG_PUSH_PATH.replace(':projectSlug', projectSlug);
+  const response = await Client.post<WireSaveTreeResponse>(path, {
+    body: JSON.stringify({
+      root: treeNodeToWire(tree),
+      tab_open_during_save: tabOpenDuringSave,
+      conversation_id: conversationId,
+    }),
```
moczolaszlo: This payload doesn't match the documented "Save tree" contract, which requires `branch`, `source_branch`, `commit_sha` (the conflict token) and `message` alongside `root`. Here we only send `root` + `tab_open_during_save` + `conversation_id`, so without `commit_sha` the documented 409 conflict detection (client sends the loaded `commit_sha`) can't fire.

Also note `BranchesApi.pushBranch({ root })` already POSTs to this same `/config/push` with the full branch/commit_sha/message set — which is what the contract's save-tree payload actually describes. So we now have two FE functions hitting `/config/push` with different shapes. Which is the canonical tree-save path? If `pushConfigTree` is meant to gain branch/commit_sha in a later PR, a quick note would help; as-is it's not contract-conformant.

source/javascripts/core/api/BitriseYmlApi.ts
```diff
@@ -92,10 +101,231 @@ async function saveCiConfig({ data, version, tabOpenDuringSave, projectSlug, con
...
+  if (!data || data.status !== 'conflict' || !Array.isArray(data.conflicts)) {
+    return undefined;
+  }
+
+  const conflicts = (data as unknown as WireSaveTreeConflict).conflicts;
+  return {
+    status: 'conflict',
+    conflicts: conflicts.map(({ path, remote }) => ({ path, remote: wireToTreeNode(remote) })),
+  };
+}
```
moczolaszlo: `mapSaveConflict` (and the `SaveTreeConflict` type) models the N-file conflict shape `{ conflicts: [{ path, remote }] }`, but the contract says iter-1 returns the single-file shape `{ status: 'conflict', commit_sha, remote_yml }` and explicitly defers the `conflicts[]` array to "a later phase". So against the iter-1 BE this would always hit the `status !== 'conflict' || !Array.isArray(conflicts)` guard and return `undefined` on a real conflict. Intentional forward-modeling, or should this match the iter-1 `remote_yml` shape for now?

source/javascripts/core/services/TreeService.ts
```diff
@@ -0,0 +1,126 @@
...
+    return {
+      ...node,
+      contents: liveNode?.contents ?? node.contents,
+      modified: liveNode?.modified ?? false,
+      includes: node.includes.map(rebuild),
+    };
+  };
+
+  return rebuild(root);
+}
```
moczolaszlo, line 71: Cycle-guard inconsistency: `walk` is explicitly cycle-guarded (with a dedicated test), and the BE has a `circular_include` error — so cycles are treated as a real, defended-against case. But `serializeTree` here, as well as `treeNodeToWire` (BitriseYmlApi) and `effectiveSourceLabel`, recurse `node.includes` with no `seen` set. The first two are on the save/merge path, so a malformed/cyclic tree would infinite-recurse exactly when building the save payload. Either guard these too, or — if `node_id` uniqueness makes cycles impossible — drop the guard from `walk` for consistency. (Separately: `effectiveSourceLabel` is exported and non-trivial but has no test.)

source/javascripts/core/services/EntityIndexService.ts
```diff
@@ -0,0 +1,75 @@
...
+          return;
+        }
+        // Pre-order append: first is top-most layer, later ones are lower layers.
+        (index[key][entityId] ||= []).push({ nodeId: node.nodeId });
+      });
+    });
+  });
+
+  return index;
+}
```
moczolaszlo, line 53: `buildFromFiles` (the live index builder that mirrors the BE — the heart of cross-file detection) and `equals` (used to skip no-op index updates) are the most logic-heavy functions in this service, but the spec only covers the trivial `definitionsOf` / `definingNodeId`. Given the `core/services` testing convention, these two deserve coverage (pre-order/merge-order accumulation, step_bundles camelCasing, empty/missing docs; equals true/false cases).

source/javascripts/core/api/BitriseYmlApi.ts
```diff
@@ -92,10 +101,231 @@ async function saveCiConfig({ data, version, tabOpenDuringSave, projectSlug, con
...
+    contents: node.contents,
+    source: node.source,
+    commit_sha: node.commitSha,
+    editable: node.editable,
+    modified: node.modified,
+    includes: node.includes.map(treeNodeToWire),
+  };
+}
+
+function mapEntityEntries(entries: WireEntityEntries = {}): EntityIndex['workflows'] {
```
moczolaszlo: Nit: the return type `EntityIndex['workflows']` works (it's structurally `EntityIndexEntries`), but since this helper is used for all three kinds, the named `EntityIndexEntries` reads more clearly than the workflows-specific accessor.

moczolaszlo (commented):
Reviewed the full core layer against the API-contract doc. Solid, well-tested, and genuinely zero-runtime-impact (nothing imports it yet) — nice scaffolding. A few questions where the FE shapes seem to diverge from the documented contract, plus two test-coverage gaps. None are blocking given the code is dormant, but worth settling before PRs 3–7 build on it.

source/javascripts/core/api/BitriseYmlApi.ts
trapacska: You're right — and tracing it through the stack settled it: `useSaveCiConfigTree` (the only consumer) had zero callers anywhere in PRs 2–7. The canonical tree-save path is `BranchesApi.pushBranch({root})` (Header → PushBranchDialog → `usePushBranch`), which sends exactly the documented payload (`branch`, `source_branch`, `commit_sha`, `message`, `root`). Rather than keep a second, non-conformant `/config/push` client around, I removed `pushConfigTree` entirely in f2e9cffb — a save-in-place client can be (re)introduced with the conflict-management work, modeled on the final contract. (Fix propagated to PRs 3–7 via forward merges, no force pushes.)

source/javascripts/core/api/BitriseYmlApi.ts
trapacska: Good catch — it was forward-modeling the deferred N-file shape, so against the iter-1 BE it would indeed return `undefined` on a real conflict. Mooted by removing the dead save-in-place client (f2e9cffb): `mapSaveConflict` and the `SaveTreeConflict`/`SaveTreeResult` types are gone. The live 409 handling in `usePushBranch` only branches on the status code and opens the merge dialog, which works with the iter-1 shape. Proper conflict management is still to be developed after the stack merges, and the FE types will be written against the contract shape that actually ships.

source/javascripts/core/services/TreeService.ts
trapacska: Done in f2e9cffb. Kept the guards (worth noting: a JSON-parsed payload can't contain object-graph cycles, so the realistic cycle source is an FE bug mutating the tree — cheap insurance either way), with intent matched to the call site: `effectiveSourceLabel` short-circuits via a seen-set like `walk`, while `serializeTree`/`treeNodeToWire` use an identity-based guard that **throws** — they build the save/merge payload, where silently skipping a node would corrupt the save, so failing loudly into the error toast is the right behavior. `effectiveSourceLabel` also got its missing spec (own ref, inherited ancestor ref, null cases, cycle).

source/javascripts/core/services/EntityIndexService.ts
trapacska: Done in f2e9cffb — added `buildFromFiles` coverage (pre-order multi-layer accumulation with top-most first, `step_bundles` → `stepBundles` camelCasing, nodes without loaded docs, undefined tree) and `equals` true/false cases (identical-by-structure, added/removed/renamed entity, layer length + order differences).

source/javascripts/core/api/BitriseYmlApi.ts
trapacska: Done in f2e9cffb — `mapEntityEntries` now returns the named `EntityIndexEntries`.

source/javascripts/core/services/TreeService.ts
```diff
@@ -0,0 +1,126 @@
...
+      return next;
+    }
+    for (const child of node.includes) {
+      const found = find(child, next);
+      if (found) {
+        return found;
+      }
+    }
+    return null;
+  };
```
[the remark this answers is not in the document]
trapacska: Fixed in f2e9cffb — `find` is now cycle-guarded with a seen-set like `walk` (and `serializeTree`/`treeNodeToWire` got identity-based guards that throw, since they build the save payload). Covered by new specs.

source/javascripts/core/api/BitriseYmlApi.ts
```diff
@@ -92,10 +101,231 @@ async function saveCiConfig({ data, version, tabOpenDuringSave, projectSlug, con
...
+  projectSlug,
+  tree,
+  branch,
+  signal,
+}: {
+  projectSlug: string;
+  tree: TreeNode;
+  branch?: string;
+  signal?: AbortSignal;
+}): Promise<MergedConfigResult> {
```
moczolaszlo: Convention: multi-param API functions take a named `*Options` type (`GetCiConfigOptions`, and `GetConfigOptions` just above). `getMergedConfig` and `pushConfigTree` both use inline object types for 4 params — inconsistent even within this file. Suggest `GetMergedConfigOptions` / `PushConfigTreeOptions` to match.

source/javascripts/core/api/BitriseYmlApi.ts
```diff
@@ -92,10 +101,231 @@ async function saveCiConfig({ data, version, tabOpenDuringSave, projectSlug, con
...
   getCiConfig,
   ciConfigPath,
   saveCiConfig,
+  getConfig,
+  configTreePath,
+  getMergedConfig,
+  pushConfigTree,
+  mapSaveConflict,
+  // Exposed so push-to-branch can serialize a full tree to the wire shape.
+  treeNodeToWire,
```
moczolaszlo: Mappers elsewhere use `from*Response` / `to*` naming and stay file-private (e.g. SecretApi's `fromMonolithResponse` / `toMonolithUpdateRequest`). Here the naming idiom differs (`wireToTreeNode` / `treeNodeToWire`) and `treeNodeToWire` is exported so `BranchesApi` can import it — no existing client imports another client's mapper, so this is a new cross-API pattern. Minor, but worth a conscious call (e.g. a small shared serializer, or having `BranchesApi` accept an already-serialized wire node).

source/javascripts/core/models/Tree.ts
```diff
@@ -0,0 +1,77 @@
...
+  conflicts: TreeConflict[];
+};
+
+export type MergedConfigResult = {
+  mergedYml: string;
+};
+
+export function emptyEntityIndex(): EntityIndex {
+  return { workflows: {}, pipelines: {}, stepBundles: {} };
+}
```
moczolaszlo: Files in `core/models` are type-only by convention; factories/empties live in services or consts (e.g. `EMPTY_ENV_VAR` in EnvVarService, not the model). `emptyEntityIndex()` here diverges — and `EntityIndexService.buildFromFiles` re-inlines `{ workflows: {}, pipelines: {}, stepBundles: {} }` instead of calling it, so the factory both breaks the convention and isn't used where it should be. Either drop it and inline at the call site, or keep it and use it in `buildFromFiles`.

source/javascripts/core/api/BitriseYmlApi.tree.mswMocks.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (error) {
+      return HttpResponse.json({ error_msg: error }, { status: 422 });
+    }
+
+    return HttpResponse.json(variant === 'modular' ? MODULAR_RESPONSE : SINGLE_NODE_RESPONSE, { status: 200 });
+  });
+};
+
+export const postMergeConfig = (error?: string) => {
+  return http.post('/api/app/:slug/config/merge', async () => {
```
moczolaszlo: `getConfigTree` reuses `BitriseYmlApi.configTreePath(...)` (line 58), but the merge/push paths are inlined as raw strings here — and in `BitriseYmlApi` itself (`CONFIG_MERGE_PATH.replace` / `CONFIG_PUSH_PATH.replace`) there's no exported getter for them. House style exposes `get*Path()` getters (`ciConfigPath`, `configTreePath`, SecretApi's path getters) precisely so mocks reuse them and can't drift from the real route. Suggest adding `mergeConfigPath` / `pushConfigPath` getters and using them here.

source/javascripts/core/api/BitriseYmlApi.tree.mswMocks.ts
```diff
@@ -0,0 +1,111 @@
...
+};
+
+const MERGED_YML = 'format_version: "13"\nworkflows:\n  build:\n    steps:\n      - script@1: {}\n';
+
+const MODULAR_RESPONSE = {
+  root: ROOT_NODE,
+  entity_index: ENTITY_INDEX,
+  merged_yml: MERGED_YML,
+  branch: 'main',
+};
```
moczolaszlo: Existing MSW mocks assert response bodies with `satisfies <WireType>` (SecretApi.mswMocks, EnvVarsApi.mswMocks) so a mock↔wire drift fails at compile time. These bodies are untyped literals because the wire types (`WireGetConfigResponse`, `WireSaveTreeResponse`, …) are file-private in BitriseYmlApi. Consider exporting those wire types as named exports and adding `satisfies` to these fixtures — it'd also have surfaced the conflict-shape question from the earlier review.

moczolaszlo (commented):
Follow-up pass comparing this against existing repo conventions (how API clients, services, models, and MSW mocks are written elsewhere). The skeleton fits well — default-export pure-function services, SCREAMING_SNAKE path consts, Client.raw/post usage, nested describe/it, named MSW handlers all match. Below are the concrete, mechanical spots where it drifts from the established house style; each maps to an existing pattern you can mirror. Two things I'm deliberately leaving as judgment calls rather than inline asks: (a) comment density here is much higher than sibling services in core/services (near-zero JSDoc there, JSDoc reserved for YmlUtils) — fine if intentional for this domain; (b) TreeService.walk adds a cycle `seen`-guard while the repo's other recursive services (getStepBundleChain, getBeforeRunChain) recurse naively — so pick one direction (and if you keep guarding, serializeTree/treeNodeToWire on the save path should guard too).

source/javascripts/core/api/BitriseYmlApi.ts
trapacska: Done in f1b93143 — `getMergedConfig` now takes a named `GetMergedConfigOptions`. (`pushConfigTree` no longer exists — removed in f2e9cffb per the round-1 contract discussion.)

source/javascripts/core/api/BitriseYmlApi.ts
trapacska: Renamed to the house `from*`/`to*` idiom in f1b93143: `fromWireTreeNode` / `toWireTreeNode` / `fromWireEntityIndex`. On the cross-client import: made it a conscious, documented call — `BitriseYmlApi` stays the single owner of the tree wire shape and `BranchesApi` imports `toWireTreeNode` (comment at the export explains why). Having `BranchesApi` accept a pre-serialized wire node would just move the import into a hook, and a copy risks drift between the two `/config` payloads.

source/javascripts/core/models/Tree.ts
trapacska: Done in f1b93143 — `emptyEntityIndex` moved out of the type-only model into `EntityIndexService`, and `buildFromFiles` now uses it instead of re-inlining the literal. (The store call site in PR 3 was updated in a follow-up commit there.)

source/javascripts/core/api/BitriseYmlApi.tree.mswMocks.ts
trapacska: Done in f1b93143 — added a `mergeConfigPath` getter and the MSW merge mock now reuses it, so the route can't drift. (No push getter needed: `pushConfigTree` + its mock were removed in f2e9cffb; the push route lives in `BranchesApi.pushBranch`.)

source/javascripts/core/api/BitriseYmlApi.tree.mswMocks.ts
trapacska: Done in f1b93143 — `WireTreeNode`/`WireEntityIndex`/`WireGetConfigResponse` are exported and all four fixtures are `satisfies`-checked. Agreed it would have caught the conflict-shape drift — good instinct.

trapacska:
Re the two judgment calls from the conventions pass: (a) **comment density** — intentional for this layer; the tree/include domain is new to the codebase and the comments capture contract decisions (identity keys, conflict tokens, merge order) that aren't recoverable from the code. Happy to thin them out if they age badly. (b) **cycle guards** — keeping the guarded direction, now applied consistently (round-1 fix f2e9cffb): `walk`/`effectiveSourceLabel` short-circuit, `serializeTree`/`toWireTreeNode` throw on the save path. The existing naive recursions (`getStepBundleChain` etc.) recurse over user-authored YAML where the BE already rejects cycles; the tree payload is a new BE-owned shape, so defense here is cheap and local. Round-2 fixes are in f1b93143 (branch 2) + a small follow-up on branch 3 for the moved factory.

source/javascripts/core/api/BitriseYmlApi.ts
moczolaszlo: Confirmed — `pushConfigTree` removed in f2e9cffb and `BranchesApi.pushBranch({ root })` is the canonical, contract-conformant tree-save path. Resolving.

source/javascripts/core/api/BitriseYmlApi.ts
moczolaszlo: Confirmed — `mapSaveConflict` + the `SaveTreeConflict`/`SaveTreeResult` types are gone; iter-1 409s are handled by status code in `usePushBranch`. Resolving.

source/javascripts/core/services/TreeService.ts
moczolaszlo: Confirmed — `serializeTree`/`toWireTreeNode` now throw on cycles and `effectiveSourceLabel` short-circuits via a seen-set, all with specs. The internal inconsistency is gone. Resolving.

source/javascripts/core/services/EntityIndexService.ts
moczolaszlo: Confirmed — `buildFromFiles` and `equals` are now covered (pre-order multi-layer, step_bundles→stepBundles, missing docs, undefined tree, reorder). Resolving.

source/javascripts/core/api/BitriseYmlApi.ts
moczolaszlo: Confirmed — now returns the named `EntityIndexEntries`. Resolving.

source/javascripts/core/api/BitriseYmlApi.ts
moczolaszlo: Confirmed — `GetMergedConfigOptions` added. Resolving.

source/javascripts/core/api/BitriseYmlApi.ts
moczolaszlo: Confirmed — renamed to the house `fromWire*`/`toWire*` idiom; the cross-client import is kept as a documented single-owner decision, which is reasonable. Resolving.

source/javascripts/core/models/Tree.ts
moczolaszlo: Confirmed — `emptyEntityIndex` moved into `EntityIndexService` and `buildFromFiles` now uses it. Resolving.

source/javascripts/core/api/BitriseYmlApi.tree.mswMocks.ts
moczolaszlo: Confirmed — `mergeConfigPath` getter added and reused by the MSW mock. Resolving.

source/javascripts/core/api/BitriseYmlApi.tree.mswMocks.ts
moczolaszlo: Confirmed — wire types exported and all four fixtures are `satisfies`-checked. Resolving.

source/javascripts/core/services/TreeService.ts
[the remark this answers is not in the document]
moczolaszlo: Same issue I flagged independently — verified fixed in f2e9cffb (`find` is cycle-guarded; `serializeTree`/`toWireTreeNode` throw on the save path), covered by new specs. Resolving.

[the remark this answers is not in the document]
moczolaszlo: Duplicate of the effectiveSourceLabel cycle-guard finding (also raised in the sibling bot thread and my own review). Verified fixed in f2e9cffb with a `seen`-set guard + new specs. Resolving.

moczolaszlo (approved):
Approving. All feedback from both review rounds is addressed and verified in f2e9cffb / f1b931438c, every review thread is resolved, and CI is green.

Highlights of the resolution:
- Dropped the dead, non-contract-conformant `pushConfigTree` + `mapSaveConflict`/`SaveTreeConflict` — `BranchesApi.pushBranch({ root })` is the canonical, contract-conformant tree-save path.
- Cycle guards added on all recursive write/label paths: `serializeTree`/`toWireTreeNode` throw (save payload — fail loud), `effectiveSourceLabel` short-circuits like `walk`; all covered by specs.
- `EntityIndexService.buildFromFiles` + `equals` now thoroughly tested.
- House-style alignment: `from*`/`to*` mapper naming, named `GetMergedConfigOptions`, `mergeConfigPath` getter reused by the mock, `emptyEntityIndex` moved into the service (and used), exported wire types with `satisfies`-checked fixtures.

Clean, low-risk core layer (still dormant — nothing imports it yet). Nice work.

PR #1788, opened 2026-06-11
BIVS-3677 [3/7] Store: modular state alongside legacy single-file state

trapacska:
Part **3/7** of the modular YAML editor stack. Stacked on #1787.

## What
- `BitriseYmlStore`: modular state (`tree`, per-file slices, open tabs, entity index, merged config) plus `initializeModularConfig`, `createFile`, `addInclude`, tab management, modular save/discard — with a 699-line spec
- `WorkflowService.assertWorkflowReferenceable`: accepts workflows known to the entity index (cross-file) when validating chaining; used by `addChainedWorkflow` and `PipelineService.addWorkflowToPipeline`
- `ContainerService`: returns `undefined` instead of throwing for step-bundle definitions absent from the active doc (cross-file bundles must not crash during render)

## Safety
**Mostly dormant** — nothing calls `initializeModularConfig` yet (that's PR 7), so `state.tree` is always `undefined`. Every touched legacy store function (`updateBitriseYmlDocument`, `discardBitriseYmlDocument`, `hasChanges`, …) branches on `tree` presence and keeps its exact old behavior; with an empty entity index the relaxed `WorkflowService` check is equivalent to the old one. Review focus: the `if (!state.tree)` fast paths.

### ⚠️ One live behavior change (affects prod, not gated on `tree`)
`ContainerService` step-bundle lookup now returns `undefined` instead of throwing when a `bundle::` definition is absent from the active doc. This ships to **all** users immediately. It's safe — the sole caller (`useContainerReferences`) already types the result as `ContainerReference[] | undefined` and handles it — and it's arguably a fix: previously a step referencing a missing/typo'd bundle definition would throw here and crash the ContainersTab card during render. Called out so QA/reviewers know one path is not dormant.

### DIFF
--- a/source/javascripts/core/services/ContainerService.ts
+++ b/source/javascripts/core/services/ContainerService.ts
@@ -259,7 +259,11 @@ function getContainerReferences(type: ContainerType, yamlMap: YAMLMap): Containe
 }
 
 function getContainerReferencesFromStepBundleDefinition(sourceId: string, type: ContainerType, doc: Document) {
-  const yamlMap = StepBundleService.getStepBundleOrThrowError(doc, sourceId);
+  // A cross-file bundle definition is absent here; return none rather than throwing (throwing crashes the card during render).
+  const yamlMap = YmlUtils.getMapIn(doc, ['step_bundles', sourceId]);
+  if (!yamlMap) {
+    return undefined;
+  }
 
   return getContainerReferences(type, yamlMap);
 }
--- a/source/javascripts/core/services/WorkflowService.spec.ts
+++ b/source/javascripts/core/services/WorkflowService.spec.ts
@@ -1087,6 +1087,11 @@ describe('WorkflowService', () => {
   describe('addChainedWorkflow', () => {
     const placements: ChainedWorkflowPlacement[] = ['after_run', 'before_run'];
 
+    // Reset regardless of assertion failures, so a populated index can't leak into later tests.
+    afterEach(() => {
+      bitriseYmlStore.setState({ entityIndex: { workflows: {}, pipelines: {}, stepBundles: {} } });
+    });
+
     placements.forEach((placement) => {
       describe(`when placement is '${placement}'`, () => {
         it('should create placement with chained workflow', () => {
@@ -1231,8 +1236,6 @@ describe('WorkflowService', () => {
         WorkflowService.addChainedWorkflow('wf1', 'after_run', 'cross-file-wf');
       }).not.toThrow();
       expect(getYmlString()).toContain('cross-file-wf');
-
-      bitriseYmlStore.setState({ entityIndex: { workflows: {}, pipelines: {}, stepBundles: {} } });
     });
   });
 
--- a/source/javascripts/core/services/WorkflowService.spec.ts
+++ b/source/javascripts/core/services/WorkflowService.spec.ts
@@ -1,6 +1,6 @@
 import { Workflows } from '../models/BitriseYml';
 import { ChainedWorkflowPlacement } from '../models/Workflow';
-import { getYmlString, updateBitriseYmlDocumentByString } from '../stores/BitriseYmlStore';
+import { bitriseYmlStore, getYmlString, updateBitriseYmlDocumentByString } from '../stores/BitriseYmlStore';
 import WorkflowService from './WorkflowService';
 
 describe('WorkflowService', () => {
@@ -1214,6 +1214,25 @@ describe('WorkflowService', () => {
       expect(() => {
         WorkflowService.addChainedWorkflow('wf1', 'invalid_placement' as ChainedWorkflowPlacement, 'wf2');
       }).toThrow(`Invalid placement: invalid_placement. It should be 'before_run' or 'after_run'.`);
+    });
+
+    it('chains a cross-file workflow (defined in another module) without throwing', () => {
+      updateBitriseYmlDocumentByString(
+        yaml`
+          workflows:
+            wf1: {}
+        `,
+      );
+      bitriseYmlStore.setState({
+        entityIndex: { workflows: { 'cross-file-wf': [{ nodeId: 'n_other' }] }, pipelines: {}, stepBundles: {} },
+      });
+
+      expect(() => {
+        WorkflowService.addChainedWorkflow('wf1', 'after_run', 'cross-file-wf');
+      }).not.toThrow();
+      expect(getYmlString()).toContain('cross-file-wf');
+
+      bitriseYmlStore.setState({ entityIndex: { workflows: {}, pipelines: {}, stepBundles: {} } });
     });
   });
 
--- a/source/javascripts/core/services/WorkflowService.ts
+++ b/source/javascripts/core/services/WorkflowService.ts
@@ -3,8 +3,9 @@ import { Document, isMap, isSeq } from 'yaml';
 
 import { Pipelines, Stages, WorkflowModel, Workflows } from '../models/BitriseYml';
 import { ChainedWorkflowPlacement } from '../models/Workflow';
-import { updateBitriseYmlDocument } from '../stores/BitriseYmlStore';
+import { bitriseYmlStore, updateBitriseYmlDocument } from '../stores/BitriseYmlStore';
 import YmlUtils from '../utils/YmlUtils';
+import EntityIndexService from './EntityIndexService';
 
 const WORKFLOW_NAME_REGEX = /^[A-Za-z0-9-_.]+$/;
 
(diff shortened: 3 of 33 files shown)


moczolaszlo (commented):
Reviewed the store layer + the cross-file-tolerant service changes. Verified the dormant claim holds: with `tree === undefined` every legacy path (discard / updateBitriseYmlDocument(ByString) / hasChanges) keeps its exact old behavior, the `initializeBitriseYmlDocument` rewrite is behavior-identical, and `assertWorkflowReferenceable` is equivalent to `getWorkflowOrThrowError` when the entity index is empty. No import cycle. The 699-line modular spec is genuinely thorough (createFile, addInclude validation, tabs, active-file binding, discard incl. session-file drop, live entity index, tree-wide hasChanges, applyModularSaveResult). A few notes below — all non-blocking, mostly framing/questions.

source/javascripts/core/services/ContainerService.ts
```diff
@@ -259,7 +259,11 @@ function getContainerReferences(type: ContainerType, yamlMap: YAMLMap): Containe
...
 function getContainerReferencesFromStepBundleDefinition(sourceId: string, type: ContainerType, doc: Document) {
-  const yamlMap = StepBundleService.getStepBundleOrThrowError(doc, sourceId);
+  // A cross-file bundle definition is absent here; return none rather than throwing (throwing crashes the card during render).
+  const yamlMap = YmlUtils.getMapIn(doc, ['step_bundles', sourceId]);
+  if (!yamlMap) {
+    return undefined;
+  }
 
   return getContainerReferences(type, yamlMap);
 }
```
moczolaszlo, line 269: Worth calling out that this throw→undefined change is **not** dormant — unlike the store paths it isn't gated on `tree`, so it ships to all users now. It is safe: the only caller (`useContainerReferences`) already types this as `ContainerReference[] | undefined` and the downstream handles it. It's arguably a fix — previously a step referencing a missing/typo'd `bundle::` definition would throw here and crash the ContainersTab card during render. Just suggesting the PR description move this out from under the "Dormant" safety bullet so reviewers/QA know one behavior change is live.

source/javascripts/core/services/WorkflowService.ts
```diff
@@ -155,6 +155,19 @@ function getWorkflowOrThrowError(id: string, doc: Document) {
...
+  if (YmlUtils.getMapIn(doc, ['workflows', id])) {
+    return;
+  }
+  const isCrossFile = Boolean(
+    EntityIndexService.definingNodeId(bitriseYmlStore.getState().entityIndex, 'workflows', id),
+  );
+  if (!isCrossFile) {
+    getWorkflowOrThrowError(id, doc);
+  }
+}
```
moczolaszlo, line 169: Heads up that this makes the result depend on global store state (`bitriseYmlStore.getState().entityIndex`), not just the `doc` argument — a new flavor of coupling for an otherwise doc-pure service function (the existing store coupling here is mutation via `updateBitriseYmlDocument`, not reading derived state). Equivalent to the old `getWorkflowOrThrowError` while the index is empty, so it's safe today; just flagging it as a conscious design choice rather than an oversight. (Alternative would be passing the relevant index in, but that ripples through the call sites — fine to keep as-is.)

source/javascripts/core/stores/BitriseYmlStore.ts
```diff
@@ -90,36 +228,532 @@ export function initializeBitriseYmlDocument({
...
   const doc = YmlUtils.toDoc(ymlString);
 
-  if (doc.errors.length === 0) {
+  bitriseYmlStore.setState({
+    version,
+    configBranch: branch || undefined,
+    configCommitSha: commitSha || undefined,
+    ...(doc.errors.length === 0
+      ? { ymlDocument: doc, savedYmlDocument: doc, __invalidYmlString: undefined, __savedInvalidYmlString: undefined }
+      : { __invalidYmlString: ymlString, __savedInvalidYmlString: ymlString }),
```
moczolaszlo, line 237: Question for the activation PR (PR 7), not this one: `initializeBitriseYmlDocument` doesn't touch modular state, so a modular→single-file switch through the legacy init would leave stale `tree` / `files` / `openTabs` / `entityIndex` behind. Harmless now (init never sets `tree`). Is the teardown handled elsewhere on config/branch switch, or should the legacy init reset modular state when it runs?

source/javascripts/core/services/WorkflowService.spec.ts
```diff
@@ -1215,6 +1215,25 @@ describe('WorkflowService', () => {
...
+      bitriseYmlStore.setState({
+        entityIndex: { workflows: { 'cross-file-wf': [{ nodeId: 'n_other' }] }, pipelines: {}, stepBundles: {} },
+      });
+
+      expect(() => {
+        WorkflowService.addChainedWorkflow('wf1', 'after_run', 'cross-file-wf');
+      }).not.toThrow();
+      expect(getYmlString()).toContain('cross-file-wf');
+
+      bitriseYmlStore.setState({ entityIndex: { workflows: {}, pipelines: {}, stepBundles: {} } });
```
moczolaszlo: Minor test hygiene: this resets `entityIndex` at the end of the `it` body, so if either assertion above throws the reset is skipped and the populated index leaks into later tests. An `afterEach` (or `try/finally`) reset would be robust regardless of failure.

source/javascripts/core/stores/BitriseYmlStore.ts
trapacska: Confirmed — the legacy init didn't touch modular state, so a modular session routing through a legacy path (branch switch, repository-YAML save, manual update) could leave a stale tree/files/openTabs/entityIndex behind. Handled in PR 7 (where the two init paths coexist): `initializeBitriseYmlDocument` now resets all modular fields to empty via a `clearedModularState()` helper — a no-op in pure legacy mode, defensive on a modular→single-file transition. Added a store spec covering it. (Within a single session the bootstrap flag is fixed, so there's no live mode-flip; this guards the explicit legacy-path calls.)

source/javascripts/core/services/WorkflowService.spec.ts
trapacska: Done — replaced the inline reset with an `afterEach` on the `addChainedWorkflow` describe, so the index can't leak even if an assertion throws. Thanks.

source/javascripts/core/services/WorkflowService.ts
trapacska: Agreed it's a conscious choice, not an oversight. `assertWorkflowReferenceable` reads `entityIndex` from the store so a chained workflow defined in another module (cross-file) isn't rejected as missing; while the index is empty (flag off) it's exactly equivalent to the old `getWorkflowOrThrowError`. Passing the index in would ripple through call sites for no behavioral gain today — keeping as-is, flagged for when we revisit the service-purity boundary.

source/javascripts/core/services/ContainerService.ts
trapacska: Good call — updated the PR description: moved this out from under the "Dormant" bullet into a dedicated **⚠️ One live behavior change (affects prod)** section, noting it ships to all users now, that the sole caller already handles `undefined`, and that it's effectively a fix for the missing/typo'd-bundle render crash. Thanks for catching the framing.

moczolaszlo: Thanks — the dedicated "one live behavior change" section is exactly the framing I was after, and the change itself is safe. Resolving.

source/javascripts/core/services/WorkflowService.ts
moczolaszlo: Agreed — conscious choice, recorded rationale, equivalent to the old behavior while the index is empty. Resolving.

source/javascripts/core/stores/BitriseYmlStore.ts
moczolaszlo: Great — appropriately handled in PR 7 with `clearedModularState()` + spec; nothing needed in this PR. Resolving.

source/javascripts/core/services/WorkflowService.spec.ts
moczolaszlo: Confirmed fixed in aa4f794107 — `afterEach` reset on the describe (verified). Resolving.

moczolaszlo (approved):
Approving. All four review notes addressed: the live ContainerService behavior change is now called out in the description (and is safe — sole caller already handles `undefined`), the `assertWorkflowReferenceable` store coupling is a documented conscious choice (equivalent to the old behavior while the index is empty), the legacy/modular teardown gap is handled in PR 7 via `clearedModularState()`, and the spec now resets via `afterEach`. Verified the dormant guarantee holds and the 699-line modular spec is thorough. Nice work.

PR #1790, opened 2026-06-11
BIVS-3677 [5/7] UI: read-only views + cross-file provenance threading

trapacska:
Part **5/7** of the modular YAML editor stack. Stacked on the hooks PR. Biggest PR by file count, but mechanically repetitive.

## What
- New components: `CrossFileProvenanceText`, `JumpToDefinitionLink` + `CrossFileJumpButton`, `CrossFileDefinitionCard`, `ReadOnlyViewNotification`, `FileTreeView`
- Read-only gating (`useIsReadOnlyView` → disabled CTAs/inputs/menus) across env vars, stack & machine, containers, triggers, step/workflow/step-bundle config, and page toolbars
- `WorkflowCardContext` strips mutation callbacks in read-only views (cards hide controls by callback presence)
- Workflow/step-bundle cards show cross-file provenance + jump-to-definition on hover
- Dialog close-on-file-switch wiring on the Workflows / Pipelines / StepBundles pages

## Safety
**Dormant**: `useIsReadOnlyView` / `useCrossFileEntity` always report editable/local while `store.tree` is unset, so all gating is inert today.

## ⚠️ Touches active prod code — review these closely, verify with flag off
- `StepConfigDrawer` / `WorkflowConfig` / `StepBundleConfig` **providers extracted** into separate `*Provider.tsx` files (pure moves for Fast Refresh + reuse — should be behavior-identical)
- `StepList` empty-state title no longer depends on `onAdd` presence

Manual QA with flag off: open/edit/save in the step config drawer, workflow config drawer/panel, step bundle drawer.

### DIFF
--- a/source/javascripts/components/FileTreeViewer/FileTreeView.tsx
+++ b/source/javascripts/components/FileTreeViewer/FileTreeView.tsx
@@ -1,0 +1,154 @@
+import { Box, Text } from '@bitrise/bitkit';
+import {
+  BitkitIconButton,
+  BitkitTreeView,
+  createTreeCollection,
+  IconFileYml,
+  IconInfoCircle,
+} from '@bitrise/bitkit-v2';
+import { useMemo } from 'react';
+
+import { TreeNode } from '@/core/models/Tree';
+import FileTreeService, { FileTreeFolder, FileTreeGroup } from '@/core/services/FileTreeService';
+import PageProps from '@/core/utils/PageProps';
+
+type Props = {
+  /** The real tree root; grouping + folder layout are derived from it. */
+  rootNode: TreeNode;
+  /** When set, only files matching this predicate are shown (e.g. dirty-only). */
+  filter?: (node: TreeNode) => boolean;
+  selectedNodeId?: string;
+  onSelect: (nodeId: string) => void;
+  /** Optional predicate to disable (non-selectable, greyed) individual files. */
+  isNodeDisabled?: (node: TreeNode) => boolean;
+};
+
+// One collection node: group root, folder, or file leaf. Folder ids are path strings, file ids are node_ids.
+type RenderNode = {
+  id: string;
+  label: string;
+  isFile: boolean;
+  node?: TreeNode;
+  children: RenderNode[];
+};
+
+const GROUP_ROOT_ID = '__group_root__';
+
+function toRenderNode(folder: FileTreeFolder): RenderNode[] {
+  const folders = folder.folders.map(
+    (child): RenderNode => ({ id: child.id, label: child.label, isFile: false, children: toRenderNode(child) }),
+  );
+  const files = folder.files.map(
+    (file): RenderNode => ({
+      id: file.nodeId,
+      label: file.fileName,
+      isFile: true,
+      node: file.node,
+      children: [],
+    }),
+  );
+  // Files before folders at every level, matching the design.
+  return [...files, ...folders];
+}
+
+function collectFolderIds(nodes: RenderNode[], acc: string[]): string[] {
+  nodes.forEach((node) => {
+    if (!node.isFile) {
+      acc.push(node.id);
+      collectFolderIds(node.children, acc);
+    }
+  });
+  return acc;
+}
+
+const GroupTree = ({
+  group,
+  selectedNodeId,
+  onSelect,
+  isNodeDisabled,
+}: {
+  group: FileTreeGroup;
+} & Pick<Props, 'selectedNodeId' | 'onSelect' | 'isNodeDisabled'>) => {
+  const children = useMemo(() => toRenderNode(group.root), [group.root]);
+
+  const collection = useMemo(
+    () =>
+      createTreeCollection<RenderNode>({
+        rootNode: { id: GROUP_ROOT_ID, label: '', isFile: false, children },
+        nodeToValue: (node) => node.id,
+        nodeToString: (node) => node.label,
+        nodeToChildren: (node) => node.children,
+        isNodeDisabled: (node) => Boolean(node.isFile && node.node && isNodeDisabled?.(node.node)),
+      }),
+    [children, isNodeDisabled],
+  );
+
+  const expandedValue = useMemo(() => [GROUP_ROOT_ID, ...collectFolderIds(children, [])], [children]);
+
+  return (
+    <BitkitTreeView
+      variant="files"
+      collection={collection}
+      selectionMode="single"
+      selectedValue={selectedNodeId ? [selectedNodeId] : []}
+      defaultExpandedValue={expandedValue}
+      onSelectionChange={(details) => {
+        const id = details.selectedValue[0];
+        const selected = id ? collection.findNode(id) : undefined;
+        if (selected?.isFile) {
+          onSelect(selected.id);
+        }
+      }}
+      render={({ node, nodeState }) =>
+        nodeState.isBranch ? (
+          <BitkitTreeView.Branch label={node.label} />
+        ) : (
+          <BitkitTreeView.Leaf label={node.label} icon={IconFileYml} />
+        )
+      }
+    />
+  );
+};
+
+/**
+ * Shared file/folder selector (Open module popover, global diff switcher, jump-to-def
+ * chooser). Files grouped by effective repository (working repo first) and laid out by
+ * path with single-child chains collapsed. Pure presentation — shaping lives in `FileTreeService`.
+ */
+const FileTreeView = ({ rootNode, filter, selectedNodeId, onSelect, isNodeDisabled }: Props) => {
+  const projectRepoLabel = PageProps.app()?.gitRepoSlug || PageProps.app()?.name || 'This repository';
+
+  const groups = useMemo(
+    () => FileTreeService.buildFileTree(rootNode, { projectRepoLabel, filter }),
+    [rootNode, projectRepoLabel, filter],
+  );
+
+  return (
+    <Box display="flex" flexDirection="column" gap="16">
+      {groups.map((group) => (
+        <Box key={group.key} display="flex" flexDirection="column">
+          {/* No `px`: header left-aligns with the title while the tree's own 8px inset indents file rows. */}
+          <Box display="flex" alignItems="center" gap="4" py="6">
+            <Text textStyle="body/md/semibold">{group.header}</Text>
+            {group.isReadOnly && (
+              <BitkitIconButton
+                size="sm"
+                variant="tertiary"
+                label="Modules included from another repository or ref are read-only."
+                icon={IconInfoCircle}
+              />
+            )}
+          </Box>
+          <GroupTree
+            group={group}
+            selectedNodeId={selectedNodeId}
+            onSelect={onSelect}
+            isNodeDisabled={isNodeDisabled}
+          />
+        </Box>
+      ))}
+    </Box>
+  );
+};
+
+export default FileTreeView;
--- a/source/javascripts/components/FileTreeViewer/FileTreeView.tsx
+++ b/source/javascripts/components/FileTreeViewer/FileTreeView.tsx
@@ -13,17 +13,14 @@ import FileTreeService, { FileTreeFolder, FileTreeGroup } from '@/core/services/
 import PageProps from '@/core/utils/PageProps';
 
 type Props = {
-  /** The real tree root; grouping + folder layout are derived from it. */
   rootNode: TreeNode;
-  /** When set, only files matching this predicate are shown (e.g. dirty-only). */
   filter?: (node: TreeNode) => boolean;
   selectedNodeId?: string;
   onSelect: (nodeId: string) => void;
-  /** Optional predicate to disable (non-selectable, greyed) individual files. */
   isNodeDisabled?: (node: TreeNode) => boolean;
 };
 
-// One collection node: group root, folder, or file leaf. Folder ids are path strings, file ids are node_ids.
+// Folder ids are path strings, file ids are node_ids.
 type RenderNode = {
   id: string;
   label: string;
@@ -47,7 +44,6 @@ function toRenderNode(folder: FileTreeFolder): RenderNode[] {
       children: [],
     }),
   );
-  // Files before folders at every level, matching the design.
   return [...files, ...folders];
 }
 
@@ -110,11 +106,6 @@ const GroupTree = ({
   );
 };
 
-/**
- * Shared file/folder selector (Open module popover, global diff switcher, jump-to-def
- * chooser). Files grouped by effective repository (working repo first) and laid out by
- * path with single-child chains collapsed. Pure presentation — shaping lives in `FileTreeService`.
- */
 const FileTreeView = ({ rootNode, filter, selectedNodeId, onSelect, isNodeDisabled }: Props) => {
   const projectRepoLabel = PageProps.app()?.gitRepoSlug || PageProps.app()?.name || 'This repository';
 
--- a/source/javascripts/components/JumpToDefinitionLink/JumpToDefinitionLink.tsx
+++ b/source/javascripts/components/JumpToDefinitionLink/JumpToDefinitionLink.tsx
@@ -35,8 +35,11 @@ const JumpToDefinitionLink = ({ kind, id, children, trigger, onOpenChange }: Pro
     (nodeId: string) => {
       jumpToDefinition(kind, id, nodeId);
       setIsOpen(false);
+      // Closing via the controlled prop doesn't fire Popover.onOpenChange, so notify callers
+      // tracking open state (e.g. a card's isJumpPopoverOpen) ourselves.
+      onOpenChange?.(false);
     },
-    [jumpToDefinition, kind, id],
+    [jumpToDefinition, kind, id, onOpenChange],
   );
 
   return (
--- a/source/javascripts/components/JumpToDefinitionLink/JumpToDefinitionLink.tsx
+++ b/source/javascripts/components/JumpToDefinitionLink/JumpToDefinitionLink.tsx
@@ -1,10 +1,10 @@
 import { Link } from '@bitrise/bitkit';
 import { Popover } from '@chakra-ui/react/popover';
 import { Portal } from '@chakra-ui/react/portal';
-import { ReactElement, ReactNode, useMemo, useState } from 'react';
+import { ReactElement, ReactNode, useCallback, useMemo, useState } from 'react';
 
 import FileTreeView from '@/components/FileTreeViewer/FileTreeView';
-import { EntityKind } from '@/core/models/Tree';
+import { EntityKind, TreeNode } from '@/core/models/Tree';
 import EntityIndexService from '@/core/services/EntityIndexService';
 import { useEntityIndex } from '@/hooks/useEntityIndex';
 import useJumpToDefinition from '@/hooks/useJumpToDefinition';
@@ -37,6 +37,17 @@ const JumpToDefinitionLink = ({ kind, id, children, trigger, onOpenChange }: Pro
     [entityIndex, kind, id],
   );
 
+  // Stable identities so FileTreeView's memoized collection isn't rebuilt every render
+  // (a rebuild also silently resets the tree's expansion state).
+  const filterToDefinitions = useCallback((node: TreeNode) => definitionIds.has(node.nodeId), [definitionIds]);
+  const handleSelect = useCallback(
+    (nodeId: string) => {
+      jumpToDefinition(kind, id, nodeId);
+      setIsOpen(false);
+    },
+    [jumpToDefinition, kind, id],
+  );
+
   return (
     <Popover.Root
       open={isOpen}
@@ -67,16 +78,7 @@ const JumpToDefinitionLink = ({ kind, id, children, trigger, onOpenChange }: Pro
             boxShadow="large"
           >
             <Popover.Body p="8" maxHeight="50vh" overflowY="auto">
-              {tree && (
-                <FileTreeView
-                  rootNode={tree}
-                  filter={(node) => definitionIds.has(node.nodeId)}
-                  onSelect={(nodeId) => {
-                    jumpToDefinition(kind, id, nodeId);
-                    setIsOpen(false);
-                  }}
-                />
-              )}
+              {tree && <FileTreeView rootNode={tree} filter={filterToDefinitions} onSelect={handleSelect} />}
             </Popover.Body>
           </Popover.Content>
         </Popover.Positioner>
--- a/source/javascripts/components/JumpToDefinitionLink/JumpToDefinitionLink.tsx
+++ b/source/javascripts/components/JumpToDefinitionLink/JumpToDefinitionLink.tsx
@@ -1,0 +1,88 @@
+import { Link } from '@bitrise/bitkit';
+import { Popover } from '@chakra-ui/react/popover';
+import { Portal } from '@chakra-ui/react/portal';
+import { ReactElement, ReactNode, useMemo, useState } from 'react';
+
+import FileTreeView from '@/components/FileTreeViewer/FileTreeView';
+import { EntityKind } from '@/core/models/Tree';
+import EntityIndexService from '@/core/services/EntityIndexService';
+import { useEntityIndex } from '@/hooks/useEntityIndex';
+import useJumpToDefinition from '@/hooks/useJumpToDefinition';
+import { useTree } from '@/hooks/useTree';
+
+type Props = {
+  kind: EntityKind;
+  id: string;
+  /** Default trigger content, rendered inside a purple text Link. */
+  children?: ReactNode;
+  /** Custom trigger element used as the popover trigger. Takes precedence over `children`. */
+  trigger?: ReactElement<{ onClick?: () => void }>;
+  /** Called when the multi-definition chooser popover opens or closes. */
+  onOpenChange?: (isOpen: boolean) => void;
+};
+
+/**
+ * The "Edit / Go to definition" link in a cross-file drawer header. Always opens a
+ * chooser popover — for a single definition the user picks the one file; for multiple
+ * definitions (override, or same path from different refs) they pick which layer to open.
+ */
+const JumpToDefinitionLink = ({ kind, id, children, trigger, onOpenChange }: Props) => {
+  const tree = useTree();
+  const entityIndex = useEntityIndex();
+  const jumpToDefinition = useJumpToDefinition();
+  const [isOpen, setIsOpen] = useState(false);
+
+  const definitionIds = useMemo(
+    () => new Set(EntityIndexService.definitionsOf(entityIndex, kind, id).map((def) => def.nodeId)),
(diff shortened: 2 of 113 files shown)


source/javascripts/components/unified-editor/StepSelectorDrawer/components/StepBundleCard.tsx
```diff
@@ -136,23 +149,29 @@ const StepBundleCard = (props: StepBundleCardProps) => {
...
+      // Ghost (read-only) tint: cross-file references, and every card in a read-only view.
+      ...(isCrossFile || isReadOnlyView ? { backgroundColor: 'background/secondary' } : {}),
+      ...(isHighlighted ? { outline: '2px solid', outlineColor: 'border/selected' } : {}),
+    };
+  }, [isCollapsable, isDragging, isHighlighted, isPlaceholder, isCrossFile, isReadOnlyView]);
 
   const buttonGroup = useMemo(() => {
-    if ((!workflowId && !stepBundleId) || isDragging || (!onDeleteStep && !onSelectStep)) {
+    if ((!workflowId && !stepBundleId) || isDragging || !onDeleteStep) {
       return null;
```
moczolaszlo, line 164: This guard tightened from `(!onDeleteStep && !onSelectStep)` to `!onDeleteStep` (and `onSelectStep` left the deps), which — unlike the rest of this PR — is **not** gated on the read-only flag, so it changes the flag-off path too: a `StepBundleCard` that has `onSelectStep` but no `onDeleteStep` now renders no `StepMenu` where it previously did. It matches the read-only intent (read-only strips actions down to `onSelectStep`), but it isn't in the PR's "⚠️ Touches active prod" list. Can you confirm no current selection-only usage (`onSelectStep` present, `onDeleteStep` absent) relied on the menu showing — or that the menu was effectively empty there anyway?

moczolaszlo (commented):
Reviewed the largest PR in the stack. Verified the three provider extractions are behavior-identical pure moves (and the green tsc/Analyze CI confirms the call-site imports are rewired), the read-only gating is uniform and dormant-safe (every gate reduces to the old behavior with the flag off — no inverted operators, no Rules-of-Hooks violations), and the WorkflowCardContext callback-stripping returns the full pick today. New components (FileTreeView, JumpToDefinition*, CrossFile*) are clean. A few notes below — one flag-off behavior change worth confirming, plus house-style/latent items.

source/javascripts/components/unified-editor/WorkflowCard/components/StepList.tsx
```diff
@@ -122,7 +122,7 @@ const StepList = ({ stepBundleId, steps, onAdd, onMove, workflowId }: Props) =>
         paddingY="16"
         paddingX="16"
         iconName="Steps"
-        title={onAdd && workflowId ? 'Empty Workflow' : 'Empty Step bundle'}
+        title={workflowId ? 'Empty Workflow' : 'Empty Step bundle'}
```
moczolaszlo, line 125: Disclosed in the description (empty-state title no longer depends on `onAdd`). Just confirming there's no flag-off caller that passes `workflowId` without `onAdd` and would now show "Empty Workflow" instead of "Empty Step bundle" — if `onAdd`-absent only happens in read-only (dormant) today, this is fine.

source/javascripts/components/JumpToDefinitionLink/JumpToDefinitionLink.tsx
```diff
@@ -0,0 +1,88 @@
...
+      open={isOpen}
+      onOpenChange={(details) => {
+        setIsOpen(details.open);
+        onOpenChange?.(details.open);
+      }}
+      positioning={{ placement: 'bottom-start' }}
+    >
+      <Popover.Trigger asChild>
+        {trigger ?? (
+          <Link as="button" colorScheme="purple">
```
moczolaszlo, line 56: Per the repo guideline (new work uses `@bitrise/bitkit-v2`; migrate v1 in touched files), these brand-new components default to bitkit **v1**: `Link` here (and `ControlButton` in CrossFileJumpButton.tsx, `Card`/`Icon`/`Box`/`Text` in CrossFileDefinitionCard.tsx, `Text` in CrossFileProvenanceText.tsx). Worth using the v2 equivalents (`BitkitLink`, `BitkitControlButton`, …) since they're new files. Not a correctness issue.

source/javascripts/components/FileTreeViewer/FileTreeView.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const children = useMemo(() => toRenderNode(group.root), [group.root]);
+
+  const collection = useMemo(
+    () =>
+      createTreeCollection<RenderNode>({
+        rootNode: { id: GROUP_ROOT_ID, label: '', isFile: false, children },
+        nodeToValue: (node) => node.id,
+        nodeToString: (node) => node.label,
+        nodeToChildren: (node) => node.children,
+        isNodeDisabled: (node) => Boolean(node.isFile && node.node && isNodeDisabled?.(node.node)),
```
moczolaszlo, line 77: Latent (dormant) perf/state note: `createTreeCollection` is memoized on `[children, isNodeDisabled]` and the `groups` build on `[rootNode, projectRepoLabel, filter]`, but callers pass `filter` / `isNodeDisabled` as inline closures (e.g. JumpToDefinitionLink's `(node) => definitionIds.has(node.nodeId)`), so the identity changes every render and the memo is defeated — and a rebuilt collection silently resets TreeView expansion state. Either document "memoize these props" on the public API or wrap them in `useCallback` at the call sites.

source/javascripts/components/unified-editor/StepSelectorDrawer/components/StepBundleCard.tsx
trapacska: Good catch — it was a real flag-off regression, not just selection-only. Fixed in 4b416794: the menu's actions come from `onDeleteStep` **or** `onDeleteStepInStepBundle` (the latter is what step-bundles-nested-in-step-bundles use on the StepBundles page), so narrowing to `!onDeleteStep` hid the menu there. Gate is now `(!onDeleteStep && !onDeleteStepInStepBundle)`. On the selection-only case you asked about: `onSelectStep`-present/`onDeleteStep`-absent is the drawer, where the menu had no actionable callbacks anyway (effectively empty), so hiding it there is correct.

source/javascripts/components/unified-editor/WorkflowCard/components/StepList.tsx
trapacska: Confirmed safe — and it's actually a fix. Old title was `onAdd && workflowId ? 'Empty Workflow' : 'Empty Step bundle'`, so any workflow list rendered without `onAdd` was mislabeled 'Empty Step bundle'. The new `workflowId ? …` labels it correctly regardless of `onAdd`. `onAdd`-absent on a workflow is the read-only (dormant) case today; if any flag-off path hits it, the label is now correct rather than wrong.

source/javascripts/components/JumpToDefinitionLink/JumpToDefinitionLink.tsx
trapacska: Agreed these new files should be v2. Deferring the migration to a dedicated follow-up rather than expanding this (already large) PR's diff — the components are dormant until PR 7 wires them up, so there's no user impact in the meantime. Tracking it as a follow-up.

source/javascripts/components/FileTreeViewer/FileTreeView.tsx
trapacska: Fixed at the call site in 4b416794 — `JumpToDefinitionLink` now wraps the `filter` and `onSelect` closures in `useCallback`, so FileTreeView's memoized collection (and the TreeView expansion state) is stable across renders. It's the only caller today; will keep the 'memoize these props' expectation in mind as more call sites land in PR 7.

source/javascripts/components/unified-editor/WorkflowConfig/tabs/PropertiesTab.tsx
```diff
@@ -34,32 +35,33 @@ const NameInput = ({ onRename }: Pick<Props, 'onRename'>) => {
       name="name"
       label="Name"
       value={value}
+      isDisabled={isDisabled}
       sanitize={WorkflowService.sanitizeName}
```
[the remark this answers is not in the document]
trapacska: Fixed in 4b416794, in EditableInput itself so all three call sites (Workflow/Pipeline/StepBundle PropertiesTab) are covered: `isDisabled` now gates the Edit pencil button, the Commit handler, and the Enter-key path — not just the text field. Dormant-safe: EditableInput's only callers are these read-only-gated NameInputs, so with the flag off (isReadOnlyView=false) behavior is unchanged.

source/javascripts/pages/TriggersPage/components/LegacyTriggers/TriggerCard.tsx
```diff
@@ -6,6 +6,7 @@ import TriggerConditions from '@/components/unified-editor/Triggers/TriggerCondi
 import { trackTriggerEnabledToggled } from '@/core/analytics/TriggerAnalytics';
 import { TriggerSource } from '@/core/models/Trigger';
 import { LegacyTrigger } from '@/core/models/Trigger.legacy';
+import { useIsReadOnlyView } from '@/hooks/useTree';
```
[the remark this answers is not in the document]
trapacska: Fixed in 4b416794 — `useSortable` now gets `disabled: isReadOnlyView`, so the drag handle can't reorder/mutate in a read-only view. No-op when the flag is off.

source/javascripts/components/unified-editor/StepBundleConfig/StepBundleInputs/StepBundleInputsCategoryCard.tsx
```diff
@@ -86,7 +93,13 @@ const StepBundleInputsCategoryCard = ({ category, items = [], onAdd, onChange, o
           </Fragment>
         );
       })}
-      <Button leftIconName="Plus" variant="tertiary" size="md" onClick={() => onAdd(category)}>
+      <Button
+        leftIconName="Plus"
+        variant="tertiary"
+        size="md"
+        isDisabled={isReadOnlyView}
+        onClick={() => onAdd(category)}
```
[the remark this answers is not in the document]
trapacska: Fixed in 4b416794 — both `StepInput` and `StepSelectInput` now include `isReadOnlyView` in their `isDisabled` condition (`opts?.is_dont_change_value || isReadOnlyView`).

source/javascripts/components/unified-editor/WorkflowCard/components/ChainedWorkflowCard.tsx
```diff
@@ -85,17 +96,22 @@ const ChainedWorkflowCard = ({ id, index, uniqueId, placement, isSortable, isDra
...
+    if (
+      isDragging ||
+      (!onEditChainedWorkflow && !onChainChainedWorkflow && !onRemoveChainedWorkflow && !showJumpButton)
+    ) {
       return null;
     }
 
     return (
-      <ButtonGroup spacing="0" display="none" _groupHover={{ display: 'flex' }}>
+      <ButtonGroup spacing="0" display={isJumpPopoverOpen ? 'flex' : 'none'} _groupHover={{ display: 'flex' }}>
```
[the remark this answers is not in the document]
trapacska: Fixed in 4b416794 — the Chain Workflows button is now guarded by `!isCrossFile` (with `isCrossFile` added to the memo deps), matching WorkflowCard, so you can't chain into a workflow defined in another file.

source/javascripts/components/unified-editor/StepSelectorDrawer/components/StepBundleList.tsx
```diff
@@ -22,10 +24,14 @@ const StepBundleList = ({ onSelectStep, excludedStepBundleId }: StepBundleListPr
...
+  // Chains only exist for local bundles, so guard the lookup; a cross-file bundle has no chain and can't be excluded.
   const stepBundleChains = StepBundleService.getStepBundleChains(stepBundles);
-  const bundleIds = Object.keys(stepBundles).filter((id) => {
+  const bundleIds = allBundleIds.filter((id) => {
     if (excludedStepBundleId) {
-      return !stepBundleChains[id].includes(excludedStepBundleId);
+      return !stepBundleChains[id]?.includes(excludedStepBundleId);
     }
     return true;
   });
```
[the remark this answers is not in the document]
trapacska: Fixed in 4b416794 — added `id !== excludedStepBundleId` to the filter, so the bundle being edited is excluded from its own add-list (a chain never lists itself, and a cross-file bundle has no chain, so the old condition offered it). Also hardens the local/prod case against direct self-reference.

source/javascripts/components/unified-editor/StepSelectorDrawer/components/StepBundleCard.tsx
moczolaszlo: Confirmed fixed in 4b416794 — and it was a real flag-off regression: the menu also surfaces `onDeleteStepInStepBundle` (nested step-bundles on the StepBundles page), so the gate is now `(!onDeleteStep && !onDeleteStepInStepBundle)`. Verified. Resolving.

source/javascripts/components/unified-editor/WorkflowCard/components/StepList.tsx
moczolaszlo: Confirmed — it's a fix, not a regression (the old `onAdd && workflowId` mislabeled an onAdd-less workflow list). Resolving.

source/javascripts/components/JumpToDefinitionLink/JumpToDefinitionLink.tsx
moczolaszlo: Agreed; deferring the v1→v2 migration of these dormant new components to a dedicated follow-up is reasonable. Resolving.

source/javascripts/components/FileTreeViewer/FileTreeView.tsx
moczolaszlo: Confirmed fixed in 4b416794 — `JumpToDefinitionLink` now memoizes the `filter`/`onSelect` closures, so the collection + expansion state are stable. Resolving.

source/javascripts/pages/StepBundlesPage/StepBundlesPage.tsx
```diff
@@ -2,14 +2,33 @@ import { Box, Button, EmptyState } from '@bitrise/bitkit';
...
+  if (openedDialogType === StepBundlesPageDialogType.NONE) {
+    return null;
+  }
+
+  return {
+    stepBundleIds: [stepBundleId],
+    steps:
+      openedDialogType === StepBundlesPageDialogType.STEP_CONFIG && stepBundleId
+        ? { source: 'step_bundles', sourceId: stepBundleId, indices: selectedStepIndices }
+        : undefined,
```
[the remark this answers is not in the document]
trapacska: Good catch — confirmed the STEP_BUNDLE drawer renders at `selectedStepIndices[0]` (Drawers.tsx), so it needs the same step-index validation as STEP_CONFIG. Fixed in cda5146c: the file-switch refs now include STEP_BUNDLE in the step-index check, so a switch drops the drawer if that index no longer resolves — matching WorkflowsPage's STEP_INDEX_DIALOGS. Dormant (modular file-switch only) until PR 7.

source/javascripts/components/unified-editor/WorkflowConfig/tabs/PropertiesTab.tsx
```diff
@@ -68,23 +70,32 @@ const Priority = ({ workflowId }: { workflowId: string }) => {
...
+      targetId={workflowId}
+      statusReportName={value}
+      isDisabled={isDisabled}
+      onChange={handleOnChange}
+    />
+  );
 };
 
 const PropertiesTab = ({ variant, onRename, onDelete }: Props) => {
   const workflowId = useWorkflowConfigContext((s) => s?.id || '');
```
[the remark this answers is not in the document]
trapacska: Fixed in cda5146c — switched to `useWorkflowConfigId()`, which reads the id straight from context and resolves even for a cross-file workflow (its doc comment notes exactly this), so mutations can't target an empty id if the tab is rendered via a stale `?tab=properties` URL. Left the other `useWorkflowConfigContext(s => s?.…)` reads as-is since they're display values on a tab that's disabled for cross-file. Dormant until PR 7.

source/javascripts/components/JumpToDefinitionLink/JumpToDefinitionLink.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { EntityKind, TreeNode } from '@/core/models/Tree';
+import EntityIndexService from '@/core/services/EntityIndexService';
+import { useEntityIndex } from '@/hooks/useEntityIndex';
+import useJumpToDefinition from '@/hooks/useJumpToDefinition';
+import { useTree } from '@/hooks/useTree';
+
+type Props = {
+  kind: EntityKind;
+  id: string;
+  /** Default trigger content, rendered inside a purple text Link. */
```
k-olly: Please remove unnecessary comments from props - it makes it hard to read

source/javascripts/components/unified-editor/StepConfigDrawer/StepConfigDrawer.context.tsx
```diff
@@ -1,12 +1,10 @@
...
+import { createContext, useContext } from 'react';
 
 import { Step } from '@/core/models/Step';
-import useStep from '@/hooks/useStep';
 
-import VersionChangedDialog from '../VersionChangedDialog/VersionChangedDialog';
-
-type Props = { workflowId: string; stepIndex: number; stepBundleId?: string };
-type State = {
+// Provider component lives in StepConfigDrawerProvider.tsx so this module exports only non-components, keeping React Fast Refresh happy.
```
k-olly: Useless comment, remove it please

source/javascripts/components/unified-editor/StepBundleConfig/StepBundleConfig.context.tsx
```diff
@@ -1,30 +1,21 @@
-import { createContext, PropsWithChildren, useContext } from 'react';
+import { createContext, useContext } from 'react';
 
 import { StepBundleInstance } from '@/core/models/Step';
 import useStepBundle from '@/hooks/useStepBundle';
 
-import StepConfigDrawerProvider from '../StepConfigDrawer/StepConfigDrawer.context';
-
-type Props = PropsWithChildren<{
+// Provider component lives in StepBundleConfigProvider.tsx so this module exports only non-components, keeping React Fast Refresh happy.
```
k-olly: Useless comment, remove it

source/javascripts/pages/StepBundlesPage/StepBundlesPage.tsx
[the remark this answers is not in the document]
moczolaszlo: Endorsing this — it's dormant (file-switch only happens in modular mode, flag off today), but it's a real gap: the nested `STEP_BUNDLE` dialog uses `selectedStepIndices[0]` as a step index yet isn't represented in `fileSwitchDialogRefs`, so a file switch wouldn't re-validate/close it. Worth closing before PR 7 activates this path. Holding my approval until this and the cross-file PropertiesTab thread are addressed (or explicitly deferred with a tracked follow-up).

source/javascripts/components/unified-editor/WorkflowConfig/tabs/PropertiesTab.tsx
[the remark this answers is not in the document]
moczolaszlo: Endorsing this too — dormant (cross-file workflows only exist in modular mode), but worth confirming the empty `workflowId` ('') can't reach an editable path once the header's `isCrossFile` disable is bypassed (e.g. via a deep link straight to the tab). Holding approval on this + the StepBundlesPage thread; happy to approve once both are resolved or deferred with a tracked follow-up.

source/javascripts/pages/StepBundlesPage/StepBundlesPage.tsx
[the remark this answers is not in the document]
moczolaszlo: Fixed in cda5146c02 — `fileSwitchDialogRefs` now includes the nested `STEP_BUNDLE` dialog type in the step-index validation, so a file switch drops it when the index no longer resolves. Verified. Resolving. (Note: the analogous path on WorkflowsPage is flagged separately and still open.)

source/javascripts/components/unified-editor/WorkflowConfig/tabs/PropertiesTab.tsx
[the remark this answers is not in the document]
moczolaszlo: Fixed in cda5146c02 — `PropertiesTab` now reads the id via `useWorkflowConfigId()`, which resolves even for a cross-file workflow, so mutations can't target an empty id. Verified. Resolving.

source/javascripts/components/JumpToDefinitionLink/JumpToDefinitionLink.tsx
trapacska: Done in d0f1396e — removed the prop doc comments here (and the component doc block). Did a pass over the whole PR's comments too: dropped prop/component doc blocks, the 'split for React Fast Refresh' notes, and descriptive narration, keeping only the few that document a non-obvious gotcha.

source/javascripts/components/unified-editor/StepConfigDrawer/StepConfigDrawer.context.tsx
trapacska: Removed in d0f1396e.

source/javascripts/components/unified-editor/StepBundleConfig/StepBundleConfig.context.tsx
trapacska: Removed in d0f1396e.

source/javascripts/components/unified-editor/WorkflowConfig/tabs/PropertiesTab.tsx
```diff
@@ -4,20 +4,21 @@ import { ChangeEventHandler } from 'react';
...
 type Props = {
   variant: 'panel' | 'drawer';
   onRename: (name: string) => void;
   onDelete?: (id: string) => void;
 };
 
-const NameInput = ({ onRename }: Pick<Props, 'onRename'>) => {
+const NameInput = ({ onRename, isDisabled }: Pick<Props, 'onRename'> & { isDisabled?: boolean }) => {
   const rename = useRenameWorkflow(onRename);
   const value = useWorkflowConfigContext((s) => s?.id || '');
```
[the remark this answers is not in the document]
trapacska: Fixed in 2b8d360b — NameInput's `value` (and `useRenameWorkflow`'s `selectedWorkflowId`, same selector) now use `useWorkflowConfigId()`, which resolves cross-file refs. Identical for local workflows; removes the empty-id rename target for cross-file, consistent with the PropertiesTab fix.

source/javascripts/pages/WorkflowsPage/WorkflowsPage.tsx
```diff
@@ -3,12 +3,40 @@ import { useEffect } from 'react';
...
+  return {
+    workflowIds: [workflowId, parentWorkflowId],
+    stepBundleIds: [stepBundleId, parentStepBundleId],
+    steps: STEP_INDEX_DIALOGS.includes(openedDialogType)
+      ? {
+          source: stepBundleId ? 'step_bundles' : 'workflows',
+          sourceId: stepBundleId || workflowId,
+          indices: selectedStepIndices,
+        }
+      : undefined,
```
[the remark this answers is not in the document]
trapacska: I traced the field semantics on this one and I believe it's a false positive — `stepBundleId` here is the **container** of `selectedStepIndices`, not the opened bundle's own id. The select handlers (`StepBundleCard`/`StepCard` → `onSelectStep({ stepIndex, stepBundleId, wfId })`) pass each card's *container* props: a step-bundle card sitting directly in a workflow is invoked with `stepBundleId` empty (→ `workflows/workflowId`), and one nested inside another bundle with `stepBundleId` = the parent bundle (→ `step_bundles/stepBundleId`). So `sourceId: stepBundleId || workflowId` already validates against the correct container, and it mirrors the selection model's own precedence (`selectionParent.id = stepBundleId || wfId`). The suggested `stepBundleId && !workflowId` guard would instead break the nested-step-bundle case (where both are set), so I've left this as-is. Happy to revisit if I'm misreading the open path.

k-olly (dismissed):

source/javascripts/pages/TriggersPage/components/LegacyTriggers/LegacyTriggers.tsx
```diff
@@ -6,11 +6,13 @@ import { TriggerType } from '@/core/models/Trigger';
 import { LegacyTrigger } from '@/core/models/Trigger.legacy';
 import TriggerService from '@/core/services/TriggerService';
 import useLegacyTriggers from '@/hooks/useLegacyTriggers';
+import { useIsReadOnlyView } from '@/hooks/useTree';
 import SortableTriggerList from '@/pages/TriggersPage/components/LegacyTriggers/SortableTriggerList';
 
 import ConvertLegacyTriggers from './ConvertLegacyTriggers';
 
 const LegacyTriggers = () => {
+  const isReadOnlyView = useIsReadOnlyView();
```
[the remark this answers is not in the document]
trapacska: Valid — fixed in e3fa2037. `ConvertLegacyTriggers` is now wrapped in `{!isReadOnlyView && …}`, so the YAML-mutating 'Convert triggers' action is hidden in read-only views, consistent with the gated Add buttons.

source/javascripts/components/JumpToDefinitionLink/JumpToDefinitionLink.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  // Stable identities: a new closure rebuilds FileTreeView's collection and resets its expansion state.
+  const filterToDefinitions = useCallback((node: TreeNode) => definitionIds.has(node.nodeId), [definitionIds]);
+  const handleSelect = useCallback(
+    (nodeId: string) => {
+      jumpToDefinition(kind, id, nodeId);
+      setIsOpen(false);
+    },
+    [jumpToDefinition, kind, id],
+  );
```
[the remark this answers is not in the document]
trapacska: Valid — fixed in e3fa2037. `handleSelect` now also calls `onOpenChange?.(false)` (added to deps): closing via the controlled `open` prop doesn't fire `Popover.onOpenChange`, so callers tracking open state (WorkflowCard/ChainedWorkflowCard's `isJumpPopoverOpen`) would otherwise stay stuck true after picking a definition.

k-olly (approved):

PR #1791, opened 2026-06-11
BIVS-3677 [6/7] YML page: file tabs, module tree, multi-file diffs

trapacska:
Part **6/7** of the modular YAML editor stack. Stacked on the read-only UI PR.

## What
- `OpenFileTabs` strip: merged-config tab + per-file tabs, file create/include (`CreateFileButton`), per-tab diff (`TabDiffButton`)
- `ModularYmlEditor`: editable module files; read-only merged-config and cross-repo views
- `FileTreeViewer` popover for opening modules
- `GlobalDiffEditorDialog` (all dirty files with a tree selector) and `MergedDiffDialog`
- `Header`: per-tab visual-page memory on YAML↔visual switches; swaps in the global diff dialog when modular

## Safety
**Dormant**: `YmlPage` renders `ModularYmlEditor` and `Header` swaps dialogs only when `store.tree` is set. The tab-location memory calls are no-ops without a selected node.

## ⚠️ Touches active prod code — review closely, verify with flag off
- `DiffEditorDialog` restructured into `Shell` + `Content`; the legacy dialog now composes these pieces and must render identically
- `DiffEditor` gains an optional `readOnly` prop (default unchanged)

Manual QA with flag off: open the diff dialog from the header with unsaved changes, apply changes from it, switch YAML↔visual.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1792, opened 2026-06-11
BIVS-3677 [7/7] Activation: flag-gated bootstrap, layout + dev toggle

trapacska:
Part **7/7** — the switch for the whole stack. After this merges, modular YAML editing is fully controlled by the `enable-wfe-modular-yaml-editing` LaunchDarkly flag (default **off**).

## What
- `main.tsx`: flag-gated bootstrap — modular tree query + `initializeModularConfig` vs the legacy single-file path; ignores Monaco's benign 'Canceled' rejection on model disposal
- `MainLayout`: modular grid with the `OpenFileTabs` strip (shares the 48px top row with the settings bar in website mode, tops the content area otherwise)
- `ConfigSettingsBar`: fixed 48px height **only in modular mode** (aligns with the tab strip)
- `ModularYamlDevToggle`: dev-only floating flag override (not rendered in production)

## 🚧 Backend dependency — do NOT enable the flag before the monolith release
This PR is safe to **merge** anytime, but turning the flag on for any org requires the backend endpoints to be live first. With the flag on, the editor bootstraps from the tree endpoint and **fails to load entirely** if it 404s (no fallback to the legacy path). Required backend support:
- `GET /api/app/:slug/config/tree` — bootstrap fetch
- `POST /api/app/:slug/config/merge` — merged-config re-sync
- `POST /api/app/:slug/config/push` — modular save (incl. 409 conflict payload)
- modular `root` payload variant of the push-to-branch endpoint

Also note: the Go `apiserver/` (CLI plugin mode) has no tree endpoints — modular editing is **website-mode only** for now (CLI has no LaunchDarkly client, so it stays legacy by default). Devs using `ModularYamlDevToggle` locally need a monolith with the endpoints above, otherwise the editor won't load.

## Safety
Flag off → `useGetCiConfig` + `initializeBitriseYmlDocument` run exactly as today, `store.tree` stays unset, and **the layout/bar sizing are unchanged** — the non-modular grid (full-width bar on the YAML page) and the 65px bar are preserved verbatim. Note: the tab-strip UI is still in design concept phase, which is why all sizing changes are runtime-gated rather than unconditional.

Manual QA: flag off → website + CLI mode layouts identical to production; flag on (dev toggle) → tab strip, modular editor, merged view.

### DIFF
--- a/source/javascripts/components/CrossFileProvenanceText.tsx
+++ b/source/javascripts/components/CrossFileProvenanceText.tsx
@@ -1,0 +1,27 @@
+import { Text } from '@bitrise/bitkit';
+
+type Props = {
+  definingPath?: string;
+  sourceLabel?: string;
+  /** textStyle of the emphasized path segment. */
+  pathTextStyle?: string;
+};
+
+/** Plain-string variant for truncating card subtitles. */
+export const crossFileProvenanceLabel = (definingPath?: string) => `Defined in ${definingPath || 'another file'}`;
+
+/**
+ * Inline "Defined in `<file>` • `<source>`" provenance for a cross-file entity. Callers
+ * gate on `isCrossFile` — it renders nothing meaningful for a local entity.
+ */
+const CrossFileProvenanceText = ({ definingPath, sourceLabel, pathTextStyle = 'body/sm/semibold' }: Props) => (
+  <>
+    Defined in{' '}
+    <Text as="span" textStyle={pathTextStyle}>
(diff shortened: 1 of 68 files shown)


trapacska:
⚠️ This PR was **not actually merged** — same rebase accident as #1791 (head momentarily identical to base → GitHub auto-marked merged + auto-deleted the branch). Nothing landed. Replacement PR with identical content: #1794

PR #1794, opened 2026-06-12
BIVS-3677 [7/7] Activation: flag-gated bootstrap, layout + dev toggle

trapacska:
Part **7/7** — the switch for the whole stack. After this merges, modular YAML editing is fully controlled by the `enable-wfe-modular-yaml-editing` LaunchDarkly flag (default **off**).

## What
- `main.tsx`: flag-gated bootstrap — modular tree query + `initializeModularConfig` vs the legacy single-file path; ignores Monaco's benign 'Canceled' rejection on model disposal
- `MainLayout`: modular grid with the `OpenFileTabs` strip (shares the 48px top row with the settings bar in website mode, tops the content area otherwise)
- `ConfigSettingsBar`: fixed 48px height **only in modular mode** (aligns with the tab strip)
- `ModularYamlDevToggle`: dev-only floating flag override (not rendered in production)

## 🚧 Backend dependency — do NOT enable the flag before the monolith release
This PR is safe to **merge** anytime, but turning the flag on for any org requires the backend endpoints to be live first. With the flag on, the editor bootstraps from the tree endpoint and **fails to load entirely** if it 404s (no fallback to the legacy path). Required backend support:
- `GET /api/app/:slug/config/tree` — bootstrap fetch
- `POST /api/app/:slug/config/merge` — merged-config re-sync
- `POST /api/app/:slug/config/push` — modular save (incl. 409 conflict payload)
- modular `root` payload variant of the push-to-branch endpoint

Also note: the Go `apiserver/` (CLI plugin mode) has no tree endpoints — modular editing is **website-mode only** for now (CLI has no LaunchDarkly client, so it stays legacy by default). Devs using `ModularYamlDevToggle` locally need a monolith with the endpoints above, otherwise the editor won't load.

## Safety
Flag off → `useGetCiConfig` + `initializeBitriseYmlDocument` run exactly as today, `store.tree` stays unset, and **the layout/bar sizing are unchanged** — the non-modular grid (full-width bar on the YAML page) and the 65px bar are preserved verbatim. Note: the tab-strip UI is still in design concept phase, which is why all sizing changes are runtime-gated rather than unconditional.

Manual QA: flag off → website + CLI mode layouts identical to production; flag on (dev toggle) → tab strip, modular editor, merged view.

> **Note:** replaces #1792 — that PR was auto-marked "merged" by GitHub after a botched stack rebase momentarily made this branch identical to its base (nothing actually merged; branch restored, content unchanged apart from the useShallow review cleanup).


source/javascripts/components/DiffEditor/GlobalDiffEditorDialog.tsx
k-olly: <Box width={rem(280)}

rem imported from bitrise/bitkit-v2

k-olly: Remove comment

source/javascripts/components/DiffEditor/MergedDiffDialog.tsx
k-olly: onChange={() => {}} has no functionality here

source/javascripts/components/unified-editor/ConfigSettingsBar/ConfigSettingsBar.tsx
```diff
@@ -63,8 +64,10 @@ const ConfigSettingsBar = (props: BoxProps) => {
     <Box
       paddingLeft="32"
       paddingRight="12"
-      paddingBlock="12"
-      minHeight={rem(65)}
+      // Modular: fixed 48px border-box so it lines up with the OpenFileTabs strip; the
```
k-olly: Remove comment

source/javascripts/layouts/MainLayout.tsx
```diff
@@ -31,42 +34,65 @@ const MainLayout = () => {
...
-            <Switch>
-              {routes.map(({ path, component }) => (
-                <LazyRoute key={path} path={new RegExp(`^\\${path}`)} component={component} />
-              ))}
-              <Redirect to={paths.workflows} replace />
-            </Switch>
-          </Router>
+        <Box gridArea="content" display="flex" flexDirection="column" minW={0} minH={0}>
+          {isModular && !tabsBesideBar && <OpenFileTabs />}
+          <Box flex="1" overflowX="hidden" overflowY="auto">
```
moczolaszlo, line 88: This content-area restructure runs for **all** users, not just modular: the `content` Box gained an extra flex-column wrapper and the scroll moved into a nested `flex="1"` box. The PR says the non-modular layout is preserved verbatim, but the content subtree shape did change. It's very likely layout-equivalent (correct nested-scroll pattern with `minH={0}`), but since it's the one unconditional DOM change here, worth QA-confirming scroll/height parity with the flag **off** on the YAML page and the entity pages (sticky headers, full-height Monaco editor, inner scrolling).

source/javascripts/main.tsx
```diff
@@ -116,25 +130,44 @@ const InitialDataLoader = ({ children }: PropsWithChildren) => {
   });
 
   useEventListener('unhandledrejection', (e) => {
+    // Monaco rejects with a benign "Canceled" sentinel when a model is disposed (e.g. tab switch); not a real error.
+    const reason = e.reason as { name?: string; message?: string } | undefined;
+    if (reason?.name === 'Canceled' || reason?.message === 'Canceled') {
+      return;
```
moczolaszlo, line 136: This `'Canceled'` filter is **always-on** (not flag-gated): previously every unhandled rejection went to datadog + an error toast, now any rejection with `name`/`message === 'Canceled'` is silently dropped for all users. Almost certainly a beneficial fix — Monaco emits a benign `Canceled` on model disposal (tab switch), and the YAML editor runs in legacy mode too, so this removes spurious prod toasts today. Just confirming it's intended as an always-on change, and that matching on `message === 'Canceled'` (not only `name`) isn't broad enough to swallow a genuinely-failed promise that happens to carry that message.

source/javascripts/components/unified-editor/ConfigSettingsBar/ConfigSettingsBar.tsx
```diff
@@ -63,8 +64,10 @@ const ConfigSettingsBar = (props: BoxProps) => {
...
+      // two-row branch-switching variant fits (20px + 4px gap + 16px = 40px of content).
+      // Non-modular keeps the original sizing untouched.
+      {...(isModular ? { paddingBlock: '0', height: '48px' } : { paddingBlock: '12', minHeight: rem(65) })}
```
moczolaszlo: Heads-up (coordination, not a code issue): #1795 also edits this file and is still open — it changes the signature to `({ showValidationBadge, ...props })` and adds the validation badge, while this PR adds the `isModular` sizing on the old `(props: BoxProps)` signature. Whichever of #1794/#1795 merges second will hit a conflict here — worth deciding the merge order up front.

source/javascripts/main.tsx
```diff
@@ -11,13 +11,16 @@ import { ComponentProps, PropsWithChildren, StrictMode, useEffect, useRef } from
 import { createRoot } from 'react-dom/client';
 import { useEventListener } from 'usehooks-ts';
 
+import ModularYamlDevToggle from '@/components/ModularYamlDevToggle';
```
moczolaszlo, line 14: Minor: `ModularYamlDevToggle` is statically imported and rendered in `App`, so even though it returns `null` in production, its dev-only dependencies (`CreateFileButton`, `TabDiffButton` from OpenFileTabs) get pulled into the production bundle. Consider lazy-loading the toggle (or gating the import) so the dev-only tree stays out of the prod build.

moczolaszlo (commented):
Reviewed the activation switch. The flag-off safety is thoroughly handled and I traced it end-to-end: the main.tsx bootstrap reduces to the legacy `useGetCiConfig` + `initializeBitriseYmlDocument` path byte-for-byte (tree query disabled via `enabled`), every MainLayout grid/bar expression is gated `&& !isModular` so it equals the old layout verbatim, ConfigSettingsBar keeps its 65px sizing, the dev toggle is null + inert in production, and `clearedModularState()` in the legacy init (with a spec) closes the modular→single-file teardown gap I raised back on PR #1788. CI is green. Four notes below — two are flag-off-affecting things to confirm, one is merge coordination, one a minor nit. None blocking.

source/javascripts/components/unified-editor/ConfigSettingsBar/ConfigSettingsBar.tsx
k-olly: I don't like this ugly logic, sorry.
I was thinking of :       
paddingBlock={isModular ? '0' : '12'}
height={isModular ? '48' : undefined}
minHeight={isModular ? undefined : rem(65)}

source/javascripts/core/stores/BitriseYmlStore.ts
```diff
@@ -232,6 +247,11 @@ export function initializeBitriseYmlDocument({
...
     configBranch: branch || undefined,
     configCommitSha: commitSha || undefined,
+    // The legacy (single-file) init owns the whole store, so clear any modular
+    // tree state. A no-op in pure legacy mode (fields are already empty); guards
+    // against a stale tree/files/tabs leaking in if a modular session ever routes
+    // through a legacy path (branch switch, repository-YAML save, manual update).
+    ...clearedModularState(),
     ...(doc.errors.length === 0
       ? { ymlDocument: doc, savedYmlDocument: doc, __invalidYmlString: undefined, __savedInvalidYmlString: undefined }
       : { __invalidYmlString: ymlString, __savedInvalidYmlString: ymlString }),
```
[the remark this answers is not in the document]
trapacska: Fixed in d0822ce0 — the invalid-YAML branch now resets `ymlDocument`/`savedYmlDocument` to fresh `new Document()` (confirmed `clearedModularState()` doesn't cover those two fields), so a stale modular file's Document can't survive a modular→legacy transition.

source/javascripts/components/ModularYamlDevToggle.tsx
```diff
@@ -0,0 +1,82 @@
...
+      bottom="16"
+      left="50%"
+      transform="translateX(-50%)"
+      zIndex={9999}
+      display="flex"
+      alignItems="center"
+      gap="8"
+    >
+      <TabDiffButton />
+      <CreateFileButton />
```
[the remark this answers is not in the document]
trapacska: Fixed in d0822ce0 — `TabDiffButton`/`CreateFileButton` now render only when `enabled` (the flag), so they never mount against empty tree state with the flag off. (The toggle already returns null entirely in production.)

source/javascripts/components/DiffEditor/DiffEditorDialog.tsx
```diff
@@ -78,18 +129,41 @@ const DiffEditorDialogBody = ({ onClose }: { onClose: VoidFunction }) => {
...
+              onChange={(value) => setCurrentText(typeof value === 'string' ? value : undefined)}
+              onMount={handlePlainEditorMount}
+              options={{ minimap: { enabled: false }, readOnly: isReadOnly }}
+              wrapperProps={{ style: { flex: 1, display: 'flex', minHeight: 0 } }}
+            />
+          ) : (
+            originalText &&
+            modifiedText && (
+              <DiffEditor
+                key={nodeId}
```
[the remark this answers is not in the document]
trapacska: Fixed in d0822ce0 — `key={nodeId ?? 'legacy'}` on both editors so the key is never undefined.

source/javascripts/components/DiffEditor/GlobalDiffEditorDialog.tsx
trapacska: Fixed in d0822ce0 — `width={rem(280)}` (`rem` imported from `@bitrise/bitkit-v2`).

trapacska: Fixed in d0822ce0 — comment removed.

source/javascripts/components/DiffEditor/MergedDiffDialog.tsx
trapacska: Fixed in d0822ce0 — made `DiffEditor`'s `onChange` optional and dropped the no-op here.

source/javascripts/components/unified-editor/ConfigSettingsBar/ConfigSettingsBar.tsx
trapacska: Fixed in d0822ce0 — comment removed.

source/javascripts/components/unified-editor/ConfigSettingsBar/ConfigSettingsBar.tsx
trapacska: Done in d0822ce0 — switched to your version: `paddingBlock={isModular ? '0' : '12'}` / `height={isModular ? '48' : undefined}` / `minHeight={isModular ? undefined : rem(65)}`. Much cleaner, thanks.

source/javascripts/main.tsx
```diff
@@ -144,14 +177,14 @@ const InitialDataLoader = ({ children }: PropsWithChildren) => {
         isLoaded.current = true;
       }
     }
-  }, [data, requestedBranch, toast]);
+  }, [data, requestedBranch, toast, isModularEnabled, legacyConfig.data, treeConfig.data, configBranch]);
```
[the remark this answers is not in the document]
trapacska: Leaving the deps as-is. The effect body reads `treeConfig.data` and `legacyConfig.data` directly (lines ~145/155), so `react-hooks/exhaustive-deps` requires both in the array — dropping them would be a lint violation. The re-fire concern is already neutralized by the `loadedBranch.current !== requestedBranch` guard: redundant runs (incl. a disabled query's cache changing) hit that guard and no-op, so the store isn't re-initialized mid-session.

source/javascripts/main.tsx
trapacska: Confirmed intentional and always-on. Monaco rejects with a benign `Canceled` on model disposal (tab switch) and the YAML editor runs in legacy mode too, so this kills spurious prod toasts today. The `message === 'Canceled'` arm is a deliberate belt-and-suspenders — Monaco's CancellationError doesn't always carry `name === 'Canceled'`. The false-positive risk (a genuine failure whose message is exactly `Canceled`) is negligible in practice; happy to tighten to `name`-only if we ever see one slip through.

source/javascripts/components/unified-editor/ConfigSettingsBar/ConfigSettingsBar.tsx
trapacska: This conflict has since materialized and is resolved — master (with the `({ showValidationBadge, ...props })` signature + badge) is merged into this branch, and I combined it with the `isModular` sizing. The sizing block is now plain conditional props (per k-olly's suggestion above), and `MainLayout` passes `showValidationBadge={isYmlPage}`.

source/javascripts/layouts/MainLayout.tsx
trapacska: Good callout — yes, the content-area wrapper change is unconditional. It's the standard nested-scroll pattern (outer `flex` column + inner `flex="1"` scroll box, both `minH={0}`), intended to be layout-equivalent with the flag off. I'll QA scroll/height parity with the flag **off** on the YAML page and entity pages (sticky headers + full-height Monaco + inner scrolling) before merge.

source/javascripts/main.tsx
trapacska: Fair point. Note the heaviest dep (Monaco, via `TabDiffButton`→`DiffEditorDialog`) is already in the prod bundle through the main YAML editor, so the marginal cost here is small (`CreateFileButton` + the diff wrappers). Fully lazy-loading the toggle is awkward because its module-scope flag-override (lines 29-35) must run eagerly — before the first render and the bootstrap config fetch. I can split the override side-effect from the visual component and `React.lazy` the latter if we want the marginal savings — lmk if it's worth a follow-up.

source/javascripts/layouts/MainLayout.tsx
moczolaszlo: Thanks — the QA-before-merge commitment on the unconditional content wrapper works for me (it's the standard nested-scroll pattern, just wanted the flag-off parity confirmed). Resolving.

source/javascripts/main.tsx
moczolaszlo: Agreed, intended always-on with the belt-and-suspenders `message` arm — negligible false-positive risk, and easy to tighten to `name`-only later if one ever slips through. Resolving.

source/javascripts/components/unified-editor/ConfigSettingsBar/ConfigSettingsBar.tsx
moczolaszlo: Confirmed — master is merged in and the `isModular` sizing is combined with the badge/`showValidationBadge` path; conflict resolved (`MERGEABLE`). Resolving.

source/javascripts/main.tsx
moczolaszlo: Makes sense — Monaco is already in the prod bundle via the main editor, so the marginal cost is small, and the eager flag-override makes full lazy-loading awkward. Fine to leave as-is or split it as a follow-up. Resolving.

source/javascripts/layouts/MainLayout.tsx
```diff
@@ -31,43 +34,66 @@ const MainLayout = () => {
...
+  // between Visual and YAML mode); without it, the strip tops the content area instead.
+  const tabsBesideBar = isModular && isWebsiteMode;
+
+  useMergedConfigSync();
+
+  // Non-modular keeps the original layout: full-width bar on the YAML page, no tabs row.
+  const gridTemplateColumns = isYmlPage && !isModular ? '1fr' : '256px 1fr';
+  let gridTemplateAreas = tabsBesideBar ? `"bar tabs" "nav content"` : `"bar content" "nav content"`;
+  if (isYmlPage) {
+    gridTemplateAreas = isModular ? `"bar tabs" "content content"` : `"bar" "content"`;
```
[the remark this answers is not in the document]
trapacska: Fixed in ca49f95a — both `gridTemplateColumns` and the YAML-page `gridTemplateAreas` now key off `tabsBesideBar` instead of `isModular`. When the beside-bar tab strip isn't present (non-website mode), the tabs render inside content, so the page is now a single full-width column with no orphaned "tabs" area/column. The website modular case (tabsBesideBar = true) is unchanged.

source/javascripts/layouts/MainLayout.tsx
```diff
@@ -31,43 +34,66 @@ const MainLayout = () => {
...
+  // Modular configs get a global file-tab strip atop the editor area.
+  const isModular = Boolean(useTree());
+  // With the bar present, the tab strip shares its 48px top row (so it doesn't jump
+  // between Visual and YAML mode); without it, the strip tops the content area instead.
+  const tabsBesideBar = isModular && isWebsiteMode;
+
+  useMergedConfigSync();
+
+  // Non-modular keeps the original layout: full-width bar on the YAML page, no tabs row.
+  const gridTemplateColumns = isYmlPage && !isModular ? '1fr' : '256px 1fr';
```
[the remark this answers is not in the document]
trapacska: Fixed in ca49f95a — both `gridTemplateColumns` and the YAML-page `gridTemplateAreas` now key off `tabsBesideBar` instead of `isModular`. When the beside-bar tab strip isn't present (non-website mode), the tabs render inside content, so the page is now a single full-width column with no orphaned "tabs" area/column. The website modular case (tabsBesideBar = true) is unchanged.

moczolaszlo (approved):
Approving from my side — all four of my review points are addressed and their threads resolved: the ConfigSettingsBar conflict with #1795 is merged + reconciled, the always-on 'Canceled' rejection filter and the dev-toggle bundling are confirmed/explained, and the unconditional MainLayout content-area wrapper is acknowledged with a flag-off QA commitment. The flag-off safety holds end-to-end (legacy bootstrap byte-identical, layout/bar sizing gated, dev toggle null+inert in prod, clearedModularState in the legacy init), and CI is green.

Two things still open that this approval doesn't override — please make sure they're closed before merging:
1. @k-olly's review threads (GlobalDiffEditorDialog / MergedDiffDialog / ConfigSettingsBar) are their own cycle to resolve.
2. The flag-off scroll/height parity QA on the YAML + entity pages that you committed to on the MainLayout note.

And the standing reminder from the PR description: safe to merge with the flag off, but don't enable the flag for any org until the backend tree/merge/push endpoints are live.

PR #1795, opened 2026-06-16
fix: Visual view shouldn't throw error toasts in case of yml syntax error

k-olly:


source/javascripts/components/Header.tsx
```diff
@@ -292,19 +285,25 @@ const Header = () => {
...
+          onValueChange={(details) => handleEditorViewChange(details.value)}
+        >
+          <BitkitSegmentedControl.Item icon={IconWebUi} value="visual" disabled={ymlStatus === 'invalid'}>
+            Visual
+          </BitkitSegmentedControl.Item>
+          <BitkitSegmentedControl.Item icon={IconCode} value="yaml">
+            YAML
+          </BitkitSegmentedControl.Item>
+        </BitkitSegmentedControl>
+      </BitkitTooltip>
```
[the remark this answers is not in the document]
k-olly: Not working properly with the solution.
I think it's absolutely ok this way, and not misleading - also designer checked it.

source/javascripts/components/unified-editor/ConfigSettingsBar/ConfigSettingsBar.tsx
```diff
@@ -91,34 +110,46 @@ const ConfigSettingsBar = (props: BoxProps) => {
...
+            Change storage...
+          </BitkitActionMenu.Item>
+        </BitkitActionMenu.Root>
+        {showValidationBadge && (
+          <Skeleton loading={isCiConfigLoading} width={isCiConfigLoading ? '96' : undefined} flexShrink="0">
+            <BitkitBadge colorVariant={validationBadge.colorVariant} icon={validationBadge.icon} flexShrink="0">
+              {validationBadge.label}
+            </BitkitBadge>
+          </Skeleton>
+        )}
```
moczolaszlo, line 153: **Bug: the badge briefly flashes "Warnings" before settling on the real status.**

Root cause (pre-existing, newly surfaced by this badge): `validationStatus` is driven by Monaco markers in `MonacoUtils.onModelMarkerStatusChange` (`hasError → invalid`, `hasWarning → warnings`, else `valid`). Markers arrive in stages from multiple owners (monaco-yaml schema layer + the Bitrise language server) and `onDidChangeMarkers` fires per owner update, so during initial load there's a transient window where a warning-severity marker is present but the final error/clear pass hasn't run yet → a momentary `'warnings'`. The `Header` only ever read `=== 'invalid'`, so this transient was invisible until this badge started rendering every state.

Compounding it: the badge's Skeleton is gated only on `isCiConfigLoading` (the network fetch), not on whether validation has settled — so the badge appears (out of skeleton) while the marker status is still mid-flight.

Suggested fix (either, ideally both):
- **Debounce** the status updates (trailing ~250ms) inside `onModelMarkerStatusChange` so the intermediate marker states collapse into the settled value instead of flickering. Minimal and fixes it everywhere (the Header's Visual-enable gating tolerates a 250ms delay fine).
- **Gate the badge on validation readiness**: introduce a `'pending'` initial validation state (set to a real value by the first marker pass) and keep the Skeleton up while pending (`loading={isCiConfigLoading || isValidationPending}`), so the badge goes skeleton → final with no wrong intermediate state ever shown.

Debounce alone removes the visible flicker in practice; adding the pending-gate guarantees the first painted state is always correct.

moczolaszlo (commented):
Nice UX fix — disabling the Visual tab with a tooltip is much better than the toast, and the stories cover the states well. One bug to flag on the new validation badge (details inline). CI is green; the rest of the PR looks clean.

source/javascripts/components/unified-editor/ConfigSettingsBar/ConfigSettingsBar.tsx
moczolaszlo: Follow-up with a copy-pasteable, AI-agent-ready implementation spec for the fix.

---

### Goal
Eliminate the validation-badge flicker (`valid → warnings → final` on initial load). Introduce a `'pending'` validation state that shows a Skeleton until validation has actually settled, and debounce the marker→status updates so intermediate marker passes don't surface.

### Why both changes are required
The marker reader runs an immediate pass on subscribe when **markers are still empty** (`MonacoUtils.ts` ~line 265), which yields a premature `'valid'`. So a naive "first `setValidationStatus` resolves pending" would flip out of pending too early and still flicker. Debouncing makes the **first** emitted status the settled one, so `pending → correct value` directly.

### Change 1 — `source/javascripts/core/utils/MonacoUtils.ts`
Debounce the status updates (trailing). `debounce` is available from `es-toolkit` (already a dependency).

```ts
import { debounce } from 'es-toolkit';

function onModelMarkerStatusChange(
  model: monaco.editor.ITextModel,
  callback: (status: ValidationStatus) => void,
): monaco.IDisposable {
  const updateStatus = () => {
    const markers = monaco.editor.getModelMarkers({ resource: model.uri });
    const hasError = markers.some((m) => m.severity === monaco.MarkerSeverity.Error);
    const hasWarning = markers.some((m) => m.severity === monaco.MarkerSeverity.Warning);

    if (hasError) callback('invalid');
    else if (hasWarning) callback('warnings');
    else callback('valid');
  };

  // Trailing debounce: collapse the multi-owner marker passes (monaco-yaml schema layer
  // + Bitrise LS) that fire during load into a single settled value, so the status never
  // flashes through a transient 'warnings'/'valid' before the real result.
  const debouncedUpdate = debounce(updateStatus, 250);

  // Initial markers are often empty at subscribe time, so don't emit synchronously —
  // let the debounce settle on the first real pass.
  debouncedUpdate();

  const subscription = monaco.editor.onDidChangeMarkers((changedUris) => {
    const modelUri = model.uri.toString();
    if (!changedUris.some((uri) => uri.toString() === modelUri)) {
      return;
    }
    debouncedUpdate();
  });

  return {
    dispose: () => {
      debouncedUpdate.cancel();
      subscription.dispose();
    },
  };
}
```

### Change 2 — `source/javascripts/core/stores/BitriseYmlStore.ts`
Add `'pending'` to the type and make it the initial value. `setValidationStatus` keeps its existing signature (it is only ever called with `valid | invalid | warnings` from the marker reader; `'pending'` is purely the initial state).

```ts
// initial state (currently: validationStatus: 'valid' as 'valid' | 'invalid' | 'warnings')
validationStatus: 'pending' as 'pending' | 'valid' | 'invalid' | 'warnings',
```

```ts
// keep the setter limited to real results — it never sets 'pending'
export function setValidationStatus(status: 'valid' | 'invalid' | 'warnings') {
  bitriseYmlStore.setState({ validationStatus: status });
}
```

### Change 3 — `source/javascripts/hooks/useYmlValidationStatus.ts`
No behavior change beyond the wider return type; the `__invalidYmlString` fast-path still short-circuits to `'invalid'` (unparseable YAML never goes through `pending`). It just passes `'pending'` through now:

```ts
function useYmlValidationStatus() {
  return useBitriseYmlStore((s) => {
    if (s.__invalidYmlString !== undefined) {
      return 'invalid' as const;
    }
    return s.validationStatus; // may be 'pending' until the first settled marker pass
  });
}
```

### Change 4 — `source/javascripts/components/unified-editor/ConfigSettingsBar/ConfigSettingsBar.tsx`
Treat `'pending'` as loading and guard the badge lookup (the Chakra `Skeleton` still renders its child underneath the overlay, so `VALIDATION_BADGE['pending']` must not be dereferenced).

```ts
const ymlStatus = useYmlValidationStatus();
const isValidationPending = ymlStatus === 'pending';
// Fallback so the skeleton-covered child never dereferences an undefined map entry while pending.
const validationBadge = VALIDATION_BADGE[ymlStatus] ?? VALIDATION_BADGE.valid;
const isBadgeLoading = isCiConfigLoading || isValidationPending;
```

```tsx
{showValidationBadge && (
  <Skeleton loading={isBadgeLoading} width={isBadgeLoading ? '96' : undefined} flexShrink="0">
    <BitkitBadge colorVariant={validationBadge.colorVariant} icon={validationBadge.icon} flexShrink="0">
      {validationBadge.label}
    </BitkitBadge>
  </Skeleton>
)}
```

### Edge cases / constraints
- `'pending'` only resolves where the Monaco editor (language services in `useYmlLanguageServices`) is mounted, which is the case on the YAML page where this badge renders (`showValidationBadge={isYmlPage}`). If the badge is ever rendered without the editor mounted, it would stay `pending` forever — add a comment noting this coupling.
- The `__invalidYmlString` path keeps the flicker-free fast path for unparseable YAML (immediate `'invalid'`, no `pending`).
- The `ConfigSettingsBar` Invalid/Warnings/Valid stories set `validationStatus` directly, so they bypass `pending` and keep working; consider adding a `Pending` story (`validationStatus: 'pending'`) to lock in the skeleton state.

### Acceptance
- On first load of the YAML page, the badge shows a Skeleton and then the correct final status — it never briefly shows `Warnings` (or any wrong state) before settling.
- The Header's Visual-tab gating still works (`useYmlValidationStatus() === 'invalid'` disables it); a ~250ms debounce on enabling Visual after fixing the YAML is acceptable.
- Existing `ConfigSettingsBar` and `Header` stories still render.

k-olly: Thank you for the help, changes are done 👍

source/javascripts/core/utils/MonacoUtils.ts
```diff
@@ -245,34 +246,36 @@ function onModelMarkerStatusChange(
...
   };
 
-  // Check existing markers immediately
-  updateStatus();
+  // Trailing debounce: collapse the multi-owner marker passes (monaco-yaml schema layer
+  // + Bitrise LS) that fire during load into a single settled value, so the status never
+  // flashes through a transient 'warnings'/'valid' before the real result.
+  const debouncedUpdate = debounce(updateStatus, 250);
+
+  debouncedUpdate();
```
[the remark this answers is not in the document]
moczolaszlo: Please don't apply this suggestion as written — the immediate `updateStatus()` was removed on purpose. At subscribe time the model's markers are usually still empty, so an immediate call emits a premature `'valid'` that then flips to the real status once the monaco-yaml / Bitrise-LS marker passes land — which is exactly the flicker this PR fixes. The 250ms delay is harmless here because the badge now shows a Skeleton while `validationStatus === 'pending'`, so the user only ever sees skeleton → settled value.

One real consequence worth a note (not a reason to restore the sync call): during the ~250ms `pending` window the Header's Visual-tab gating (`useYmlValidationStatus() === 'invalid'`) sees `'pending'`, so the Visual tab is briefly clickable for *parseable-but-error* YAML. Unparseable YAML is still instantly `'invalid'` via the `__invalidYmlString` fast-path, and the pre-PR code had a similar empty-marker window. If that 250ms window matters, handle it in the Header (treat `'pending'` as "not yet known" and keep Visual disabled until settled) rather than reintroducing the synchronous empty-marker pass. Resolving as intended-by-design.

source/javascripts/components/unified-editor/ConfigSettingsBar/ConfigSettingsBar.tsx
moczolaszlo: Implemented faithfully in 2f6214b7f9 — verified: debounced marker reader (no sync empty-marker pass), `validationStatus` initial `'pending'`, and the badge gated on `isCiConfigLoading || isValidationPending` with a `?? VALIDATION_BADGE.valid` fallback. No more flicker. Thanks @k-olly! Resolving.

moczolaszlo (dismissed):
Approving. The validation-badge flicker fix is implemented faithfully (debounced marker reader + `'pending'` initial state + skeleton gate, verified in 2f6214b7), the merge conflict with master is resolved (kept master's `DiffEditorDialogContent` per-file structure and grafted the pending-safe `useModelValidationStatus` seed), and CI is green.

One non-blocking nit still open (the AI reviewer also raised it, same as my earlier note): the `BitkitTooltip` wraps the whole `BitkitSegmentedControl`, so the "YAML is invalid" tooltip shows when hovering the YAML item too — scoping it to just the disabled Visual item would be cleaner. Fine to address in a follow-up; not blocking the merge.

moczolaszlo (approved):
Re-approving after the debounce refinements (verified): the validation-status reader now uses a long window only for the initial multi-validator settle and a short 250ms window for live edits, so the badge no longer flashes 'Warnings' on load while staying responsive to valid→invalid edits. CI green, merge conflict with master resolved earlier. The one open item (tooltip scoped to the whole control vs just the Visual item) is a non-blocking follow-up nit.

PR #1812, opened 2026-06-22
Fix error notification in Configuration YAML dialog

k-olly:
## Summary

Migrates the Configuration YAML dialog's error notification from the legacy `@bitrise/bitkit` `Notification` to the `@bitrise/bitkit-v2` `BitkitAlert`, and moves it into the dialog footer so it renders correctly.

## Changes

- **`YmlDialogErrorNotification`**: replaced legacy `Notification` with `BitkitAlert`
  - `status="error"` → `variant="critical"`
  - notification body → `messageText`; the bold "Split configuration requires an Enterprise plan" heading is now a proper `titleText` instead of a manual bold `<Text>`
  - the `action` (`href`/`label`/`target`) shape carries over unchanged
- **`ConfigurationYmlStorageDialog`**: moved the error notification out of `BitkitDialog.Body` into `BitkitDialog.Footer`

## Test plan

- [ ] Trigger a save/load error in the Configuration YAML dialog and confirm the `BitkitAlert` renders with the correct error message
- [ ] Verify the 404 case shows the "Couldn't find the bitrise.yml file…" message
- [ ] Verify the "Split configuration requires an Enterprise plan" case shows the title + "Contact us" action button linking to bitrise.io/contact

### DIFF
--- a/source/javascripts/components/unified-editor/UpdateConfigurationDialog/YmlDialogErrorNotification.tsx
+++ b/source/javascripts/components/unified-editor/UpdateConfigurationDialog/YmlDialogErrorNotification.tsx
@@ -1,4 +1,4 @@
-import { Notification, NotificationProps, Text } from '@bitrise/bitkit';
+import { BitkitAlert, BitkitAlertProps } from '@bitrise/bitkit-v2';
 import { ReactNode } from 'react';
 
 import { ClientError } from '@/core/api/client';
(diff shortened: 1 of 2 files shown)


moczolaszlo (approved):
LGTM — clean `@bitrise/bitkit` → `@bitrise/bitkit-v2` migration.

Verified against the bitkit-v2 API:
- `variant="critical"` is the correct error variant (`NOTIFICATION_VARIANTS` has no `error`; `critical` is the one).
- `titleText` / `messageText` / `action` ({ href, label, target }) match `BitkitAlertProps` exactly — and promoting the bold heading to `titleText` (instead of a manual bold `<Text>`) is the right call.
- `BitkitAlertProps extends AlertRootProps`, so the forwarded style props (`marginBlockEnd`) and the `Omit<…> & { error }` + `{...rest}` passthrough are type-safe.

The footer placement is consistent with the warning `BitkitAlert` that already lived there above the buttons, and the `marginBlockEnd` spacing handles the both-alerts case. Prop change stays backward compatible (`error` retained, rest optional).

Nice cleanup. 🎉

PR #1816, opened 2026-06-23
refactor: tooltip info icon (not icon button) for read-only file-tree groups

moczolaszlo:
## What

In `FileTreeView`, the read-only group header showed an info icon via a `BitkitIconButton`. The button has no `onClick` — it existed only to carry the "…read-only" tooltip — so a button is the wrong semantics (and a stray tab stop).

Replace it with a `BitkitTooltip`-wrapped `IconInfoCircle` — a plain, non-interactive info indicator.

```tsx
// before
<BitkitIconButton size="sm" variant="tertiary" label="…read-only." icon={IconInfoCircle} />
// after
<BitkitTooltip text="…read-only."><IconInfoCircle size="16" color="icon/tertiary" /></BitkitTooltip>
```

## Testing
- `npm run lint` — clean
- `tsc --noEmit` — clean

### DIFF
--- a/source/javascripts/components/FileTreeViewer/FileTreeView.tsx
+++ b/source/javascripts/components/FileTreeViewer/FileTreeView.tsx
@@ -1,11 +1,5 @@
 import { Box, Text } from '@bitrise/bitkit';
-import {
-  BitkitIconButton,
-  BitkitTreeView,
-  createTreeCollection,
-  IconFileYml,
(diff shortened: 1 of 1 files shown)


moczolaszlo:
Folded into #1815 — combined the two file-tree / jump-to-definition refactors into a single PR (the read-only info-icon change is now commit 74b65269 there). Closing this one.

PR #1818, opened 2026-06-24
refactor: hide the collapse chevron on ghost cards

moczolaszlo:
Follow-up to the ghost-card work (#1817). Continues tidying the ghost (cross-file) cards.

### Problem
A cross-file ("ghost") workflow / chained workflow / step bundle has no expandable content on this side — its nested steps/workflows live in another file and aren't rendered here (`{!isCrossFile && …}`). The collapse chevron was still shown, just **disabled** — a dead, greyed control that does nothing.

### Change
Hide the chevron entirely for cross-file refs (render-gate on `!isCrossFile`) instead of rendering it disabled. Affected: `WorkflowCard`, `ChainedWorkflowCard`, `StepBundleCard`.

The **merged read-only view** is unchanged: there the content *is* shown (entities resolve locally, so `isCrossFile` is false), so the chevron still renders and works.

Not touched: the `WorkflowConfigHeader` Properties/Triggers tabs that are disabled for cross-file — that's a separate, intentional behaviour.

### Verify
- `tsc`, ESLint clean.

### DIFF
--- a/source/javascripts/components/unified-editor/StepSelectorDrawer/components/StepBundleCard.tsx
+++ b/source/javascripts/components/unified-editor/StepSelectorDrawer/components/StepBundleCard.tsx
@@ -214,10 +214,10 @@ const StepBundleCard = (props: StepBundleCardProps) => {
               onClick={handleClick}
               role={isButton ? 'button' : 'div'}
             >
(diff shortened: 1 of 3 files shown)


moczolaszlo:

PR #1819, opened 2026-06-24
fix: keep the loading bitbot up until the config tree finishes loading

moczolaszlo:
### Problem
On bootstrap the loader runs in two phases:
1. the *"does the repo have a config"* settings check (`useCiConfigSettings`, which gates the config query), and
2. the **tree** fetch that returns the module structure + ymls (`useGetCiConfigTree`).

The progress bitbot showed during phase 1, but `InitialDataLoader` rendered its children as soon as the settings resolved — so during phase 2 (tree still loading) the app rendered against an **empty store** and the bitbot disappeared mid-load.

### Fix
Keep the loading bitbot up until the config has actually loaded, covering **both** phases (and branch switches, as agreed).

The header + navigation stay visible during the load — only the **content area** shows the bitbot:
- `InitialDataLoader` exposes the loading flag (`!data`) via a small `ConfigLoading` context instead of taking over the whole screen.
- `MainLayout` renders `<LoadingState />` inside the content box while the config loads, and the routes once it's ready.

### Verify
- `tsc`, ESLint clean; no import cycle.

### DIFF
--- a/source/javascripts/layouts/ConfigLoading.context.tsx
+++ b/source/javascripts/layouts/ConfigLoading.context.tsx
@@ -1,0 +1,13 @@
+import { createContext, useContext } from 'react';
+
+/**
+ * True while the initial CI config is still loading — covering both bootstrap phases: the
+ * "does the repo have a config" settings check and the subsequent tree/legacy fetch. Provided by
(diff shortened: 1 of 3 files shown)


moczolaszlo:

PR #1820, opened 2026-06-24
fix: close the Open module picker when selecting the already-active file

moczolaszlo:
### Problem
In the **Open module** picker, files that already have a tab are listed in the tree too. Clicking one navigates to its tab and closes the picker — **except** when you click the file that's *already the active tab*: nothing happens and the popover stays open.

### Cause
The tree drove navigation off `BitkitTreeView`'s `onSelectionChange`, which only fires when the selection actually **changes**. Re-clicking the already-selected file doesn't change the selection → no callback → no `onSelect` → the picker never closes.

### Fix
Handle that one case via the leaf's `onClick`, guarded to the currently-selected node:
```tsx
onClick={() => { if (node.id === selectedNodeId) onSelect(node.id); }}
```
Other files keep going through `onSelectionChange` (so it fires exactly once — no double navigation), and keyboard selection is unaffected.

### Verify
- `tsc`, ESLint clean.

### DIFF
--- a/source/javascripts/components/FileTreeViewer/FileTreeView.tsx
+++ b/source/javascripts/components/FileTreeViewer/FileTreeView.tsx
@@ -93,7 +93,18 @@ const GroupTree = ({
         nodeState.isBranch ? (
           <BitkitTreeView.Branch label={node.label} />
         ) : (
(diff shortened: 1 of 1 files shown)


moczolaszlo:

PR #1827, opened 2026-06-29
chore: migrate all dialogs to new @bitrise/bitkit-v2 dialog API

k-olly:
## Summary

- Migrates all 5 dialog components to the new API introduced by the breaking change

## Changes

The new API removes the compound dialog pattern (`BitkitDialogRoot` + `BitkitDialog.FormContent/Body/Footer/Buttons`) and replaces it with self-contained components that accept `footerButtons` and `footerContent` as props.

| File | Change |
|------|--------|
| `SwitchBranchDialog.tsx` | `BitkitDialogRoot` + `BitkitDialog.FormContent` → `BitkitFormDialog` with `footerButtons` |
| `PushBranchDialog.tsx` | `BitkitDialogRoot` + `BitkitDialog.FormContent` → `BitkitFormDialog` with `footerButtons` |
| `DiscardFileTabDialog.tsx` | `BitkitDialog.Body/.Footer/.Buttons` → `footerButtons` prop |
| `UpdateConfigurationDialog.tsx` | Inner `DialogContent` inlined; `BitkitDialog.Body/.Footer/.Buttons` → `footerButtons` prop |
| `ConfigurationYmlStorageDialog.tsx` | Inner `DialogContent` inlined; `BitkitDialog.Body/.Footer/.Buttons` → `footerContent` + `footerButtons` props |

## Test plan

- [ ] Open the Switch Branch dialog — branches list, cancel, and switch all work
- [ ] Open the Push Changes dialog — radio group, branch name validation, commit message, and push work; "Manual update" link works
- [ ] Open a file tab and try to close it with unsaved changes — Discard dialog appears, "Keep editing" and "Discard and close" both work
- [ ] Open Update Configuration YAML dialog — download and copy enable the "Done" button
- [ ] Open Configuration YAML Storage dialog — storage toggle, path input, validate & save work; error and warning alerts appear in footer when applicable

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   },
   "dependencies": {
     "@bitrise/bitkit": "^13.358.0",
-    "@bitrise/bitkit-v2": "0.3.273",
+    "@bitrise/bitkit-v2": "0.3.274-beta.1812",
     "@bitrise/languageserver": "git+https://github.com/bitrise-io/bitrise-yml-language-server.git#833b15995769404349b63986ea779b60fccce531",
     "@dagrejs/dagre": "^3.0.0",
     "@datadog/browser-rum": "^7.3.0",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   },
   "dependencies": {
     "@bitrise/bitkit": "^13.358.0",
-    "@bitrise/bitkit-v2": "0.3.274-beta.1812",
+    "@bitrise/bitkit-v2": "0.3.274",
     "@bitrise/languageserver": "git+https://github.com/bitrise-io/bitrise-yml-language-server.git#833b15995769404349b63986ea779b60fccce531",
     "@dagrejs/dagre": "^3.0.0",
     "@datadog/browser-rum": "^7.3.0",
--- a/source/javascripts/components/unified-editor/UpdateConfigurationDialog/UpdateConfigurationDialog.tsx
+++ b/source/javascripts/components/unified-editor/UpdateConfigurationDialog/UpdateConfigurationDialog.tsx
@@ -12,7 +12,7 @@ import { Box } from '@chakra-ui/react/box';
 import { Card } from '@chakra-ui/react/card';
 import { Table } from '@chakra-ui/react/table';
 import { Text } from '@chakra-ui/react/text';
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import { useCopyToClipboard } from 'usehooks-ts';
 
 import { trackCopyYmlClicked, trackDownloadYmlClicked } from '@/core/analytics/ConfigManagementAnalytics';
@@ -41,6 +41,12 @@ const UpdateConfigurationDialog = ({ isOpen, onClose }: Props) => {
   const [, copyToClipboard] = useCopyToClipboard();
   const { defaultBranch, gitRepoSlug } = PageProps.app() ?? {};
   const [isCopiedOrDownloaded, setIsCopiedOrDownloaded] = useState(false);
+
+  useEffect(() => {
+    if (!isOpen) {
+      setIsCopiedOrDownloaded(false);
+    }
+  }, [isOpen]);
 
   const isModular = useBitriseYmlStore((s) => Boolean(s.tree));
   const changedModules = useBitriseYmlStore(
--- a/source/javascripts/components/unified-editor/UpdateConfigurationDialog/UpdateConfigurationDialog.tsx
+++ b/source/javascripts/components/unified-editor/UpdateConfigurationDialog/UpdateConfigurationDialog.tsx
@@ -37,7 +37,7 @@ type ChangedFileRow = {
 
 const moduleCountLabel = (count: number) => `${count} ${count === 1 ? 'module' : 'modules'} changed`;
 
-const DialogContent = ({ onClose }: Pick<Props, 'onClose'>) => {
+const UpdateConfigurationDialog = ({ isOpen, onClose }: Props) => {
   const [, copyToClipboard] = useCopyToClipboard();
   const { defaultBranch, gitRepoSlug } = PageProps.app() ?? {};
   const [isCopiedOrDownloaded, setIsCopiedOrDownloaded] = useState(false);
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -46,7 +46,7 @@
   },
   "dependencies": {
     "@bitrise/bitkit": "^13.358.0",
-    "@bitrise/bitkit-v2": "0.3.273",
+    "@bitrise/bitkit-v2": "0.3.274-beta.1812",
```
moczolaszlo: Blocker: pinned to a **beta** (`0.3.274-beta.1812`) — `latest` is still `0.3.273`, i.e. the new Dialog API isn't on a stable release yet (gated on the approved-but-unmerged Bitkit #251). Repin to a stable `0.3.274+` before merge. Exact pin (no caret) is correct; only the beta-vs-stable point remains.

source/javascripts/components/unified-editor/UpdateConfigurationDialog/UpdateConfigurationDialog.tsx
```diff
@@ -37,7 +37,7 @@ type ChangedFileRow = {
 
 const moduleCountLabel = (count: number) => `${count} ${count === 1 ? 'module' : 'modules'} changed`;
 
-const DialogContent = ({ onClose }: Pick<Props, 'onClose'>) => {
+const UpdateConfigurationDialog = ({ isOpen, onClose }: Props) => {
```
moczolaszlo, line 40: State-persistence regression: `isCopiedOrDownloaded` (declared just below) now lives in this always-mounted component instead of the old lazily-mounted `DialogContent` child. So after a user copies/downloads once and closes, reopening shows the "Done" button already enabled instead of disabled. Reset on close, e.g. `useEffect(() => { if (!isOpen) setIsCopiedOrDownloaded(false) }, [isOpen])`, or move the state back into an inner component rendered only while open.

source/javascripts/pages/YmlPage/components/ConfigurationYmlStorageDialog.tsx
```diff
@@ -268,7 +268,7 @@ const GitToBitriseSection = ({ isDisabled, lastModifiedFormatted, onChange, valu
   );
 };
 
-const DialogContent = ({ onClose }: Pick<ConfigurationYmlStorageDialogProps, 'onClose'>) => {
+const ConfigurationYmlSourceDialog = ({ isOpen, onClose }: ConfigurationYmlStorageDialogProps) => {
```
moczolaszlo, line 271: Same state-persistence regression, larger surface: `selectedStorage` / `ymlRootPath` / `gitToBitriseStorage` / `asyncError` and the seeding `useEffect`s now live in the always-mounted component. Consequence: open → change the storage selection → close without saving → reopen shows the **stale** prior selection instead of a fresh state seeded from current settings, and the seeding effect also runs while the dialog is closed. Re-seed/reset on `isOpen`, or keep this logic in an inner component rendered only when open. (The `useCiConfigSettings()` query running while closed is fine — shared query key, no extra request.)

moczolaszlo (dismissed):
## Review — migrate dialogs to the new Bitkit v2 Dialog API

Well-executed migration overall: footer mapping (`footerButtons` fragment-wrapped, alerts → `footerContent`), `BitkitFormDialog` for the `<form>` dialogs (PushBranch/SwitchBranch), and removal of the old scaffolding are all correct. No leftover old API anywhere, dialogs are mounted unconditionally and driven by `open` (exit animations preserved). CI green.

Requesting changes for two state-persistence regressions introduced by hoisting hooks out of the previously lazily-mounted dialog child, plus the version pin.

The pattern: the old code kept `useState`/seeding hooks inside a `DialogContent` child that was passed as `children`, so it only mounted while the dialog was open (Chakra `Dialog.Content` lazy-mounts). The migration moved those hooks into the now always-mounted top-level component, so their state no longer resets between close and reopen. See inline comments on the two affected dialogs. Fix either way: reset/re-seed on `isOpen` (`useEffect(() => { if (!isOpen) reset() }, [isOpen])`), or keep the state in an inner component that only renders while open.

source/javascripts/components/unified-editor/UpdateConfigurationDialog/UpdateConfigurationDialog.tsx
k-olly: Done

source/javascripts/pages/YmlPage/components/ConfigurationYmlStorageDialog.tsx
k-olly: Done

package.json
moczolaszlo: Resolved — now pinned to stable `0.3.274` (the new Dialog API shipped to `latest` via bitkit-v2 #251). Thanks!

source/javascripts/components/unified-editor/UpdateConfigurationDialog/UpdateConfigurationDialog.tsx
moczolaszlo: Resolved in 38ae56e6 — `useEffect(() => { if (!isOpen) setIsCopiedOrDownloaded(false) }, [isOpen])` resets it on close. Thanks!

source/javascripts/pages/YmlPage/components/ConfigurationYmlStorageDialog.tsx
moczolaszlo: Resolved in 38ae56e6 — the seeding effect is now gated on `isOpen` and re-seeds all four state values (and clears `asyncError`) on open. Thanks!

moczolaszlo (approved):
## Approve ✅

Both state-persistence regressions are fixed correctly (38ae56e6): `UpdateConfigurationDialog` resets `isCopiedOrDownloaded` on close, and `ConfigurationYmlStorageDialog` re-seeds its state (and clears `asyncError`) on open via the `isOpen`-gated effect. Version is now on stable `0.3.274`. Clean migration — thanks!

PR #1828, opened 2026-06-30
Add per-file conflict resolution for modular YML push [BIVS-3703]

trapacska:
When a modular push 409s, resolve every conflicting module file individually instead of treating the config as one blob.

- New tabbed ModularConfigMergeDialog: one tab per conflicting file, 3-way auto-merge per file (remote-wins + editable), re-push after resolution with the reconciled branch HEAD; loops on nested conflicts.
- BranchesApi.parsePushConflict parses the per-file 409 body; usePushBranch forwards it to onMergeConflict and supports a baseCommitSha override.
- Header opens the new dialog for modular conflicts; the legacy single-file ConfigMergeDialog is left untouched.
- Scoped to the modular path (enable-wfe-modular-yaml-editing); flag-off behaviour is unchanged.

### DIFF
--- a/source/javascripts/components/ConfigMergeDialog/ModularConfigMergeDialog.tsx
+++ b/source/javascripts/components/ConfigMergeDialog/ModularConfigMergeDialog.tsx
@@ -1,0 +1,328 @@
+import { Box, Button, ButtonGroup, DialogBody, DialogFooter, Icon, Notification, Text, Tooltip } from '@bitrise/bitkit';
+import { BitkitTabs } from '@bitrise/bitkit-v2';
+import { DiffEditor, MonacoDiffEditor } from '@monaco-editor/react';
+import { ModalCloseButton, ModalHeader } from 'chakra-ui-2--react';
+import type { editor } from 'monaco-editor';
+import { useMemo, useRef, useState } from 'react';
+import { useEventListener } from 'usehooks-ts';
+
+import { PushBranchConflict } from '@/core/api/BranchesApi';
+import TreeService from '@/core/services/TreeService';
+import { getFileSlice, updateFileDocumentByString } from '@/core/stores/BitriseYmlStore';
+import MonacoUtils from '@/core/utils/MonacoUtils';
+import YmlUtils from '@/core/utils/YmlUtils';
+import usePushBranch from '@/hooks/usePushBranch';
+
+import { DiffEditorDialogShell } from '../DiffEditor/DiffEditorDialog';
+import { diffEditorOptions, mergeYamls, readOnlyDiffEditorOptions } from './mergeYamls';
+
+type Props = {
+  isOpen: boolean;
+  onClose: VoidFunction;
+  branch: string;
+  message: string;
+  conflict: PushBranchConflict;
+};
+
+/** Per-file 3-way merge state derived once from the conflict + the local file slice. */
+type FileMerge = {
+  nodeId: string;
+  path: string;
+  name: string;
+  yourYml: string;
+  baseYml: string;
+  remoteYml: string;
+  mergedYml: string;
+  decorations: editor.IModelDeltaDecoration[];
+};
+
+function buildFileMerges(conflict: PushBranchConflict): FileMerge[] {
+  // The BE may list the same file more than once; dedupe by node_id so each
+  // conflicting file gets exactly one tab (tab value/selection keys off node_id).
+  const seen = new Set<string>();
+  const uniqueFiles = conflict.conflicts.filter((file) => {
+    if (seen.has(file.nodeId)) {
+      return false;
+    }
+    seen.add(file.nodeId);
+    return true;
+  });
+
+  return uniqueFiles.map((file) => {
+    const slice = getFileSlice(file.nodeId);
+    const yourYml = slice ? YmlUtils.toYml(slice.ymlDocument) : '';
+    const baseYml = slice ? YmlUtils.toYml(slice.savedYmlDocument) : '';
+    const { mergedYml, decorations } = mergeYamls(yourYml, baseYml, file.remoteYml);
+
+    return {
+      nodeId: file.nodeId,
+      path: file.path,
+      name: TreeService.fileName(file.path),
+      yourYml,
+      baseYml,
+      remoteYml: file.remoteYml,
+      mergedYml,
+      decorations,
+    };
+  });
+}
+
+const ModularConfigMergeDialogBody = ({
+  onClose,
+  branch,
+  message,
+  initialConflict,
+}: {
+  onClose: VoidFunction;
+  branch: string;
+  message: string;
+  initialConflict: PushBranchConflict;
+}) => {
+  // Held in state so a nested conflict (someone pushed again while resolving) can
+  // replace the whole set and re-seed the per-file merges from the new remote.
+  const [conflict, setConflict] = useState(initialConflict);
+  const fileMerges = useMemo(() => buildFileMerges(conflict), [conflict]);
+
+  const [selectedTab, setSelectedTab] = useState(fileMerges[0]?.nodeId ?? '');
+  const effectiveTab = fileMerges.some((f) => f.nodeId === selectedTab) ? selectedTab : (fileMerges[0]?.nodeId ?? '');
+  const activeMerge = fileMerges.find((f) => f.nodeId === effectiveTab);
+
+  // Resolved text + validity per file, keyed by node_id. Seeded from each file's
+  // auto-merge so unvisited tabs still contribute their merged result on apply.
+  const [resolved, setResolved] = useState<Record<string, string>>(() =>
+    Object.fromEntries(fileMerges.map((f) => [f.nodeId, f.mergedYml])),
+  );
+  const [validity, setValidity] = useState<Record<string, 'valid' | 'invalid' | 'warnings'>>(() =>
+    Object.fromEntries(fileMerges.map((f) => [f.nodeId, 'valid' as const])),
+  );
+  // Files the user has edited — skip re-decorating them when their tab remounts.
+  const touched = useRef<Set<string>>(new Set());
+  const [clientError, setClientError] = useState<string>();
+
+  const seedFromConflict = (next: PushBranchConflict) => {
+    const merges = buildFileMerges(next);
+    setConflict(next);
+    setResolved(Object.fromEntries(merges.map((f) => [f.nodeId, f.mergedYml])));
+    setValidity(Object.fromEntries(merges.map((f) => [f.nodeId, 'valid' as const])));
+    touched.current = new Set();
+    setSelectedTab(merges[0]?.nodeId ?? '');
+  };
+
+  const { isPushPending, pushBranch } = usePushBranch({
+    onSuccess: onClose,
+    onMergeConflict: (_branch, nextConflict) => {
+      if (nextConflict) {
+        setClientError('There are changes outside your current editing session. Review and resolve them again.');
+        seedFromConflict(nextConflict);
+      } else {
+        setClientError('Failed to push the resolved changes. Please try again.');
+      }
+    },
+  });
+
+  const onResultEditorMount = (mountedNodeId: string, decorations: editor.IModelDeltaDecoration[]) => {
+    return (diff: MonacoDiffEditor) => {
+      const modified = diff.getModifiedEditor();
+
+      modified.onDidChangeModelContent((e) => {
+        touched.current.add(mountedNodeId);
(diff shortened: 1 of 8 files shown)


source/javascripts/components/ConfigMergeDialog/ModularConfigMergeDialog.tsx
```diff
@@ -0,0 +1,328 @@
...
+        modified.removeDecorations(Array.from(removable));
+      });
+
+      if (!touched.current.has(mountedNodeId)) {
+        diff.createDecorationsCollection(decorations);
+      }
+
+      const model = modified.getModel();
+      if (model) {
+        MonacoUtils.onModelMarkerStatusChange(model, (status) =>
```
moczolaszlo: `MonacoUtils.onModelMarkerStatusChange(...)` returns a disposable that's never disposed, and `onMount` has no React cleanup. Each tab switch remounts the active Results editor (`key={`${nodeId}-results`}`), so every visit registers a new global marker listener while the old one stays alive (the model also survives via `keepCurrentModifiedModel`). Over repeated tab switches these accumulate and keep firing `setValidity` for stale models — a listener leak plus redundant re-renders. (The `onDidChangeModelContent` listener is editor-scoped so it's disposed with the editor; this one isn't.) Suggest capturing the disposable and disposing it when the editor unmounts.

source/javascripts/components/ConfigMergeDialog/mergeYamls.ts
```diff
@@ -0,0 +1,86 @@
...
+      }
+
+      if (!remoteChangeIsADeletion) {
+        decorations.push({
+          options: {
+            isWholeLine: true,
+            blockClassName: 'conflict',
+          },
+          range: {
+            startLineNumber: region.conflict.bIndex + 1,
```
moczolaszlo: The conflict decoration line numbers come from diff3's input indices (`bIndex` here, `oIndex` for the deletion branch above), but they're applied to the merged-result editor. The merged text is `region.ok` + `region.conflict.b` concatenated in order, so a conflict's real line in the output is the count of lines emitted before it — generally not `bIndex` (an offset into the remote input). For a single whole-file conflict (the spec case) `bIndex === 0` so it's correct, but with multiple regions the red marker lands on the wrong line (and the deletion branch marks `oIndex+1`, a base-relative line that doesn't exist in the merged output at all). Decorations should be positioned by the running merged-output line count.

source/javascripts/components/ConfigMergeDialog/ModularConfigMergeDialog.tsx
```diff
@@ -0,0 +1,328 @@
...
+  const [selectedTab, setSelectedTab] = useState(fileMerges[0]?.nodeId ?? '');
+  const effectiveTab = fileMerges.some((f) => f.nodeId === selectedTab) ? selectedTab : (fileMerges[0]?.nodeId ?? '');
+  const activeMerge = fileMerges.find((f) => f.nodeId === effectiveTab);
+
+  // Resolved text + validity per file, keyed by node_id. Seeded from each file's
+  // auto-merge so unvisited tabs still contribute their merged result on apply.
+  const [resolved, setResolved] = useState<Record<string, string>>(() =>
+    Object.fromEntries(fileMerges.map((f) => [f.nodeId, f.mergedYml])),
+  );
+  const [validity, setValidity] = useState<Record<string, 'valid' | 'invalid' | 'warnings'>>(() =>
```
moczolaszlo, line 105: `validity` is seeded `'valid'` for every file and only updated for the *mounted* (visited) tab via Monaco markers. If an auto-merged file is invalid YAML but the user never opens its tab, its validity stays `'valid'`, so `isInvalid` is false and **Apply changes** stays enabled — the client-side "YAML is invalid" guard is bypassed for unvisited tabs. The BE will reject it, but the user gets the generic "Failed to push" instead of the inline invalid-YAML hint. Consider validating each file's merged result up front (not only on mount).

source/javascripts/components/ConfigMergeDialog/ModularConfigMergeDialog.tsx
```diff
@@ -0,0 +1,328 @@
...
+              }
+            });
+          }
+          return acc;
+        }, new Set<string>());
+        modified.removeDecorations(Array.from(removable));
+      });
+
+      if (!touched.current.has(mountedNodeId)) {
+        diff.createDecorationsCollection(decorations);
```
moczolaszlo, line 166: On remount, decorations are only re-applied when the file is *not* in `touched` (`if (!touched.current.has(mountedNodeId))`). So after you edit one conflict in a tab and switch away and back, the editor remounts but decorations are skipped — which also drops the markers for the conflicts you *haven't* resolved yet, making remaining conflicts hard to find. Consider re-deriving/re-applying the still-unresolved decorations on remount instead of skipping all of them for touched files.

moczolaszlo (commented):
A few things worth a look on the modular conflict dialog — mostly the marker-listener leak (1) and the client-side validity guard being bypassable (3). The parse + usePushBranch changes look solid. 🤖 review via Claude Code

k-olly:
Review — Per-file conflict resolution for modular YML push

Solid, well-scoped feature: `BranchesApi.parsePushConflict` cleanly separates the modular vs. legacy 409, the `usePushBranch` `baseCommitSha` override correctly re-pushes against the reconciled branch HEAD, and the new logic is properly unit-tested (`BranchesApi.spec.ts`, `usePushBranch.spec.tsx`). Seeding `validity` to `valid` so unopened tabs don't block Apply is a nice touch, and the snake_case→camelCase mapping with undefined fallbacks is thorough.

Requesting changes on one Bitkit-convention point (the main blocker you flagged), plus one behavioral question and a couple of minor nits.

**1. Mixed Bitkit v1/v2 in a brand-new file (convention violation) — main point**

`ModularConfigMergeDialog.tsx` is a new file but pulls almost everything from v1 / Chakra v2, with a single v2 component dropped in:

```tsx
import { Box, Button, ButtonGroup, DialogBody, DialogFooter, Icon, Notification, Text, Tooltip } from '@bitrise/bitkit'; // v1
import { BitkitTabs } from '@bitrise/bitkit-v2';                                                                         // v2 (only this)
import { ModalCloseButton, ModalHeader } from 'chakra-ui-2--react';                                                      // Chakra v2
```

CLAUDE.md/AGENTS.md is explicit — *"use v2 for all new work"*. As-is this is the worst of both worlds: not consistently v1, not v2, and a Chakra-v3 `BitkitTabs` mounted inside an otherwise Chakra-v2 modal tree.

I checked that the migration is realistic before requesting it:
- **Every v1 component used has a v2 equivalent** — `BitkitButton`, `BitkitAlert` (for `Notification`), `BitkitTooltip`, `BitkitDialog` (which supports `size="full"` — there's a dedicated "Full" story), `Box`/`Text` from Chakra v3, v2 icons (`IconCross`, `IconArrowRight`, …).
- **v2 dialogs are already the norm in this exact push flow** — `PushBranchDialog`, `UpdateConfigurationDialog`, and `SwitchBranchDialog` are all `BitkitDialog`. So the sibling dialog in the same flow is already v2.

The reason it landed on v1 is that it reuses the legacy `DiffEditorDialogShell` (`DiffEditorDialog.tsx:166`), which is itself a v1 Chakra-v2 `Dialog` — the component is effectively a clone of the legacy `ConfigMergeDialog.tsx`, so it inherited the shell's stack.

**Suggested direction:** rebuild on a v2 `BitkitDialog` (`size="full"`) using the `BitkitDialog.Body`/`Footer`/`Buttons` slots instead of the v1 shell, matching `PushBranchDialog` in the same flow. If keeping the legacy shell is a conscious call for now, then at minimum don't mix — drop back to the v1 `Tabs` so the tree is consistently Chakra v2, and migrate the whole `ConfigMergeDialog`/`DiffEditor` subsystem to v2 in a separate ticket. Given v2 dialogs already exist in this flow, I'd go with the full v2 rewrite.

**2. Stale editors on nested-conflict re-seed (behavioral question)**

`seedFromConflict` (`ModularConfigMergeDialog.tsx:108`) swaps `conflict` in state when someone pushes again mid-resolution. But the body is remounted via `key={conflict.commitSha}` (`:319`) off the **prop**, which never changes on a nested conflict (`Header`'s `modularConflict` state isn't updated) — so only the manual re-seed runs, not the remount. Meanwhile the `DiffEditor`s are keyed by `activeMerge.nodeId`; if the new conflict reuses the same node IDs, the keys don't change, and with `keepCurrentModifiedModel`/`keepCurrentOriginalModel` Monaco keeps its existing models — so the fresh `remoteYml`/`baseYml`/`mergedYml` may not actually render.

The two mechanisms (key-remount on the prop + manual seed on state) are redundant and overlapping. **Suggested direction:** key the editors off `conflict.commitSha` from *state* inside the still-mounted body, and drop the parent-level remount. Is there a test or manual repro for the push-during-resolution path?

**3. Minor nits**

- **`mergeYamls.ts` duplication** — the new `mergeYamls.ts` is a verbatim copy of the legacy `mergeYamls` in `ConfigMergeDialog.tsx`. The doc comment owns this ("legacy keeps its own copy until the flag graduates"), but since it's now a shared module, the legacy dialog could just import from it at zero risk — one fewer file to delete when the flag goes.
- **Raw color codes** — `bg="rgb(30,30,30)"` and `opacity="0.9"` (`:223`) violate the "always semantic colors" rule. Copied from the legacy dialog (Monaco dark-theme container), so not introduced here — but worth a semantic token (or at least a comment on why it's raw) when this moves to v2.

source/javascripts/components/ConfigMergeDialog/ModularConfigMergeDialog.tsx
```diff
@@ -0,0 +1,328 @@
...
+      }
+
+      const model = modified.getModel();
```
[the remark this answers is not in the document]
trapacska: Fixed in 88cad682 — the marker listener (and the content-change listener) are now collected in `onMount` and disposed via `diff.onDidDispose`, so nothing leaks across the per-tab remounts.

source/javascripts/components/ConfigMergeDialog/ModularConfigMergeDialog.tsx
trapacska: Fixed in 88cad682 — both listeners are now collected in `onMount` and disposed on `diff.onDidDispose`, so the global `onModelMarkerStatusChange` listener no longer accumulates across tab-switch remounts. Good catch on the distinction vs. the editor-scoped content listener.

source/javascripts/components/ConfigMergeDialog/mergeYamls.ts
trapacska: Fixed in 88cad682 — decorations are now positioned by the running merged-output line count instead of diff3 `bIndex`/`oIndex`, and a pure remote deletion marks the boundary line (clamped to line 1). Added a multi-region regression test in `mergeYamls.spec.ts` asserting the second conflict lands on output line 3.

source/javascripts/components/ConfigMergeDialog/ModularConfigMergeDialog.tsx
trapacska: Fixed in 88cad682 — per-file validity is now seeded up front from a parse of each merged result (`seedValidity` → `YmlUtils.toDoc(...).errors`), so an invalid auto-merge in an unvisited tab keeps **Apply changes** disabled. The Monaco marker listener still refines the visited tab live.

source/javascripts/components/ConfigMergeDialog/ModularConfigMergeDialog.tsx
trapacska: Partially addressed in 88cad682: an unedited tab now restores its markers on revisit. For an already-edited file we intentionally do NOT re-apply the original decorations — after free-form edits the original line positions no longer map to the displayed text, so re-applying them would mark the wrong lines. Doing this correctly needs a single persistent Results editor (swap models per tab instead of remounting) so Monaco tracks decoration positions live across edits; we are accepting the current behavior for this PR and tracking the persistent-editor refactor as a follow-up. Within an edited tab the base↔result diff still guides the user.

moczolaszlo (approved):
All four review findings are addressed in 88cad682 and verified: the marker/content listeners are now disposed on editor unmount, decorations are positioned by the merged-output line count, per-file validity is seeded up front (so unvisited tabs gate Apply), and unedited tabs restore their markers on revisit. Nice per-file conflict flow — LGTM. 🤖 via Claude Code

PR #1831, opened 2026-07-01
Tool catalog: make tool ID a dropdown of known tools

ofalvai:
*PR status: Stacked on top of #1834 for now*

Scope of PR: just the tool ID selector. Tool version selection is going to be wired to the catalog API as well in a follow-up PR.

### Decisions

1. Tool catalog doesn't contain all working tools: the dropdown suggests known tools, but users can still type any custom tool ID (via "Other" or when the catalog is loading/errored).
2. Tool alias matching: catalog entries carry a canonical `name` plus `aliases` (e.g. `golang`/`go`), and the dropdown/known-ID logic treats them as equivalent while preserving whichever spelling the user already has in their YAML.
3. Custom tool input with know tool ID: when `Other` and the free-form input is active, typing a known tool ID turns the UI back to a dropdown with that known tool ID selected.

### DIFF
--- a/source/javascripts/components/ToolVersions/ToolRow.tsx
+++ b/source/javascripts/components/ToolVersions/ToolRow.tsx
@@ -31,7 +31,6 @@ type ToolRowProps = {
   catalog: ToolCatalog | undefined;
   allowUnset?: boolean;
   autoFocus?: boolean;
-  isCatalogReady: boolean;
   isCatalogLoading: boolean;
   onIdChange: (newId: string) => void;
   onStrategyChange: (strategy: VersionStrategy, version: string) => void;
@@ -47,7 +46,6 @@ const ToolRow = ({
   catalog,
   allowUnset,
   autoFocus,
-  isCatalogReady,
   isCatalogLoading,
   onIdChange,
   onStrategyChange,
@@ -58,6 +56,7 @@ const ToolRow = ({
   const [manualOther, setManualOther] = useState(false);
   const [idError, setIdError] = useState<string | undefined>();
 
+  const isCatalogReady = !!catalog;
   const isToolIdKnown = ToolsService.isKnownToolId(catalog, toolId);
   const dropdownOptions = ToolsService.getAvailableToolIdOptions(catalog, toolId, existingToolIds);
 
--- a/source/javascripts/components/ToolVersions/ToolRow.tsx
+++ b/source/javascripts/components/ToolVersions/ToolRow.tsx
@@ -31,8 +31,6 @@ type ToolRowProps = {
   catalog: ToolCatalog | undefined;
   allowUnset?: boolean;
   autoFocus?: boolean;
-  dropdownOptions: { value: string; label: string }[];
-  isToolIdKnown: boolean;
   isCatalogReady: boolean;
   isCatalogLoading: boolean;
   onIdChange: (newId: string) => void;
@@ -49,8 +47,6 @@ const ToolRow = ({
   catalog,
   allowUnset,
   autoFocus,
-  dropdownOptions,
-  isToolIdKnown,
   isCatalogReady,
   isCatalogLoading,
   onIdChange,
@@ -61,6 +57,9 @@ const ToolRow = ({
   // Whether the user has explicitly picked "Other" from the dropdown.
   const [manualOther, setManualOther] = useState(false);
   const [idError, setIdError] = useState<string | undefined>();
+
+  const isToolIdKnown = ToolsService.isKnownToolId(catalog, toolId);
+  const dropdownOptions = ToolsService.getAvailableToolIdOptions(catalog, toolId, existingToolIds);
 
   // Only treat a tool as custom once the catalog has actually resolved. While it's
   // still loading, or if it failed to load, an unknown toolId isn't proof it's custom.
@@ -93,8 +92,8 @@ const ToolRow = ({
       return;
     }
     setIdError(undefined);
+    setManualOther(false);
     if (newId !== toolId) {
-      setManualOther(false);
       onIdChange(newId);
     }
   };
@@ -116,6 +115,7 @@ const ToolRow = ({
             items={dropdownItems}
             value={showCustomInput ? OTHER_VALUE : toolId}
             onValueChange={handleDropdownChange}
+            triggerProps={showCustomInput ? undefined : { autoFocus }}
           />
           {showCustomInput && (
             <BitkitTextInput
--- a/source/javascripts/components/ToolVersions/ToolRow.tsx
+++ b/source/javascripts/components/ToolVersions/ToolRow.tsx
@@ -81,7 +81,9 @@ const ToolRow = ({
     }
     setManualOther(false);
     setIdError(undefined);
-    onIdChange(newValue);
+    if (newValue !== toolId) {
+      onIdChange(newValue);
+    }
   };
 
   const handleIdBlur: FocusEventHandler<HTMLInputElement> = (e) => {
--- a/source/javascripts/components/ToolVersions/ToolRow.tsx
+++ b/source/javascripts/components/ToolVersions/ToolRow.tsx
@@ -6,6 +6,7 @@ import {
   BitkitTextInput,
   IconMinusCircle,
   IconOpenInNew,
+  rem,
 } from '@bitrise/bitkit-v2';
 import { Box } from '@chakra-ui/react/box';
 import { Text } from '@chakra-ui/react/text';
@@ -104,7 +105,7 @@ const ToolRow = ({
   return (
     <Box display="flex" flexDirection="column" gap="8">
       <Box display="flex" alignItems="flex-start" gap="12">
-        <Box display="flex" flexDirection="column" gap="8" width="160px" flexShrink="0">
+        <Box display="flex" flexDirection="column" gap="8" width={rem(160)} flexShrink="0">
           <BitkitSelect
             size="md"
             placeholder="Select one"
@@ -142,7 +143,7 @@ const ToolRow = ({
           {strategy !== 'unset' && (
             <BitkitTextInput
               size="md"
-              width="160px"
+              width={rem(160)}
               flexShrink="0"
               placeholder={strategy === 'exact' ? 'e.g. 24.7.0' : 'prefix, e.g. 22'}
               inputProps={{
--- a/source/javascripts/components/ToolVersions/ToolRow.tsx
+++ b/source/javascripts/components/ToolVersions/ToolRow.tsx
@@ -1,5 +1,14 @@
-import { BitkitIconButton, BitkitNativeSelect, BitkitTextInput, IconMinusCircle } from '@bitrise/bitkit-v2';
+import {
+  BitkitIconButton,
+  BitkitLink,
+  BitkitNativeSelect,
+  BitkitSelect,
+  BitkitTextInput,
+  IconMinusCircle,
+  IconOpenInNew,
+} from '@bitrise/bitkit-v2';
 import { Box } from '@chakra-ui/react/box';
+import { Text } from '@chakra-ui/react/text';
 import { FocusEventHandler, useState } from 'react';
 
 import { VersionStrategy } from '@/core/models/Tools';
@@ -12,6 +21,8 @@ const STRATEGY_LABELS: Record<VersionStrategy, string> = {
   unset: 'Do nothing (unset global setting)',
 };
 
+const OTHER_VALUE = '__other__';
+
 type ToolRowProps = {
   toolId: string;
   strategy: VersionStrategy;
@@ -19,6 +30,10 @@ type ToolRowProps = {
   existingToolIds: string[];
   allowUnset?: boolean;
   autoFocus?: boolean;
+  dropdownOptions: { value: string; label: string }[];
+  isToolIdKnown: boolean;
+  isCatalogReady: boolean;
+  isCatalogLoading: boolean;
   onIdChange: (newId: string) => void;
   onStrategyChange: (strategy: VersionStrategy, version: string) => void;
   onVersionChange: (version: string) => void;
@@ -32,12 +47,39 @@ const ToolRow = ({
   existingToolIds,
   allowUnset,
   autoFocus,
+  dropdownOptions,
+  isToolIdKnown,
+  isCatalogReady,
+  isCatalogLoading,
   onIdChange,
   onStrategyChange,
   onVersionChange,
   onRemove,
 }: ToolRowProps) => {
+  // Whether the user has explicitly picked "Other" from the dropdown.
+  const [manualOther, setManualOther] = useState(false);
   const [idError, setIdError] = useState<string | undefined>();
+
+  // Only treat a tool as custom once the catalog has actually resolved. While it's
+  // still loading, or if it failed to load, an unknown toolId isn't proof it's custom.
+  const showCustomInput = manualOther || (isCatalogReady && toolId !== '' && !isToolIdKnown);
+
+  const dropdownItems = [
+    ...dropdownOptions,
+    // Keep the current value selectable while the catalog hasn't confirmed it one way or the other.
+    ...(!showCustomInput && toolId !== '' && !isToolIdKnown ? [{ value: toolId, label: toolId }] : []),
+    { value: OTHER_VALUE, label: 'Other' },
+  ];
+
+  const handleDropdownChange = (newValue: string) => {
+    if (newValue === OTHER_VALUE) {
+      setManualOther(true);
+      return;
+    }
+    setManualOther(false);
+    setIdError(undefined);
+    onIdChange(newValue);
+  };
 
   const handleIdBlur: FocusEventHandler<HTMLInputElement> = (e) => {
     const newId = e.target.value.trim();
@@ -48,6 +90,7 @@ const ToolRow = ({
     }
     setIdError(undefined);
     if (newId !== toolId) {
+      setManualOther(false);
       onIdChange(newId);
     }
   };
@@ -59,47 +102,75 @@ const ToolRow = ({
   };
 
   return (
-    <Box display="flex" alignItems="flex-start" gap="12">
-      <BitkitTextInput
-        size="md"
-        width="160px"
-        flexShrink="0"
-        placeholder="Tool (ruby, node, etc.)"
-        errorText={idError}
-        inputProps={{ defaultValue: toolId, autoFocus, onBlur: handleIdBlur }}
-      />
-
-      <Box display="flex" flex="1" alignItems="flex-start" gap="8">
-        <BitkitNativeSelect
-          flex="1"
-          size="md"
-          value={strategy}
-          onChange={(e) => handleStrategyChange(e.target.value as VersionStrategy)}
-        >
-          {Object.entries(STRATEGY_LABELS)
-            .filter(([value]) => allowUnset || value !== 'unset')
-            .map(([value, label]) => (
-              <option key={value} value={value}>
-                {label}
-              </option>
-            ))}
-        </BitkitNativeSelect>
-
-        {strategy !== 'unset' && (
-          <BitkitTextInput
+    <Box display="flex" flexDirection="column" gap="8">
+      <Box display="flex" alignItems="flex-start" gap="12">
+        <Box display="flex" flexDirection="column" gap="8" width="160px" flexShrink="0">
+          <BitkitSelect
             size="md"
-            width="160px"
-            flexShrink="0"
-            placeholder={strategy === 'exact' ? 'e.g. 24.7.0' : 'prefix, e.g. 22'}
-            inputProps={{
-              value: version,
-              onChange: (e) => onVersionChange(e.target.value),
-            }}
+            placeholder="Select one"
+            isLoading={isCatalogLoading}
+            items={dropdownItems}
+            value={showCustomInput ? OTHER_VALUE : toolId}
+            onValueChange={handleDropdownChange}
           />
-        )}
+          {showCustomInput && (
+            <BitkitTextInput
+              size="md"
+              placeholder="Tool ID (e.g. deno)"
+              errorText={idError}
+              inputProps={{ defaultValue: toolId, autoFocus, onBlur: handleIdBlur }}
+            />
+          )}
+        </Box>
+
+        <Box display="flex" flex="1" alignItems="flex-start" gap="8">
+          <BitkitNativeSelect
+            flex="1"
+            size="md"
+            value={strategy}
+            onChange={(e) => handleStrategyChange(e.target.value as VersionStrategy)}
+          >
+            {Object.entries(STRATEGY_LABELS)
+              .filter(([value]) => allowUnset || value !== 'unset')
+              .map(([value, label]) => (
+                <option key={value} value={value}>
+                  {label}
+                </option>
+              ))}
+          </BitkitNativeSelect>
+
+          {strategy !== 'unset' && (
+            <BitkitTextInput
+              size="md"
+              width="160px"
+              flexShrink="0"
+              placeholder={strategy === 'exact' ? 'e.g. 24.7.0' : 'prefix, e.g. 22'}
+              inputProps={{
+                value: version,
+                onChange: (e) => onVersionChange(e.target.value),
+              }}
+            />
+          )}
+        </Box>
+
+        <BitkitIconButton variant="tertiary" icon={IconMinusCircle} label="Remove tool" onClick={onRemove} />
       </Box>
 
-      <BitkitIconButton variant="tertiary" icon={IconMinusCircle} label="Remove tool" onClick={onRemove} />
+      {showCustomInput && (
+        <Text textStyle="body/md/regular">
+          The system is designed to support a growing list of tools and languages, but Bitrise only verifies and tests
+          the stability of the most common tools. If you need a tool not listed here, read{' '}
+          <BitkitLink
+            colorVariant="purple"
+            isExternal
+            suffixIcon={IconOpenInNew}
+            href="https://docs.bitrise.io/en/bitrise-ci/configure-builds/configuring-build-settings/configuring-tool-versions#supported-tools"
+          >
+            how to use community plugins
+          </BitkitLink>
+          .
+        </Text>
+      )}
     </Box>
   );
 };
--- a/source/javascripts/components/ToolVersions/ToolRow.tsx
+++ b/source/javascripts/components/ToolVersions/ToolRow.tsx
@@ -109,7 +109,7 @@ const ToolRow = ({
       <Box display="flex" alignItems="flex-start" gap="12">
         <Box display="flex" flexDirection="column" gap="8" width={rem(160)} flexShrink="0">
           <BitkitSelect
-            size="md"
+            size="lg"
             placeholder="Select one"
             isLoading={isCatalogLoading}
             items={dropdownItems}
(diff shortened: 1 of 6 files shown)


source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -59,47 +102,75 @@ const ToolRow = ({
...
-                {label}
-              </option>
-            ))}
-        </BitkitNativeSelect>
-
-        {strategy !== 'unset' && (
-          <BitkitTextInput
+    <Box display="flex" flexDirection="column" gap="8">
+      <Box display="flex" alignItems="flex-start" gap="12">
+        <Box display="flex" flexDirection="column" gap="8" width="160px" flexShrink="0">
```
k-olly: Use width={rem(160)} here please, imported from bitrise/bitkit-v2.

It's in the documents how to handle tokens properly, but sometimes Claude miss it.

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -59,47 +102,75 @@ const ToolRow = ({
...
+                <option key={value} value={value}>
+                  {label}
+                </option>
+              ))}
+          </BitkitNativeSelect>
+
+          {strategy !== 'unset' && (
+            <BitkitTextInput
+              size="md"
+              width="160px"
```
k-olly: Use width={rem(160)} here, rem is imported from bitrise/bitkit-v2

k-olly:
## Review findings

The overall structure follows the repo's api/service/hook layering well. The findings below survived an adversarial verify pass (8 finder angles → per-finding verification), most severe first.

### 1. `ToolsService.ts:64` — the branch doesn't typecheck: `ToolCatalog` is used but never imported **(CONFIRMED — reproduced with `npx tsc --noEmit`)**

The four new helpers reference `ToolCatalog` in their signatures without importing it from `../models/Tools`, producing `TS2304` at lines 64/69/78/89 plus cascading errors (`aliases` degrades to `never[]`, so `aliases.includes(toolId)` is a type error). Jest passes only because `@swc/jest` strips types without checking. Any typechecked build/CI will fail.

### 2. `ToolsService.ts:93` — alias-aware dedup is missing, so the same tool can be pinned twice **(CONFIRMED)**

With `tools: { go: '1.22' }` and catalog entry `{name: 'golang', aliases: ['go']}`, the "Add new" dropdown still offers `golang` — the filter `value === toolId || !existingToolIds.includes(value)` compares literal strings and never expands aliases against other rows' IDs. Picking it (or typing it into the "Other" input, since `validateToolId` is also literal-string-only) writes both `go` and `golang` into the YAML — two potentially conflicting pins for one tool, which the dropdown UI implies is impossible. Note that the dropdown path bypasses `validateToolId` entirely (it only runs on the text input's blur).

### 3. `ToolVersions.tsx:57-60` — changing the ID of an `unset` tool at root scope deletes the tool and throws **(CONFIRMED)**

`onIdChange` runs `deleteTool` then `setTool` with no try/catch, and they commit in separate store transactions. If root-level YAML manually contains `tools: { ruby: unset }` (`parseToolVersion` maps it to strategy `unset`), `setTool` throws `'Cannot use "unset" strategy at root scope'` *after* `deleteTool` already removed the entry — the tool silently disappears from the YAML and an uncaught exception escapes the event handler. (`allowUnset` only gates the strategy dropdown options, not what parses from existing YAML.)

### 4. `ToolRow.tsx:74` — no `newValue !== toolId` guard, causing a spurious YAML reorder **(CONFIRMED)**

The old blur handler kept an `if (newId !== toolId)` no-op guard; `handleDropdownChange` doesn't. Pick "Other" on an existing row, then re-pick the row's original tool from the dropdown: `onIdChange` fires with an unchanged ID, running `deleteTool` + `setTool`, which re-appends the entry at the end of the `tools` map — a spurious diff/dirty state with no real change, violating the YAML-preservation rule.

### 5. `useTools.ts:17` — a deterministic 404 on the catalog is retried 3× (~7s) before the fallback UI appears **(CONFIRMED)**

The app's `QueryClient` sets `retry: 3` globally (`main.tsx`), and `useToolCatalog` doesn't override it. If `catalog.json` 404s, every row's select shows a spinner through ~7s of exponential backoff before the "Couldn't load tool suggestions" alert and the custom-input fallback appear. Precedent: `useStep.ts:82` uses `retry: false`.

### 6. `ToolCatalogApi.ts:58` — `fetchJson` + `ToolCatalogError` re-implement the shared `client.ts` **(CONFIRMED)**

`Client.get` already supports `excludeCSRF: true`, per-call `timeout`, external `signal` chaining, and header overrides — `StepApi.getDirectGitStepByCvs` uses exactly this for external hosts. The file's CORS-preflight justification doesn't hold: overriding `Content-Type` to a CORS-safelisted value keeps the request simple. `ClientError` already carries status/cause/response, so `ToolCatalogError` is a second, structurally identical error taxonomy that shared error handling won't recognize.

### 7. `ToolCatalogApi.mswMocks.ts:13` — the mock hardcodes the catalog URL instead of importing a URL getter **(CONFIRMED)**

`ToolCatalogApi` has `getCatalogIndexUrl()`/`getToolCatalogUrl()` but keeps them private, so the mock re-declares the URL string. Existing mocks (`EnvVarsApi.mswMocks.ts:6`, `StacksAndMachinesApi.mswMocks.ts:537`) import the path builder from their API module so mock and client can't drift. If the base URL ever changes, MSW silently stops intercepting and stories/tests hit the real network.

### 8. `ToolRow.tsx:121` — `autoFocus` is dead for the pending row **(PLAUSIBLE)**

The old always-rendered text input received `autoFocus`; now the prop only reaches the custom text input, which is hidden until the user picks "Other", and the `BitkitSelect` never receives focus. Clicking "Add new" leaves focus on the now-disabled button.

---

**Also noted, but cut for lower value:** the `width="160px"` → `width={rem(160)}` token issue is already flagged inline; `useToolVersions`/`getToolVersions` ship with no consumer (intentional per the PR description — though note it fetches `${toolId}.json` with the raw ID, so an aliased row like `go` would request `go.json` instead of `golang.json`; worth designing the canonical-name resolution before the follow-up PR); minor memoization opportunities; and the derivable `isCatalogReady`/`isToolIdKnown` props.

The most urgent one is #1 (the branch doesn't compile under `tsc`); #2 and #3 are the real behavioral bugs.

AndrasEszes:
## Review — Tool ID dropdown (`BE-2126`)

The api/service/hook layering is clean, unit coverage on `ToolsService` is solid, and `tsc --noEmit` is green on the branch. Findings below, most severe first — line refs against HEAD (`2bc7a54a`).

> Already addressed since the earlier review: the missing `ToolCatalog` import (compiles now) and `rem` on the version input (`ToolRow.tsx:146`).

### 🐛 Correctness — fix before merge

**1. Alias-aware dedup missing → the same tool can be pinned twice** — `ToolsService.ts:93`
`getAvailableToolIdOptions` filters with literal string compare (`value === toolId || !existingToolIds.includes(value)`), so with a row already using `go`, the dropdown still offers the canonical `golang`. `validateToolId` is literal-only too, so the "Other" input doesn't catch it either. Result: both `go` and `golang` land in the YAML — two conflicting pins for one tool, which the dropdown UI implies is impossible. Needs a canonical-name resolver used on both the dedup filter and `validateToolId`. The existing spec (`ToolsService.spec.ts:127`) only puts the *current* row's alias in `existingToolIds`, never a *different* row's — so it doesn't cover this; please add that case.

**2. Changing the ID of an `unset` tool at root scope deletes it and throws** — `ToolVersions.tsx:57-60`
`onIdChange` runs `deleteTool` then `setTool` in two separate store transactions, no try/catch. If root-level YAML contains `tools: { ruby: unset }` (valid to parse; `allowUnset` only gates the dropdown, not what's parsed from existing YAML), `setTool` throws `'Cannot use "unset" strategy at root scope'` *after* `deleteTool` already committed — the tool silently disappears and an uncaught exception escapes the handler.

**3. No no-op guard in `handleDropdownChange` → spurious YAML reorder** — `ToolRow.tsx:75`
Pick "Other" on an existing row, then re-pick the row's original tool: `onIdChange` fires with an unchanged ID → `deleteTool` + `setTool` re-appends the entry at the end of the `tools` map. Dirty diff / reorder with no real change, against the YAML-preservation rule. The old blur handler had an `if (newId !== toolId)` guard; the dropdown path lost it.

**4. Typing a known ID equal to the current one doesn't switch back to the dropdown** — `ToolRow.tsx:93`
`handleIdBlur` resets `manualOther` only when `newId !== toolId`. Editing a row with `toolId="golang"`, picking "Other", and typing `golang` again leaves the custom input visible even though the tool is known — contrary to PR decision #3. Reset `manualOther` on every successful validation and let `showCustomInput`'s `isToolIdKnown` branch decide.

### ⚠️ Quality

**5. Deterministic catalog 404/500 is retried 3× (~7s) before the fallback shows** — `useTools.ts:17`
The global `QueryClient` sets `retry: 3` (`main.tsx:80`) and `useToolCatalog` doesn't override it, so a missing `catalog.json` spins every row's select through ~7s of backoff before the "Couldn't load tool suggestions" alert + custom-input fallback appear. Add `retry: false` (precedent: `useStep.ts:82`).

**6. `ToolCatalogApi` re-implements `client.ts`** — `ToolCatalogApi.ts:58`
`fetchJson` + `ToolCatalogError` duplicate what `client.ts` already does (timeout, external-signal chaining, status/cause on `ClientError`). The CORS-preflight justification doesn't require a bespoke fetch: `Client.get(url, { excludeCSRF: true, headers: { 'Content-Type': 'text/plain' } })` stays a CORS-simple request (`text/plain` and `Accept` are safelisted). The parallel `ToolCatalogError` taxonomy is something shared error handling won't recognize. Not blocking, but worth collapsing onto `Client`.

**7. MSW mock hardcodes the catalog URL** — `ToolCatalogApi.mswMocks.ts:13`
`getCatalogIndexUrl()`/`getToolCatalogUrl()` are private, so the mock re-declares the URL string; if the base URL changes, MSW silently stops intercepting and stories/tests hit the real network. Export the URL builders and import them in the mock (precedent: `EnvVarsApi.mswMocks.ts`, `StacksAndMachinesApi.mswMocks.ts`).

### 🎨 Layout & Bitkit / AGENTS.md

**8. Uneven horizontal gaps within a row + one redundant `<div>`** — `ToolRow.tsx:107,127`
The row uses `gap="12"` but the strategy+version sub-group uses `gap="8"`, so left-Select↔strategy (12) is visibly wider than strategy↔version (8) — almost certainly unintentional. Flatten the sub-group into the row with a single gap: this makes spacing uniform **and** removes one wrapper `<div>` per row.
```tsx
<Box display="flex" alignItems="flex-start" gap="12">
  <Box ... width={rem(160)} flexShrink="0">{/* tool-ID: select + custom input */}</Box>
  <BitkitNativeSelect flex="1" size="lg" .../>
  {strategy !== 'unset' && <BitkitTextInput width={rem(160)} flexShrink="0" size="lg" .../>}
  <BitkitIconButton .../>
</Box>
```

**9. Input size inconsistency on the Stacks & Machines card** — `ToolRow.tsx`
The Stack/Machine selectors are legacy `@bitrise/bitkit` (v1) `Dropdown`s (taller); the new Tool setup inputs are bitkit-v2 at `size="md"`, so they don't line up on the same card. Bump the three Tool setup v2 inputs to `size="lg"` to match the v1 dropdown height. Leave `BitkitIconButton` at its default `lg` — it already matches `lg` inputs, no `size` prop needed. (Verify `BitkitNativeSelect` accepts `size="lg"`.) Long term, migrating `StackSelector`/`MachineTypeSelector` to bitkit-v2 is a separate ticket.

**10. Arbitrary px instead of a token** — `ToolRow.tsx:108`
`width="160px"` on the tool-ID column Box → use `width={rem(160)}` (consistent with line 146). AGENTS.md: "avoid arbitrary px."

**11. Numeric size instead of a string** — `ToolVersions.tsx:31`
`maxWidth={640}` → AGENTS.md wants sizes as strings; use `maxWidth={rem(640)}`.

### Minor

- **`autoFocus` is dead on the pending row** (`ToolRow.tsx:122`) — the prop only reaches the hidden custom text input; clicking "Add new" leaves focus on the (now-disabled) button, not the tool-ID select.
- **Dead code (intentional per PR desc, for the follow-up)** — `useToolVersions` / `ToolCatalogApi.getToolVersions` ship with no consumer. Note it fetches `${toolId}.json` with the raw ID, so an aliased row (`go`) would request `go.json` instead of `golang.json` — design the canonical-name resolution before wiring version selection.

### Nice-to-have

- Per AGENTS.md, the repeated `<Box display="flex" flexDirection="column" gap>` / `...row...` could be `VStack` / `HStack` — same DOM, less prop noise. Optional.

source/javascripts/components/ToolVersions/ToolRow.tsx
ofalvai: Fixed in 2bc7a54a

source/javascripts/components/ToolVersions/ToolRow.tsx
ofalvai: Fixed in 2bc7a54a

ofalvai:
@k-olly regarding this:


> 3. ToolVersions.tsx:57-60 — changing the ID of an unset tool at root scope deletes the tool and throws (CONFIRMED)
onIdChange runs deleteTool then setTool with no try/catch, and they commit in separate store transactions. If root-level YAML manually contains tools: { ruby: unset } (parseToolVersion maps it to strategy unset), setTool throws 'Cannot use "unset" strategy at root scope' after deleteTool already removed the entry — the tool silently disappears from the YAML and an uncaught exception escapes the event handler. (allowUnset only gates the strategy dropdown options, not what parses from existing YAML.)


`unset` is invalid in the root scope, it can only appear in the workflow scope. It cannot be set on the UI, and the raw `bitrise.yml` is validated anyway, so I think this risk is not that practical.

ofalvai:
@k-olly, @AndrasEszes:

> 1. Alias-aware dedup missing → the same tool can be pinned twice — ToolsService.ts:93
getAvailableToolIdOptions filters with literal string compare (value === toolId || !existingToolIds.includes(value)), so with a row already using go, the dropdown still offers the canonical golang. validateToolId is literal-only too, so the "Other" input doesn't catch it either. Result: both go and golang land in the YAML — two conflicting pins for one tool, which the dropdown UI implies is impossible. Needs a canonical-name resolver used on both the dedup filter and validateToolId. The existing spec (ToolsService.spec.ts:127) only puts the current row's alias in existingToolIds, never a different row's — so it doesn't cover this; please add that case.

Fixed here: https://github.com/bitrise-io/bitrise-workflow-editor/pull/1831/commits/9bcfe4ce0098066ab60f1ccdf2c4f83d3e4967de

ofalvai:
@k-olly, @AndrasEszes:

> 4. ToolRow.tsx:74 — no newValue !== toolId guard, causing a spurious YAML reorder (CONFIRMED)
The old blur handler kept an if (newId !== toolId) no-op guard; handleDropdownChange doesn't. Pick "Other" on an existing row, then re-pick the row's original tool from the dropdown: onIdChange fires with an unchanged ID, running deleteTool + setTool, which re-appends the entry at the end of the tools map — a spurious diff/dirty state with no real change, violating the YAML-preservation rule.

Fixed here: https://github.com/bitrise-io/bitrise-workflow-editor/pull/1831/commits/3e792307fd25637254886d6f9484ec394c7840c8

zoltanszabo-bitrise:
Is this the intended UI behaviour (Input under the selector in case of Other)? It seems a bit odd to me.
image

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -59,47 +107,75 @@ const ToolRow = ({
...
+              size="md"
+              placeholder="Tool ID (e.g. deno)"
+              errorText={idError}
+              inputProps={{ defaultValue: toolId, autoFocus, onBlur: handleIdBlur }}
+            />
+          )}
+        </Box>
+
+        <Box display="flex" flex="1" alignItems="flex-start" gap="8">
+          <BitkitNativeSelect
```
zoltanszabo-bitrise: Why do we mix native select with a styled select, why not use the styled select in both places?

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -1,22 +1,25 @@
...
 const ToolVersions = ({ workflowId }: { workflowId?: string }) => {
   const scope: ToolScope = workflowId ? { type: 'workflow', workflowId } : { type: 'root' };
   const tools = useToolsForScope(scope);
+  const { data: catalog, isLoading: isCatalogLoading, isError: isCatalogError } = useToolCatalog();
   const [hasPendingRow, setHasPendingRow] = useState(false);
   const [pendingStrategy, setPendingStrategy] = useState<VersionStrategy>('latest-released');
   const [pendingVersion, setPendingVersion] = useState('');
 
   const allowUnset = scope.type === 'workflow';
+  const isCatalogReady = !isCatalogLoading && catalog !== undefined;
```
zoltanszabo-bitrise: We could use `!!catalog` since catalog will be only defined if successfully loaded. Also the useQuery hook  exposes an `isSuccess` flag too, which could be used. But I would keep it simple an assert whether the catalog is defined or not. (so you dont have to deal with background refetch, etc. - which you might also want to disable in the useToolCatalog hook).

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -45,8 +48,13 @@ const ToolVersions = ({ workflowId }: { workflowId?: string }) => {
               toolId={toolId}
               strategy={parsed.strategy}
               version={versionValue}
-              existingToolIds={Object.keys(tools)}
+              existingToolIds={existingToolIds}
+              catalog={catalog}
               allowUnset={allowUnset}
+              dropdownOptions={ToolsService.getAvailableToolIdOptions(catalog, toolId, existingToolIds)}
+              isToolIdKnown={ToolsService.isKnownToolId(catalog, toolId)}
```
zoltanszabo-bitrise: Do we really need to calculate and pass these here, the component already gets every information so `ToolRow` could hide these as implementation details.

source/javascripts/components/ToolVersions/ToolVersions.stories.tsx
```diff
@@ -3,6 +3,7 @@ import { Meta, StoryObj } from '@storybook/react-vite';
 import { set } from 'es-toolkit/compat';
 import { stringify } from 'yaml';
 
+import ToolCatalogApiMocks from '@/core/api/ToolCatalogApi.mswMocks';
```
zoltanszabo-bitrise, line 6: Here is the proof for having an mswMock, since you need it for storybooks, you can reuse it in tests. (Actually if you want to, you also write integration-like tests against storybook use-case, but it is more of a pattern in the Monolith than in the WFE).

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -79,6 +92,8 @@ const ToolVersions = ({ workflowId }: { workflowId?: string }) => {
         )}
       </Box>
 
+      {isCatalogError && <BitkitAlert variant="warning" messageText="Couldn't load tool suggestions." />}
```
zoltanszabo-bitrise, line 88: Since it is just checking the isError flag of the query, the actual error message never surfaces from the ToolCatalogApi to the user. It worth reconsidering showing the returned error from the API.

source/javascripts/components/ToolVersions/ToolRow.tsx
ofalvai: Good point, it's some old code but fixed it in https://github.com/bitrise-io/bitrise-workflow-editor/pull/1831/changes/85f0bd41186efc55cb67aa4c6bf6060d71398957, thank you!

source/javascripts/components/ToolVersions/ToolVersions.tsx
ofalvai: Is there a kind of error that is "expected" and the user should know about? When writing this, my perspective was that these are all unexpected errors and there is nothing the user can do (other than refresh and pray :D)

source/javascripts/hooks/useTools.ts
```diff
@@ -10,3 +13,22 @@ export function useToolsForScope(scope: ToolScope): Tools {
...
+export function useToolCatalog() {
+  return useQuery({
+    queryKey: ['tool-catalog'],
+    queryFn: () => ToolCatalogApi.getToolCatalog(),
+    staleTime: 30 * 60 * 1000,
+    gcTime: 30 * 60 * 1000,
+  });
+}
+
+export function useToolVersions(toolId: string, enabled: boolean) {
```
zoltanszabo-bitrise, line 26: Where is this used? Are aliases resolved to tool ids before passed to this hook? Or both `go` and `golang` would be served just fine by the API?

ofalvai: Coming in the next PR! :) I can remove if needed, the reasoning was that it's just a small utility and looks exactly like the other one.

source/javascripts/components/ToolVersions/ToolVersions.tsx
ofalvai: Thank you for spotting this, fully agree! This was in place before I added the `catalog` prop, missed how that makes `isToolIdKnown` redundant.

ofalvai:
@AndrasEszes addressing this in #1834 (this PR is just stack on top of that)

> 5. useTools.ts:17 — a deterministic 404 on the catalog is retried 3× (~7s) before the fallback UI appears (CONFIRMED)
The app's QueryClient sets retry: 3 globally (main.tsx), and useToolCatalog doesn't override it. If catalog.json 404s, every row's select shows a spinner through ~7s of exponential backoff before the "Couldn't load tool suggestions" alert and the custom-input fallback appear. Precedent: useStep.ts:82 uses retry: false.

source/javascripts/components/ToolVersions/ToolVersions.tsx
ofalvai: Fixed in https://github.com/bitrise-io/bitrise-workflow-editor/pull/1831/changes/33ac3042a967a89c415dbc8f80172811f56fd2a2

source/javascripts/components/ToolVersions/ToolVersions.tsx
ofalvai: Fixed in https://github.com/bitrise-io/bitrise-workflow-editor/pull/1831/changes/33ac3042a967a89c415dbc8f80172811f56fd2a2

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -25,7 +28,7 @@ const ToolVersions = ({ workflowId }: { workflowId?: string }) => {
   };
 
   return (
-    <Box display="flex" flexDirection="column" gap="24" marginBlockStart="24" maxWidth={640}>
+    <Box display="flex" flexDirection="column" gap="24" marginBlockStart="24" maxWidth={rem(640)}>
       <Box display="flex" flexDirection="column" gap="8">
         <Text textStyle="heading/h3">Tool setup</Text>
         <Text textStyle="body/md/regular" color="text/secondary">
```
AndrasEszes, line 33: The `Learn more` link just below renders grey instead of purple. With no `colorVariant`, `BitkitLink` inherits `currentColor`, and here that resolves to this Text's `text/secondary`. Add `colorVariant="purple"` for the branded style. While here: the `href="#"` is a dead placeholder, so point it at the real docs and add `isExternal suffixIcon={IconOpenInNew}` to match the community-plugins link in ToolRow.

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -1,22 +1,25 @@
...
 const ToolVersions = ({ workflowId }: { workflowId?: string }) => {
   const scope: ToolScope = workflowId ? { type: 'workflow', workflowId } : { type: 'root' };
   const tools = useToolsForScope(scope);
+  const { data: catalog, isLoading: isCatalogLoading, isError: isCatalogError } = useToolCatalog();
   const [hasPendingRow, setHasPendingRow] = useState(false);
   const [pendingStrategy, setPendingStrategy] = useState<VersionStrategy>('latest-released');
   const [pendingVersion, setPendingVersion] = useState('');
 
   const allowUnset = scope.type === 'workflow';
+  const isCatalogReady = !!catalog;
```
AndrasEszes: Small simplification: `isCatalogReady` is just `!!catalog`, and `ToolRow` already receives `catalog`, so it can derive this itself. Dropping the prop removes one thing to keep in sync. (`isCatalogLoading` should stay, since loading vs error both yield an undefined catalog.)

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -59,36 +105,42 @@ const ToolRow = ({
...
-            .map(([value, label]) => (
-              <option key={value} value={value}>
-                {label}
-              </option>
-            ))}
-        </BitkitNativeSelect>
+            .map(([value, label]) => ({ value, label }))}
+          value={strategy}
+          onValueChange={(v) => handleStrategyChange(v as VersionStrategy)}
+        ></BitkitSelect>
```
AndrasEszes: Nit: no children here, so this can self-close: `<BitkitSelect ... />`.

AndrasEszes (commented):
A few small non-blocking suggestions inline.

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -45,8 +48,11 @@ const ToolVersions = ({ workflowId }: { workflowId?: string }) => {
...
               version={versionValue}
-              existingToolIds={Object.keys(tools)}
+              existingToolIds={existingToolIds}
+              catalog={catalog}
               allowUnset={allowUnset}
+              isCatalogReady={isCatalogReady}
+              isCatalogLoading={isCatalogLoading}
               onIdChange={(newId) => {
                 ToolsService.deleteTool(toolId, scope);
                 ToolsService.setTool(newId, parsed.strategy, versionValue, scope);
```
AndrasEszes: Renaming a tool via delete + re-add makes `setTool` append `newId` at the end of the `tools` map, so renaming a non-last tool reorders the YAML and produces a phantom diff on untouched sibling tools (plus it runs as two store transactions). `YmlUtils.updateKeyByPath` renames the key in place, keeping both its position and the existing value node's style. Worth wrapping it in a `ToolsService.renameTool(oldId, newId, scope)` and calling that here instead. It also removes the delete-then-throw fragility.

AndrasEszes (commented):
Phantom YAML change on tool rename.

source/javascripts/components/ToolVersions/ToolVersions.tsx
ofalvai: Thank you, applied the styling suggestions. The `href` is still a placeholder, we'll do a final round of copy review in a follow-up PR (the whole view is behind a feature flag for now)

source/javascripts/components/ToolVersions/ToolVersions.tsx
ofalvai: Thank you for spotting this, fixed

source/javascripts/components/ToolVersions/ToolRow.tsx
ofalvai: Whoops, fixed (force-pushed original commit)

source/javascripts/components/ToolVersions/ToolVersions.tsx
ofalvai: Makes sense, thank you for the explanation. Added `renameTool()`

AndrasEszes (approved):

AndrasEszes (approved):

PR #1834, opened 2026-07-02
Tool version catalog fetching

ofalvai:
Adds initial data source for tool catalog. Exposes via React hooks for future consumption in the UI. The UI work can also be viewed [here](https://github.com/bitrise-io/bitrise-workflow-editor/pull/1831), stacked on top of this PR

API details:

It's a dead simple JSON API with two endpoints (for now):
1. https://bitrise.io/stacks/tools/v1/catalog.json
2. Versions of a tool, e.g. https://bitrise.io/stacks/tools/v1/nodejs.json


benbitrise:
@k-olly - I saw your LGTM emoji. Could you give this an approval so we can merge?

source/javascripts/core/api/ToolCatalogApi.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+
+function jsonResponse(body: unknown): Response {
+  return { ok: true, status: 200, json: () => Promise.resolve(body) } as unknown as Response;
+}
+
+function errorResponse(status: number): Response {
+  return { ok: false, status, json: () => Promise.reject(new Error('no body')) } as unknown as Response;
+}
+
+const fetchMock = jest.fn<Promise<Response>, [RequestInfo | URL, RequestInit?]>();
```
zoltanszabo-bitrise: Don't we use MSW mocks in the WFE also?

source/javascripts/core/api/ToolCatalogApi.ts
```diff
@@ -0,0 +1,120 @@
...
+
+function getToolCatalogUrl(toolId: string): string {
+  return `${TOOL_CATALOG_BASE_URL}/${encodeURIComponent(toolId)}.json`;
+}
+
+/**
+ * Fetch and parse a JSON document. Throws a `ToolCatalogError` on any failure —
+ * network error, non-2xx status, or malformed JSON.
+ */
+async function fetchJson<T>(url: string, signal?: AbortSignal): Promise<T> {
```
zoltanszabo-bitrise: Why not reuse the existing fetch client?

source/javascripts/core/api/ToolCatalogApi.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+}
+
+/** Fetch the catalog index. Throws a `ToolCatalogError` when unavailable or malformed. */
+async function getToolCatalog(signal?: AbortSignal): Promise<ToolCatalog> {
+  const url = getCatalogIndexUrl();
+  const dto = await fetchJson<ToolCatalogIndexDto>(url, signal);
+  if (!dto || !Array.isArray(dto.tools)) {
+    throw new ToolCatalogError('Catalog index is missing a "tools" array', url);
+  }
```
zoltanszabo-bitrise, line 48: Isn't it too defensive (missing tools array)? We are in control of the API we are calling, aren't we?

source/javascripts/core/api/ToolCatalogApi.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  return { tools: dto.tools };
+}
+
+/** Fetch a single tool's version list. Throws a `ToolCatalogError` when unavailable or malformed. */
+async function getToolVersions(toolId: string, signal?: AbortSignal): Promise<ToolVersions> {
+  const url = getToolCatalogUrl(toolId);
+  const dto = await fetchJson<ToolVersionsDto>(url, signal);
+  if (!dto || !Array.isArray(dto.versions)) {
+    throw new ToolCatalogError(`Version list for "${toolId}" is missing a "versions" array`, url);
```
zoltanszabo-bitrise: Same as above, if we control the API we might not need to be thid defensive about a missing versions array.

source/javascripts/core/api/ToolCatalogApi.ts
ofalvai: The API is more like a hosted JSON in this case. We have [some schema validation](https://github.com/bitrise-io/bitrise/blob/master/bitrise.yml#L58-L86) though.

But if you say it's too defensive and it's not how things are handled in this codebase, I'm happy to remove.

AndrasEszes:
Nice, well-scoped PR. The API layer is clean, error handling is thorough, and the test coverage is genuinely good (network, non-2xx, malformed JSON, missing array, mixed and empty version lists all covered). A few suggestions below, mostly small, plus my take on the open review threads.

**Heads up on stale bot comments:** the two AI-reviewer comments asking to thread `AbortSignal` through `ToolsService.getToolCatalog`/`getToolVersions` are no longer relevant, since those wrappers were removed in favor of calling the API directly from the hooks. Safe to resolve them.

**1. Forward React Query's `AbortSignal` from the hooks (`useTools.ts`).** The API threads `signal` all the way through (with the timeout plus external-abort handling, which is great), but the hooks call `queryFn: () => ToolCatalogApi.getToolCatalog()`, which drops the `{ signal }` that React Query passes in. As a result the in-flight request is never cancelled on unmount or navigation. Small change to make the plumbing actually used:
```ts
queryFn: ({ signal }) => ToolCatalogApi.getToolCatalog(signal),
// and: queryFn: ({ signal }) => ToolCatalogApi.getToolVersions(toolId, signal),
```

**2. Consider `retry: false` on `useToolCatalog` (`useTools.ts:17`).** The global `QueryClient` defaults to `retry: 3` (`main.tsx:80`). Since the catalog is a static JSON asset, a deterministic 404/500 currently retries three times (~7s of backoff) before consumers can fall back. The UI impact shows up in the stacked PR, but the hook is defined here, so this feels like the right place to opt out. Precedent: `useStep.ts:82`.

**3. Reusing the shared `client.ts` (`ToolCatalogApi.ts:58`), following up on @zoltanszabo-bitrise's question.** The existing client already covers timeout, external-signal chaining, and a `ClientError` that carries status and cause. The preflight concern is still satisfiable through it: `Client.get(url, { excludeCSRF: true, headers: { 'Content-Type': 'text/plain' } })` keeps the request CORS-simple (both `text/plain` and `Accept` are safelisted). Reusing it would also mean the app's shared error handling recognizes these failures, rather than a separate `ToolCatalogError` type. Not a blocker, but probably worth aligning on before this lands, since it's the main design choice in the PR.

**4. MSW for the spec, following up on @zoltanszabo-bitrise's question (`ToolCatalogApi.spec.ts`).** The rest of the codebase mocks at the network layer with MSW (`.mswMocks.ts`) rather than replacing `global.fetch`. Two small benefits: it matches the existing convention, and it avoids the un-restored `global.fetch` (it's reassigned in `beforeEach` but never restored). The stacked PR already introduces a `ToolCatalogApi.mswMocks.ts`, so the spec could lean on that and keep the mock in one place.

**5. On the defensiveness of the missing-array checks (`ToolCatalogApi.ts:97`, `:108`), re @zoltanszabo-bitrise.** My vote is to keep this one guard. Even though we control the endpoint, it's a hosted static JSON on a CDN, so a misroute or 404 can return HTML (the `.json()` failure path already handles that) or, less often, a JSON object without the expected array. The check is cheap, gives a clear error, and the fallback UI relies on it throwing. Totally reasonable to drop it if the team prefers leaner code though, it isn't a correctness issue either way.

**6. Alias resolution for `getToolVersions` (`ToolCatalogApi.ts:105`), for the follow-up.** It fetches `${toolId}.json` with the raw ID, so an aliased entry (e.g. `go`) would request `go.json` instead of the canonical `golang.json`. No consumer yet, so nothing to change here, but worth designing the canonical-name resolution before wiring version selection.

**7. Minor: `isSemver` semantics (`ToolCatalogApi.ts:112`).** `semver.valid('6.3')` returns null, so partial versions like `6.3` or `22` are flagged `isSemver: false` (the spec captures this well). That reads as "is a complete valid semver" rather than "is a usable version selector". A short doc comment would help future readers not mistake it for "parseable".

Thanks for this. Items 1 and 2 are one-liners; 3 and 4 are the design calls that would be good to align on.

source/javascripts/hooks/useTools.ts
```diff
@@ -10,3 +13,22 @@ export function useToolsForScope(scope: ToolScope): Tools {
     return yml.tools ?? {};
   });
 }
+
+export function useToolCatalog() {
+  return useQuery({
+    queryKey: ['tool-catalog'],
+    queryFn: () => ToolCatalogApi.getToolCatalog(),
```
zoltanszabo-bitrise: React Query passes a `QueryFunctionContext` here with an `AbortSignal` that it aborts on unmount / when the query is superseded. `getToolCatalog` already accepts a `signal`, so this can be:

```ts
queryFn: ({ signal }) => ToolCatalogApi.getToolCatalog(signal),
```

As-is, the `signal` composition in `fetchJson` is effectively dead code until it's wired through here.

source/javascripts/hooks/useTools.ts
```diff
@@ -10,3 +13,22 @@ export function useToolsForScope(scope: ToolScope): Tools {
...
+    staleTime: 30 * 60 * 1000,
+    gcTime: 30 * 60 * 1000,
+  });
+}
+
+export function useToolVersions(toolId: string, enabled: boolean) {
+  return useQuery({
+    queryKey: ['tool-versions', toolId],
+    queryFn: () => ToolCatalogApi.getToolVersions(toolId),
```
zoltanszabo-bitrise: Same here — and this one benefits most: changing `toolId` supersedes the previous query, so a wired signal cancels the stale in-flight fetch instead of letting it resolve and get discarded:

```ts
queryFn: ({ signal }) => ToolCatalogApi.getToolVersions(toolId, signal),
```

source/javascripts/core/api/ToolCatalogApi.ts
zoltanszabo-bitrise: `client.ts` already implements this timeout + AbortController composition + error mapping — and it distinguishes `AbortError` from `Failed to fetch`, which this wrapper collapses into a single message. Worth considering reusing it instead of a parallel `fetchJson` + `ToolCatalogError`.

The comment's reasoning is sound — CSRF and a request `Content-Type: application/json` would force a cross-origin preflight — but both are addressable in the shared client. `excludeCSRF: true` already exists, and the request `Content-Type` is the only real blocker: it's meaningless on a bodyless GET and the sole non-safelisted header. `Accept: application/json` stays, so responses are still JSON.

One clean way to generalize (rather than adding more `exclude*` booleans): a `defaultOpts` that caller opts merge over, plus a `null`-sentinel so a caller can *remove* a default header:

```ts
Client.get(url, { headers: { 'Content-Type': null, 'X-CSRF-TOKEN': null } });
```

That drops ~50 lines of duplicated plumbing here and lets any caller opt into a CORS-simple request uniformly. Since it reshapes the shared client, totally fine as a follow-up if you'd rather keep this PR self-contained.

zoltanszabo-bitrise (commented):
## Review summary

Nicely scoped data-source layer — clear `core/` separation, thorough failure-path test coverage, and a genuinely useful header comment documenting *why* it bypasses `client.ts`. CI is green. A few refinements below; none are blockers.

**Suggested changes**
1. **Wire React Query's `signal` through the hooks.** The API already accepts an `AbortSignal` and `fetchJson` builds the composition for it, but the hooks drop it, so that plumbing is currently dead code. (inline)
2. **Consider reusing `client.ts`** instead of the parallel `fetchJson` + `ToolCatalogError`. The header comment's reasoning is sound, but both concerns are addressable in the shared client. Since it reshapes `client.ts`, this is reasonable as a follow-up rather than a merge blocker if you'd rather keep this PR tight. (inline)
3. **Prefer MSW for the spec** — matches repo convention, and the stacked UI PR #1831 will likely need a `.mswMocks.ts` anyway. (inline)

**Considered but not blocking / out of scope**
- *Element-level DTO validation* (e.g. checking each `versions[]` entry is actually a string): the type-vs-runtime gap on untrusted JSON is real but an acceptable pragmatic trade-off here.
- *Pre-aborted external signal*: if `options.signal` is already aborted at call time, the `{ once: true }` abort listener never fires and the fetch still proceeds. Note `client.ts` has the **same** pattern, so this is pre-existing in the shared client, not introduced here. If you care, it's a one-line fix in `client.ts` (`if (options?.signal?.aborted) controller.abort()`) as a separate tiny PR benefiting all callers.

Suggested verdict: **Approve with comments.**

source/javascripts/core/api/ToolCatalogApi.spec.ts
zoltanszabo-bitrise: Repo convention is to mock at the HTTP layer with MSW (`*.mswMocks.ts`) rather than replacing `global.fetch`. MSW would avoid the `as unknown as Response` casts, auto-restore between tests (this mock is never restored), and let you assert the real URL/headers and abort behavior. The stacked UI PR #1831 will likely need a `ToolCatalogApi.mswMocks.ts` for stories anyway, so adding it here would be reusable.

zoltanszabo-bitrise (changes requested):

benbitrise:
@zoltanszabo-bitrise @AndrasEszes - thank you for your reviews! The PR is updated with improvements based on your feedback. Summary:

- React Query's AbortSignal is now passed into API calls
- Retries are disabled. Makes sense for something which is fetching static files.
- Nuked the raw fetch() use and now use client.ts (with appropriate inputs to avoid preflight request). This also simplifies the testing (we can just mock the client).
- I did not add MSW, since, from what I could see, that is not used for testing (mocked client.ts for that). Instead it is used for storybook. Oliver added MSW for that purpose in his UI PR.
- I left in the defensiveness on response type checking. As mentioned in commentary, it's not a real API, and overall it felt relatively lightweight to keep it in there.
- Bandi, your comment about resolving versions for aliases will come in a future PR

It is ready for another round of review by you.

AndrasEszes (dismissed):

zoltanszabo-bitrise:
> I did not add MSW, since, from what I could see, that is not used for testing

That's sensible for now, maybe worth keeping it as a follow-up item, so we use the same network mocking everywhere later on.

source/javascripts/core/api/ToolCatalogApi.ts
```diff
@@ -0,0 +1,107 @@
...
+      signal,
+      timeout: REQUEST_TIMEOUT_MS,
+      excludeCSRF: true,
+      headers: { 'Content-Type': 'text/plain' },
+    });
+  } catch (error) {
+    const status = error instanceof ClientError ? error.status : undefined;
+    const message = status ? `Request to ${url} failed with status ${status}` : `Failed to fetch ${url}`;
+    throw new ToolCatalogError(message, url, { status, cause: error });
+  }
```
zoltanszabo-bitrise: Can't we just rely on the ClientError altogether, do we really need a separate ToolCatalogError?

source/javascripts/hooks/useTools.ts
```diff
@@ -10,3 +13,26 @@ export function useToolsForScope(scope: ToolScope): Tools {
...
 }
+
+export function useToolCatalog() {
+  return useQuery({
+    queryKey: ['tool-catalog'],
+    queryFn: ({ signal }) => ToolCatalogApi.getToolCatalog(signal),
+    staleTime: 30 * 60 * 1000,
+    gcTime: 30 * 60 * 1000,
+    // As of July 2026, this fetches a static asset, so there is no benefit to retries
+    retry: false,
```
zoltanszabo-bitrise: I mean intermittent network issue can happen what retry could solve, I think what we dont need is background or window focus refetch, but probably no casing any issue if we keep them enabled.
https://tanstack.com/query/v5/docs/framework/react/reference/useQuery

benbitrise:
@zoltanszabo-bitrise thank you for the clarification on retries. I also went ahead and took the feedback on the custom error type. Results in a much slimmer PR. Ready for follow-up review.

zoltanszabo-bitrise (approved):
LGTM! 👍🏻

PR #1842, opened 2026-07-13
Exact tool version dropdown

benbitrise:
When a tool row's strategy is Exact, the free-text input becomes a filterable dropdown of concrete versions from the tool catalog. Free text remains for prefix strategies and custom tools.

- If the value (already set in yml) in not in the catalog, that value kept, but a warning is displayed
- If the version fetch fails, the field becomes read-only (keeping the current value visible) and an alert links to editing the version manually in the YAML tab
- If the user moves focus away from an empty exact input, an error message is shown
- Renaming a tool to a different one clears its version/prefix, since a value picked for the old tool is very unlikely to be valid for the new one. The strategy (exact/latest/installed/unset) is carried over though.

NOTE: There is no input validation when clicking save.

### DIFF
--- a/source/javascripts/components/ToolVersions/ToolRow.tsx
+++ b/source/javascripts/components/ToolVersions/ToolRow.tsx
@@ -193,13 +193,6 @@ const ToolRow = ({
                 applies whenever one is possible at all. */}
             {isExactKnownTool ? (
               <BitkitCombobox
-                // BitkitCombobox snapshots `items` on mount and never resets its own filtered
-                // list. Picking an item narrows that list, and it stays narrowed even after
-                // the popover closes and reopens. The key forces a fresh mount whenever the
-                // list should actually change: another tool's versions, the async load
-                // completing, or the version itself changing (without these, the dropdown
-                // would reopen showing only the previously picked item).
-                key={`${canonicalToolId}:${toolVersions ? 'loaded' : 'pending'}:${version}`}
                 size="lg"
                 placeholder="Select"
                 emptyLabel="No matches"
--- a/source/javascripts/components/ToolVersions/ToolRow.tsx
+++ b/source/javascripts/components/ToolVersions/ToolRow.tsx
@@ -132,9 +132,19 @@ const ToolRow = ({
   // (bare `latest`/`installed`). Only a hint — saving is not blocked.
   const versionError = strategy === 'exact' && version.trim() === '' ? 'Tool version is required' : undefined;
   const displayedVersionError = versionTouched ? versionError : undefined;
-  const displayedVersionWarning = isVersionMissingFromCatalog
+  // Shown next to a value the combobox is displaying as already set, where it may be a
+  // stale or mistaken leftover. Typed values get a different, less alarming message below.
+  const catalogMismatchWarning = isVersionMissingFromCatalog
     ? `${version} isn't in the list of installable versions`
     : undefined;
+  // Shown wherever the user is actually typing a version by hand (via "Other", or because
+  // no suggestions exist at all), where it is an expected, deliberate choice rather than
+  // a possible mistake. Skipped for a custom tool, whose own explanatory paragraph below
+  // already covers this.
+  const customVersionWarning =
+    !showCustomInput && isVersionMissingFromCatalog
+      ? `${version} is a custom version and might not work properly`
+      : undefined;
 
   const dropdownItems = [
     ...dropdownOptions,
@@ -232,7 +242,7 @@ const ToolRow = ({
                   onOpenChange: (details) => !details.open && !manualVersion && setVersionTouched(true),
                 }}
                 errorText={manualVersion ? undefined : displayedVersionError}
-                warningText={manualVersion ? undefined : displayedVersionWarning}
+                warningText={manualVersion ? undefined : catalogMismatchWarning}
                 value={manualVersion ? OTHER_VERSION_VALUE : version || undefined}
                 onValueChange={(newVersion) => {
                   if (newVersion === OTHER_VERSION_VALUE) {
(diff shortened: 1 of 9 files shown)


source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -136,22 +181,61 @@ const ToolRow = ({
...
+          ))}
 
         <BitkitIconButton variant="tertiary" icon={IconMinusCircle} label="Remove tool" onClick={onRemove} />
       </Box>
 
+      {useVersionDropdown && isVersionsError && (
+        <BitkitAlert
+          variant="critical"
+          messageText={`Couldn't load the available versions of ${toolId}. You can still set the version manually in the YAML editor.`}
+          action={{ label: 'Open YAML editor', onClick: () => replace('/yml') }}
```
ofalvai: I wonder if it's a good idea to hardcode URL patterns deep in components. I like the idea of an actionable error message though!

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -32,6 +39,8 @@ type ToolRowProps = {
   allowUnset?: boolean;
   autoFocus?: boolean;
   isCatalogLoading: boolean;
+  /** Bumped by the parent when the user moves on (adds another row) — reveals this row's incomplete-version error. */
+  versionTouchSignal?: number;
```
ofalvai: I have a feeling that this is a manual reimplementation of what `react-hook-form` provides out of the box, and this codebase uses the lib extensively, e.g. https://github.com/bitrise-io/bitrise-workflow-editor/blob/49afab53ada4f26245671442187698853c97cf8f/source/javascripts/pages/ContainersPage/components/CreateOrEditContainerDialog.tsx#L62

ofalvai: Also, can this input validation fix be a separate PR?

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -64,6 +90,23 @@ const ToolRow = ({
...
+  const canonicalToolId = ToolsService.resolveToolName(catalog, toolId);
+  const {
+    data: toolVersions,
+    isLoading: isVersionsLoading,
+    isError: isVersionsError,
+  } = useToolVersions(canonicalToolId, useVersionDropdown);
+
+  const versionOptions = ToolsService.getVersionOptions(toolVersions, version);
+  const isVersionMissingFromCatalog =
+    !!toolVersions && version !== '' && !ToolsService.isVersionInCatalog(toolVersions, version);
```
ofalvai, line 103: nit: `isVersionCatalog()` already handles an undefined `ToolVersions`, we could simplify this expression by just calling `isVersionCatalog()` and moving every edge case handling inside

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -64,6 +90,23 @@ const ToolRow = ({
...
+
+  // An exact strategy needs a concrete version; prefix strategies are valid without one
+  // (bare `latest`/`installed`). Only a hint — saving is not blocked.
+  const exactVersionValidation = strategy === 'exact' ? ToolsService.validateToolVersion(version) : true;
```
ofalvai: `validateToolVersion()` predates data catalog fetching and only validates the syntax, not the version values. This expression would return true in all cases.

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -136,22 +181,61 @@ const ToolRow = ({
...
+        {strategy !== 'unset' &&
+          (useVersionDropdown ? (
+            <BitkitCombobox
+              // BitkitCombobox snapshots `items` on mount and ignores later changes, so the
+              // key forces a remount whenever the list actually differs: another tool's
+              // versions, the async load completing, or the injected not-in-catalog option
+              // (getVersionOptions adds the current version (even unavailable version) when
+              // the catalog lacks it, and that entry must drop out once a real catalog version
+              // is picked).
+              key={`${canonicalToolId}:${toolVersions ? 'loaded' : 'pending'}:${isVersionMissingFromCatalog ? version : ''}`}
```
ofalvai: I don't feel qualified to comment on the validity of this, but if this is indeed required, it might be a Bitkit limitation that we could raise.

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -108,15 +143,19 @@ const ToolRow = ({
   };
 
   const handleStrategyChange = (newStrategy: VersionStrategy) => {
```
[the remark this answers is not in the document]
ofalvai: https://github.com/bitrise-io/bitrise-workflow-editor/pull/1842/commits/fbb3db4c294bbfe2b71e6f5e75e8755f0d1a5caf

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -108,15 +143,23 @@ const ToolRow = ({
...
+    // it before flagging it.
+    if (newVersion !== version) {
+      setVersionTouched(false);
+    } else if (newStrategy === 'exact' && newVersion.trim() === '') {
+      // The field was already empty, so it won't hit the branch above -> flag it immediately
+      // since it's already invalid.
+      setVersionTouched(true);
+    }
     onStrategyChange(newStrategy, newVersion);
   };
```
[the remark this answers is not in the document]
ofalvai: This is not valid, the behavior is intentional and explained in the comment: `// The strategy switch empties the field on the user's behalf -> give them a chance to fill it before flagging it.`

We don't want to immediately display an error when *there was a previous version value* before the strategy switch (e.g. `latest released version of` + `22`)

k-olly (approved):

PR #1844, opened 2026-07-14
Tweak helper text, empty state

ofalvai:
image

### Notes, decisions

- Java icon not included for obvious legal reasons :)
- I know `<EmptyState>` exists, but it doesn't allow arbitrary child components, just a title and description strings. I'd like to display a few popular tool logos because it's the easiest way to catch people's attention and convince them to use the feature, but tell me if it's too much and it should just be a standard `<EmptyState>`

### DIFF
--- a/source/javascripts/components/ToolVersions/ToolVersions.tsx
+++ b/source/javascripts/components/ToolVersions/ToolVersions.tsx
@@ -2,6 +2,7 @@ import {
   BitkitAlert,
   BitkitButton,
   BitkitLink,
+  BitkitLinkButton,
   BitkitTooltip,
   IconFlutter,
   IconNodejs,
@@ -16,13 +17,16 @@ import { useState } from 'react';
 
 import { VersionStrategy } from '@/core/models/Tools';
 import ToolsService, { ToolScope } from '@/core/services/ToolsService';
+import useNavigation from '@/hooks/useNavigation';
 import { useToolCatalog, useToolsForScope } from '@/hooks/useTools';
+import { paths } from '@/routes';
 
 import ToolRow from './ToolRow';
 
 const ToolVersions = ({ workflowId }: { workflowId?: string }) => {
   const scope: ToolScope = workflowId ? { type: 'workflow', workflowId } : { type: 'root' };
   const tools = useToolsForScope(scope);
+  const { replace } = useNavigation();
   const { data: catalog, isLoading: isCatalogLoading, isError: isCatalogError } = useToolCatalog();
   const [hasPendingRow, setHasPendingRow] = useState(false);
   const [pendingStrategy, setPendingStrategy] = useState<VersionStrategy>('latest-released');
@@ -64,6 +68,15 @@ const ToolVersions = ({ workflowId }: { workflowId?: string }) => {
             CLI and step use
           </BitkitLink>
         </Text>
+        {scope.type === 'workflow' && (
+          <Text textStyle="body/md/regular" color="text/secondary">
+            Looking for global settings which apply to all workflows? Go to the{' '}
+            <BitkitLinkButton onClick={() => replace(paths.stacksAndMachines)}>
+              Stacks &amp; Machines page
+            </BitkitLinkButton>
+            .
+          </Text>
+        )}
       </Stack>
 
       <Stack gap="16">
--- a/source/javascripts/components/ToolVersions/ToolVersions.tsx
+++ b/source/javascripts/components/ToolVersions/ToolVersions.tsx
@@ -1,5 +1,16 @@
-import { BitkitAlert, BitkitButton, BitkitLink, IconOpenInNew, rem } from '@bitrise/bitkit-v2';
-import { Box } from '@chakra-ui/react/box';
+import {
+  BitkitAlert,
+  BitkitButton,
+  BitkitLink,
+  BitkitTooltip,
+  IconFlutter,
+  IconNodejs,
+  IconOpenInNew,
+  IconPython,
+  IconRuby,
+  rem,
+} from '@bitrise/bitkit-v2';
+import { Stack } from '@chakra-ui/react/stack';
 import { Text } from '@chakra-ui/react/text';
 import { useState } from 'react';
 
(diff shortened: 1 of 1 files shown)


pgyula (dismissed):

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -87,6 +126,22 @@ const ToolVersions = ({ workflowId }: { workflowId?: string }) => {
 
       {isCatalogError && <BitkitAlert variant="warning" messageText="Couldn't load tool suggestions." />}
 
+      {Object.entries(tools).length === 0 && !hasPendingRow && (
+        <Box display="flex" flexDirection="row" gap="8" alignItems="center">
+          <Text textStyle="body/md/regular" color="text/primary">
+            Set up the first tool. Supports{' '}
+          </Text>
+          <IconRuby size="16" />
```
AndrasEszes: **[Medium] a11y:** These logo icons carry the only mention of the supported tools, but have no accessible name, so a screen reader announces the line as 'Supports  and many more.' Give each icon an aria-label (or move the tool names into the text and mark the icons aria-hidden). A BitkitTooltip would also help sighted users identify an unfamiliar logo.

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -87,6 +126,22 @@ const ToolVersions = ({ workflowId }: { workflowId?: string }) => {
 
       {isCatalogError && <BitkitAlert variant="warning" messageText="Couldn't load tool suggestions." />}
 
+      {Object.entries(tools).length === 0 && !hasPendingRow && (
+        <Box display="flex" flexDirection="row" gap="8" alignItems="center">
```
AndrasEszes: **[Low] convention:** This is a single sentence, so per the bitkit-v2 AGENTS.md it should be one `<Text>` with the icons inline, not a flex `<Box>` around split `<Text>` nodes. Collapsing it also drops the manual spacing and fixes the icon baseline alignment. The same change resolves the a11y comment above.

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -28,15 +42,40 @@ const ToolVersions = ({ workflowId }: { workflowId?: string }) => {
 
   return (
     <Box display="flex" flexDirection="column" gap="24" marginBlockStart="24" maxWidth={rem(640)}>
-      <Box display="flex" flexDirection="column" gap="8">
+      <Box display="flex" flexDirection="column" gap="4">
```
AndrasEszes: **[Low] simplification:** Prefer a named layout primitive over a raw flex Box: `<Stack gap='4'>` here (column) and `<HStack>` for the row hint below. AGENTS.md recommends Stack/HStack/VStack for flex layouts. Applies to the other flex Boxes in this file too (lines 44, 81, 130).

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -87,6 +126,22 @@ const ToolVersions = ({ workflowId }: { workflowId?: string }) => {
 
       {isCatalogError && <BitkitAlert variant="warning" messageText="Couldn't load tool suggestions." />}
 
+      {Object.entries(tools).length === 0 && !hasPendingRow && (
```
AndrasEszes: **[Nit] reuse:** `Object.entries(tools).length === 0` builds an array just to read its length. `existingToolIds` (Object.keys(tools)) is already in scope, so `existingToolIds.length === 0` does the same with less work.

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -28,15 +42,40 @@ const ToolVersions = ({ workflowId }: { workflowId?: string }) => {
...
+      <Box display="flex" flexDirection="column" gap="4">
         <Text textStyle="heading/h3">Tool setup</Text>
         <Text textStyle="body/md/regular" color="text/secondary">
-          Configure what tools and versions are required for this workflow to work. Tool setup runs before the first
-          step.{' '}
-          <BitkitLink href="#" isExternal suffixIcon={IconOpenInNew} colorVariant="purple">
+          Customize tools and versions required in {scope.type === 'workflow' ? 'this workflow' : 'workflows'}. Tool
+          setup runs before the first step.{' '}
+          <BitkitLink
+            href="https://docs.bitrise.io/en/bitrise-ci/configure-builds/configuring-build-settings/configuring-tool-versions"
```
AndrasEszes, line 53: **[Low] maintainability:** This docs base URL is hardcoded here, again on the next link, and once more in ToolRow.tsx. If the docs path is renamed, a missed copy silently 404s. Consider a shared constant with the anchor appended.

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -28,15 +42,40 @@ const ToolVersions = ({ workflowId }: { workflowId?: string }) => {
...
+          <BitkitLink
+            href="https://docs.bitrise.io/en/bitrise-ci/configure-builds/configuring-build-settings/configuring-tool-versions#tool-setup-during-workflow-execution"
+            isExternal
+            suffixIcon={IconOpenInNew}
+            colorVariant="purple"
+          >
+            CLI and step use
+          </BitkitLink>
+        </Text>
+        {workflowId && (
```
AndrasEszes: **[Nit] consistency:** This branch keys off `workflowId`, while the copy just above uses `scope.type === 'workflow'` for the same question. Not a bug (scope is derived from workflowId), but one consistent form reads clearer.

AndrasEszes (commented):
Review of the Tool setup text and empty-state tweaks. Severity legend: Medium = real user impact but not blocking, Low = cosmetic / maintainability / convention, Nit = trivial cleanup. No blocking (High) issues found.

Most findings are left inline. One spans several elements, so it goes here:

**[Low] UX (layout shift):** Clicking 'Add new' removes the empty-state hint and, at the same moment, mounts a taller ToolRow in the (previously empty) tools container above it, so the content reflows and the 'Add new' button jumps. The empty tools container also holds two 24px column gaps while empty, which adds to the movement. Consider keeping the vertical rhythm stable, e.g. render the hint in the slot the pending row will occupy so the toggle swaps content in place.

Summary: 1 Medium (a11y), 4 Low, 2 Nit. The a11y item is the only should-fix in my view; the single-Text refactor in the inline comments resolves the a11y, convention and part of the layout concern together.

source/javascripts/components/ToolVersions/ToolVersions.tsx
ofalvai: Fixed with `aria-label`

source/javascripts/components/ToolVersions/ToolVersions.tsx
ofalvai: Fixed

source/javascripts/components/ToolVersions/ToolVersions.tsx
ofalvai: Fixed

source/javascripts/components/ToolVersions/ToolVersions.tsx
ofalvai: Fixed

source/javascripts/components/ToolVersions/ToolVersions.tsx
ofalvai: Fixed

AndrasEszes (approved):

PR #1845, opened 2026-07-14
fix: cache step queries in useEnvVars and stop Algolia 20-hit truncation

zoltanszabo-bitrise:
## What

Two flag-independent Algolia-cost/correctness fixes from the Algolia Cost Reduction plan (Phase 1.1 + 1.2). Both are safe regardless of the `enable-wfe-bitrise-lsp-integration` flag state.

### 1.1 — Cache step queries in `useEnvVars` (`useEnvVars.ts`)
`useStepLevelEnvVars`' `useQueries` reuse the same `['steps', { cvs, defaultStepLibrary }]` queryKey as `useStep`, but set no cache options. Because `staleTime` is evaluated **per-observer**, those observers considered the shared cache entry stale on mount and refetched on every EnvVar popover open — plus retried 3× (inherited `retry: 3`). Now mirrors `useStep`: `staleTime/gcTime: Infinity, retry: false`. Each step's outputs (2 browse calls each) stop refetching per popover open.

### 1.2 — Stop silent 20-hit truncation (`AlgoliaApi.ts`)
`getStepsByMultipleCvs` and `getAllAvailableVersionsByIds` omitted `hitsPerPage`, so they defaulted to Algolia's 20-hit cap and silently truncated the `$`-output completion list for configs with >20 steps/versions. `getAllAvailableVersionsByIds` is the riskier one — one id matches every version record, so hits multiply across ids. Added `hitsPerPage: 1000`, and corrected the misleading `searchSingleIndex … to cache` comment (browse caches too, with `{ cacheable: true }`).

## Verification
- `eslint` clean on both files.
- No behavioral test added: both changes are declarative param additions (React Query / Algolia config), and 1.1 mirrors the already-working `useStep` pattern.

## Out of scope / parked
- The browse↔search API choice for the two `$`-completion calls — gated on V1/V2 billing verification (Algolia Usage/Logs API). The corrected comments document the `browse + { cacheable: true }` option for a future pass.
- All LSP-repo changes (batching, worker lifetime) live in `bitrise-yml-language-server`.

### DIFF
--- a/source/javascripts/core/api/AlgoliaApi.ts
+++ b/source/javascripts/core/api/AlgoliaApi.ts
@@ -104,11 +104,13 @@ async function getAllStepsById(id: string) {
 }
 
 async function getStepsByMultipleCvs(cvss: string[], attributesToRetrieve?: string[]) {
-  // NOTE: Using searchSingleIndex instead of browseObjects to cache the results
+  // NOTE: searchSingleIndex is used because it is cached by the client (browse is not, unless
+  // called with { cacheable: true }). hitsPerPage avoids the default 20-hit truncation.
   const response = await client.searchSingleIndex<AlgoliaStepResponse>({
     indexName: ALGOLIA_STEPLIB_STEPS_INDEX,
     searchParams: {
       filters: cvss.map((cvs) => `cvs:${cvs}`).join(' OR '),
+      hitsPerPage: 1000,
       analytics: false,
       clickAnalytics: false,
       attributesToRetrieve,
@@ -121,11 +123,14 @@ async function getStepsByMultipleCvs(cvss: string[], attributesToRetrieve?: stri
 async function getAllAvailableVersionsByIds(ids: string[]) {
   const results: Map<string, Set<string>> = new Map();
 
-  // NOTE: Using searchSingleIndex instead of browseObjects to cache the results
+  // NOTE: searchSingleIndex is used because it is cached by the client (browse is not, unless
+  // called with { cacheable: true }). hitsPerPage avoids the default 20-hit truncation — one id
+  // matches every version record, so hits multiply across ids and 20 is easily exceeded.
   const response = await client.searchSingleIndex<AlgoliaStepResponse>({
     indexName: ALGOLIA_STEPLIB_STEPS_INDEX,
     searchParams: {
       filters: ids.map((id) => `id:${id}`).join(' OR '),
+      hitsPerPage: 1000,
       analytics: false,
       clickAnalytics: false,
       attributesToRetrieve: ['id', 'version'],
--- a/source/javascripts/core/api/AlgoliaApi.ts
+++ b/source/javascripts/core/api/AlgoliaApi.ts
@@ -104,40 +104,58 @@ async function getAllStepsById(id: string) {
 }
 
 async function getStepsByMultipleCvs(cvss: string[], attributesToRetrieve?: string[]) {
-  // NOTE: searchSingleIndex is used because it is cached by the client (browse is not, unless
-  // called with { cacheable: true }). hitsPerPage avoids the default 20-hit truncation.
-  const response = await client.searchSingleIndex<AlgoliaStepResponse>({
-    indexName: ALGOLIA_STEPLIB_STEPS_INDEX,
-    searchParams: {
-      filters: cvss.map((cvs) => `cvs:${cvs}`).join(' OR '),
-      hitsPerPage: 1000,
-      analytics: false,
-      clickAnalytics: false,
-      attributesToRetrieve,
-    },
-  });
+  // No cvss → skip the query. An empty `filters` string is match-all in Algolia, which would
+  // return unrelated steps instead of nothing.
+  if (cvss.length === 0) {
+    return [];
+  }
 
-  return response.hits;
+  // browseObjects (not searchSingleIndex) to get every matching record: searchSingleIndex returns
+  // a single relevance-ranked page and would silently truncate large configs. { cacheable: true }
+  // opts browse into the client's response cache (browse is not cached by default).
+  const results: AlgoliaStepResponse[] = [];
+  await client.browseObjects<AlgoliaStepResponse>(
+    {
+      indexName: ALGOLIA_STEPLIB_STEPS_INDEX,
+      aggregator: ({ hits }) => results.push(...hits),
+      browseParams: {
+        filters: cvss.map((cvs) => `cvs:${cvs}`).join(' OR '),
+        attributesToRetrieve,
+      },
+    },
+    { cacheable: true },
+  );
+
+  return results;
 }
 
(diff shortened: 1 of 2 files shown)


source/javascripts/core/api/AlgoliaApi.ts
```diff
@@ -104,11 +104,13 @@ async function getAllStepsById(id: string) {
...
 
 async function getStepsByMultipleCvs(cvss: string[], attributesToRetrieve?: string[]) {
-  // NOTE: Using searchSingleIndex instead of browseObjects to cache the results
+  // NOTE: searchSingleIndex is used because it is cached by the client (browse is not, unless
+  // called with { cacheable: true }). hitsPerPage avoids the default 20-hit truncation.
   const response = await client.searchSingleIndex<AlgoliaStepResponse>({
     indexName: ALGOLIA_STEPLIB_STEPS_INDEX,
     searchParams: {
       filters: cvss.map((cvs) => `cvs:${cvs}`).join(' OR '),
+      hitsPerPage: 1000,
```
AndrasEszes: When `cvss` is empty, `cvss.map(...).join(' OR ')` produces `filters: ''`, which Algolia treats as match-all. With `hitsPerPage: 1000` that now returns up to 1000 unrelated steps instead of the previous 20.

The only caller, `MonacoUtils.loadStepOutputs`, passes `[]` here whenever a bitrise.yml has no Bitrise library steps (e.g. only script/git/path steps), and then injects all those steps' outputs into the `$` env var autocomplete. The sibling `getAllAvailableVersionsByIds([])` fires the same wasted match-all query.

Suggestion: early-return on empty input (either here or in `loadStepOutputs`) so the degenerate case does no query and adds no bogus suggestions.

source/javascripts/core/api/AlgoliaApi.ts
```diff
@@ -121,11 +123,14 @@ async function getStepsByMultipleCvs(cvss: string[], attributesToRetrieve?: stri
...
 
-  // NOTE: Using searchSingleIndex instead of browseObjects to cache the results
+  // NOTE: searchSingleIndex is used because it is cached by the client (browse is not, unless
+  // called with { cacheable: true }). hitsPerPage avoids the default 20-hit truncation — one id
+  // matches every version record, so hits multiply across ids and 20 is easily exceeded.
   const response = await client.searchSingleIndex<AlgoliaStepResponse>({
     indexName: ALGOLIA_STEPLIB_STEPS_INDEX,
     searchParams: {
       filters: ids.map((id) => `id:${id}`).join(' OR '),
+      hitsPerPage: 1000,
```
AndrasEszes: `searchSingleIndex` only returns the first page, so `hitsPerPage: 1000` raises the cap but does not remove truncation. Here one id matches every version record, so a large config (roughly 40+ ids with 25+ versions each) can exceed 1000 records.

The returned 1000 hits are relevance-ranked, not grouped per id, so a step's latest version can be dropped, and `VersionUtils.resolveVersion` in `MonacoUtils` then resolves the wrong version.

The comment is accurate that the 20-hit default is gone, but it would help to note the residual 1000 cap (or switch to browse if full completeness is required for this call).

source/javascripts/hooks/useEnvVars.ts
```diff
@@ -120,6 +120,11 @@ const useStepLevelEnvVars = (ids: string[], enabled: boolean) => {
       enabled,
       queryKey: ['steps', { cvs, defaultStepLibrary }],
       queryFn: () => StepApi.getStepByCvs(cvs, defaultStepLibrary),
+      // Same queryKey as useStep — mirror its cache policy so opening the EnvVar popover reuses
+      // already-fetched steps instead of refetching (staleTime is per-observer) and retrying 3×.
+      staleTime: Infinity,
+      gcTime: Infinity,
```
AndrasEszes, line 126: `gcTime: Infinity` keeps every fetched step in cache for the tab's lifetime. Unlike `useStep` (which caches only steps rendered as StepCards), this hook walks the full workflow chain (before_run/after_run/utility), so steps that are never rendered are now retained permanently too.

This matches the cost-reduction plan's accepted no-TTL tradeoff, so no change needed if intentional. Just flagging that the retained set is broader here than in `useStep`.

AndrasEszes (commented):
Automated review of the two changes. Both stated fixes are correct and match the `useStep` pattern. Three inline notes below (mostly follow-ups and one optional). One more item is outside this diff, so I could not attach it inline:

**useEnvVars.ts (around line 107, in `useStepLevelEnvVars`, outside this diff):** the step selector does `const [cvs, step] = Object.entries(ymlStepObject)[0]` with no fallback, while `useStep.ts` guards the identical pattern with `?? ['', {}]`. An empty or malformed step entry (`- {}`, or a null list item) makes `Object.entries(...)[0]` undefined, and destructuring undefined throws inside the Zustand selector, crashing the EnvVar popover render. Adding `?? ['', {}]` would match `useStep` and avoid the crash. Worth fixing here since this PR already touches the function.

source/javascripts/core/api/AlgoliaApi.ts
zoltanszabo-bitrise: It's an incremental improvement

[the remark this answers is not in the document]
zoltanszabo-bitrise: Switched to `browseObjects` (with `{ cacheable: true }`) so all matching records are enumerated — no page cap — while staying off the billed `/query` metric and still cached.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

source/javascripts/core/api/AlgoliaApi.ts
zoltanszabo-bitrise: Added an early return when `cvss`/`ids` is empty, so the degenerate match-all query no longer fires (in both this and `getAllAvailableVersionsByIds`).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

source/javascripts/core/api/AlgoliaApi.ts
zoltanszabo-bitrise: Right — `hitsPerPage` only raised the cap. Replaced it with `browseObjects`, which enumerates every version record, so the latest-version drop / wrong resolution can't happen.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

source/javascripts/hooks/useEnvVars.ts
zoltanszabo-bitrise: Intentional — this is the plan's accepted no-TTL tradeoff, and `staleTime: Infinity` needs `gcTime: Infinity` to actually avoid refetch on popover reopen. Noted that the retained set (full workflow chain) is broader than useStep's; leaving as-is.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

AndrasEszes (approved):

PR #1848, opened 2026-07-15
refactor(modular): build entity index in WFE, drop BE index consumption [BIVS-3728]

trapacska:
## What

The WFE now builds the modular **entity index itself** from the loaded file documents and no longer consumes the backend-served `entity_index`. The WFE is the single source of truth for the index.

## Why

Index building is a pure function of the file tree, and the WFE already has all the files. It was previously dual-sourced — the BE shipped an `entity_index` snapshot at load/save *and* the FE re-derived it live from file documents (`EntityIndexService.buildFromFiles`) for unsaved-edit correctness. This removes the BE dependency, so cloud and (upcoming) local behave identically and there's one implementation to reason about. Retiring the BE `EntityIndexBuilder` is a follow-up (BIVS-3728 next step).

## Changes

- `getConfig` no longer parses `wire.entity_index`; `entityIndex` dropped from `GetConfigResponse` (removed `fromWireEntityIndex`/`fromWireEntityEntries`). The BE still emits `entity_index` on the wire — it's simply ignored.
- `initializeModularConfig` / `applyModularSaveResult` build the index from files via `EntityIndexService.buildFromFiles`. `entityIndex` is kept only as an **optional override** for stories/tests; production omits it.
- The existing `files` subscription already re-derives the index on every edit, so this only changes the *seed* (build-from-files instead of BE snapshot).

## Scope

Index only. The tree/files still come from the BE `/config/tree` on cloud; local file-sourcing is a separate ticket (BIVS-3729).

## Testing

`tsc` clean; 77/77 modular tests pass (`BitriseYmlStore.modular`, `EntityIndexService`, `TreeService`); lint clean on changed files.

### DIFF
--- a/source/javascripts/components/FileTreeViewer/FileTreeViewer.stories.tsx
+++ b/source/javascripts/components/FileTreeViewer/FileTreeViewer.stories.tsx
@@ -3,7 +3,7 @@ import { BitkitTabs } from '@bitrise/bitkit-v2';
 import { Meta, StoryObj } from '@storybook/react-vite';
 import { useRef, useState } from 'react';
 
-import { EntityIndex, TreeNode } from '@/core/models/Tree';
+import { TreeNode } from '@/core/models/Tree';
 import { initializeModularConfig } from '@/core/stores/BitriseYmlStore';
 
 import FileTreeViewer from './FileTreeViewer';
@@ -70,12 +70,10 @@ const ROOT: TreeNode = {
   ],
 };
 
-const ENTITY_INDEX: EntityIndex = { workflows: {}, pipelines: {}, stepBundles: {} };
-
 export default {
   component: FileTreeViewer,
   beforeEach: () => {
-    initializeModularConfig({ root: ROOT, entityIndex: ENTITY_INDEX, branch: 'main', commitSha: ROOT.commitSha });
+    initializeModularConfig({ root: ROOT, branch: 'main', commitSha: ROOT.commitSha });
   },
   decorators: [
     (Story) => (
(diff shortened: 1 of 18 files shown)


moczolaszlo (dismissed):
LGTM. Nice consolidation to a single source of truth for the index.

Checked the things most likely to bite:
- `buildFromFiles` covers all kinds **and** `app.envs` (keyed by var name, `opts` skipped), so cross-file project env var provenance keeps working.
- Precedence (node before its includes, siblings in reverse) mirrors the Go merger, matching the BE index semantics.
- It's the same function the `files` subscription already uses to re-derive on every edit, so the seed and the live index are identical by construction — this actually removes the old dual-source (BE snapshot vs FE re-derivation) discrepancy risk rather than adding one.
- Callers (`main.tsx`, `usePushBranch`) updated; the optional `entityIndex` override keeps stories/tests working.

Minor (non-blocking): `WireEntityIndex` / `WireEntityEntries` are now only referenced by the unparsed `entity_index` wire field — fine to keep given the BE still emits it, as the PR notes.

moczolaszlo:
Re-reviewed the follow-up commit (`8f5f6e96`, drop the injection param) — still good, arguably cleaner: `entityIndex` is gone from the `initializeModularConfig` / `applyModularSaveResult` API entirely, so there's a single source (always `buildFromFiles`) with no inject-vs-build duality. All callers (stories + specs) updated to match.

Checked out the branch and ran it: `tsc` clean, and the affected suites are green (78/78 across `BitriseYmlStore.modular`, `EntityIndexService`, `useTree`, `useContainers`, `useContainerWorkflowUsage`, `useTargetBasedTriggers`, `useAIButton`, `useMergedConfigSync`). The specs that previously injected a hand-crafted index now derive it from their own file contents via `buildFromFiles`, and the assertions still hold. Approval stands.

moczolaszlo (approved):
Re-checked after `2e63c31a` (drop `entity_index` from the wire layer) — clean, and it resolves the earlier nit: `WireEntityIndex` / `WireEntityEntries` and the `entity_index` wire field are gone, along with the mock fixtures. Grepped the branch: no lingering references; the remaining `entityIndex` usages are the store's own state, built via `EntityIndexService.buildFromFiles`. Approval stands.

PR #1851, opened 2026-07-16
feat(local-modular): modular YML editing in the local (CLI) Workflow Editor [BIVS-3729]

trapacska:
## What

Brings modular (`include:`) YML editing to the **local** Workflow Editor (the CLI apiserver), with the same experience as cloud — tree of module tabs, per-file editing, merged view, and save. The frontend is unchanged in shape; only the endpoints differ by mode.

## Backend (CLI apiserver, Go)

New handlers in `apiserver/service/bitrise_config_tree.go` (+ routes):

- `GET /api/bitrise-yml/tree` — resolves the include tree from disk via the bitrise CLI's `configmerge`, returns the same wire shape as the hosted `/config/tree` plus the merged YAML. A non-modular config comes back as a single root node, so the FE consumes one shape. `node_id` / `source` / `editable` are derived from the configmerge tree key (cross-ref nodes — other repo/branch/tag/commit — load **read-only**, matching cloud). No entity index is sent; the FE builds it.
- `POST /api/bitrise-yml/tree` — validates the **merged** tree (merged-only, like cloud), then writes each **editable + modified** module file to disk. Read-only files are never written; unmodified files are skipped.
- `POST /api/bitrise-yml/tree/merge` — flattens the posted (possibly-edited) tree so the merged view reflects in-memory edits.

## Frontend

Same code as cloud, mode-aware endpoints:

- `BitriseYmlApi`: `configTreePath` / `mergeConfigPath` branch on `isWebsiteMode` (local → `/api/bitrise-yml/tree[/merge]`, no app slug); new `saveConfigTree` for the local disk save.
- `useSaveConfigTree`: write the tree to disk, then reload it.
- `Header`: local modular save writes to disk + reloads (there's no branch/PR locally — cloud still saves via push-to-branch); Save button pending state includes it.
- `main.tsx`: the modular LD flag gates **website only**. Local has no LaunchDarkly, so modular is simply on — a non-modular config resolves to a single-node tree and is driven exactly like the single-file flow.

## Testing

- Go: `parseNodeKey` (local / x-repo / x-tag), tree GET (non-modular + modular), tree save (writes editable+modified), tree merge — all green. `go build` / `go vet` clean.
- FE: `tsc` clean, eslint clean; jest green for the API path-branching spec, the modular store, and merged-config sync.

## Notes / follow-ups

- Cross-ref (x-repo / branch / tag / commit) files load read-only locally, same as cloud; `configmerge` fetches x-repo modules via local git for the merged view.
- Not yet exercised end-to-end inside the packaged CLI plugin against a real modular repo — worth a manual smoke test before GA.

### DIFF
--- a/apiserver/routes.go
+++ b/apiserver/routes.go
@@ -27,6 +27,12 @@ func SetupRoutes(isServeFilesThroughMiddlemanServer bool) (*mux.Router, error) {
 	r.HandleFunc("/api/bitrise-yml.json", wrapHandlerFunc(service.GetBitriseYMLAsJSONHandler)).Methods("GET")
 	r.HandleFunc("/api/bitrise-yml.json", wrapHandlerFunc(service.PostBitriseYMLFromJSONHandler)).Methods("POST")
 
+	// Modular config (include tree): resolve the tree from disk, save changed module files, and
+	// merge the live tree. The FE drives these in local mode exactly like the hosted /config/tree.
+	r.HandleFunc("/api/bitrise-yml/tree", wrapHandlerFunc(service.GetBitriseYMLTreeHandler)).Methods("GET")
+	r.HandleFunc("/api/bitrise-yml/tree", wrapHandlerFunc(service.PostBitriseYMLTreeHandler)).Methods("POST")
+	r.HandleFunc("/api/bitrise-yml/tree/merge", wrapHandlerFunc(service.PostBitriseYMLTreeMergeHandler)).Methods("POST")
+
 	r.HandleFunc("/api/secrets", wrapHandlerFunc(service.GetSecretsAsJSONHandler)).Methods("GET")
 	r.HandleFunc("/api/secrets", wrapHandlerFunc(service.PostSecretsYMLFromJSONHandler)).Methods("POST")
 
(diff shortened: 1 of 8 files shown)


moczolaszlo (approved):
LGTM — nice cross-stack work, and it reuses the cloud wire shape so the FE stays single-shape.

Backend: the handlers look right — cross-ref (repo/branch/tag/commit) nodes load read-only, POST validates the *merged* tree (like cloud) and only writes editable+modified files, non-modular collapses to a single root node. Good test coverage (parseNodeKey, GET modular/non-modular, POST write, merge).

Frontend: mode-aware endpoints, save-to-disk + reload, and `main.tsx` keeps non-modular on the single-file path via the existing bootstrap. Header's onError handles both a `ClientError` and the plain `Error` from the guard.

On the bot's `useCiConfig.ts` comment (plain `Error` → runtime crash): it's a false positive — `Header`'s onError guards `error instanceof ClientError` before `getResponseErrorMessage()` and falls back to `error.message`, so the thrown `Error` renders fine.

One optional, non-blocking hardening: `PostBitriseYMLTreeHandler` trusts the client-supplied `editable`/`path` for the disk write. It's a localhost tool on the user's own repo (and mirrors the cloud tree-save serialization), so low risk — but as defense-in-depth you could re-derive `editable` server-side from `node.Source` so a read-only cross-ref file can never be written even if the flag is stale. Fine to leave for a follow-up.

PR #1852, opened 2026-07-17
Replace manual validation with react-hook-form

ofalvai:
Current tool setup input validation is hand-rolled, and before https://github.com/bitrise-io/bitrise-workflow-editor/pull/1842 adds even more complexity to it, I think it would make sense to use `react-hook-form`, just like in other parts of the codebase.

### Changes:
- Replace manual validation and state handling with RHF
- [Separate commit] The above change caused a weird bug around the `manualOther` state, and form validation flipping the state back to `false` when selecting `Other` from the dropdown. I'm not 100% sure of the root cause, but this made me realize that the whole autofocus logic is largely irrelevant now that a new pending row doesn't render a text input as its initial state. So I decided to clean up autofocus logic completely, and it resolved the original bug.

### DIFF
--- a/source/javascripts/components/ToolVersions/ToolRow.tsx
+++ b/source/javascripts/components/ToolVersions/ToolRow.tsx
@@ -9,10 +9,15 @@ import {
 } from '@bitrise/bitkit-v2';
 import { Box } from '@chakra-ui/react/box';
 import { Text } from '@chakra-ui/react/text';
-import { FocusEventHandler, useState } from 'react';
+import { useState } from 'react';
+import { useController, useForm } from 'react-hook-form';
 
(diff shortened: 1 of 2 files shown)


ofalvai:
@zoltanszabo-bitrise well, now you do, sorry 😆 https://github.com/search?q=repo:bitrise-io/bitrise-workflow-editor%20react-hook-form&type=code

This made me believe this is an existing and "blessed" pattern. The context is this other PR: https://github.com/bitrise-io/bitrise-workflow-editor/pull/1842/changes#r3577355910

What's your recommendation, is there a good and recommended pattern to use, or this mostly depends on the context?

AndrasEszes (approved):

PR #1854, opened 2026-07-20
feat: support the multi-file Bitrise language service

zoltanszabo-bitrise:
## What

Integrates the **multi-file** version of `@bitrise/languageserver` ([language-server PR #33](https://github.com/bitrise-io/bitrise-yml-language-server/pull/33)) into the Workflow Editor. Cross-file symbol resolution (definition, references, hover, diagnostics, completion) and clickable `include:` links now work across a modular `bitrise.yml` (local **and** cross-repo includes), and the read-only merged-config tab gets its own isolated in-file navigation.

## Why

The language service now treats every registered document as **one workspace** joined by top-level `include:`. The host's job is to register each include-tree file under a `bitrise://<repo>/<path>[?ref]` URI composed via the shared codec, so an include target the core resolves matches the model the host registered (exact-string match). WFE previously registered a single `file:///bitrise.yml` model and only mounted the selected modular file — so cross-file resolution and include links never lined up.

## What's in it

- **`bitrise://` model URIs from the include tree** — `lspModelUris.ts` (`buildNodeUris`) walks the `Tree` and composes one URI per node via `composeBitriseUri` (root → working repo `.`; a local include inherits its includer's repo+ref; a cross-repo include uses its own repository + ref). Unit-tested.
- **Whole tree as one workspace** — `useYmlLanguageServices` now reconciles one Monaco model per tree file under those URIs (create / sync-if-detached / dispose-when-gone), replacing the single background model, and aggregates validation markers across all of them. `ModularYmlEditor` binds each editor's model to the shared `bitrise://` URI (`+keepCurrentModel`) so the visible model **is** the one the service analyzes.
- **Isolated merged preview** — the read-only merged tab runs a *second* language-service instance on the `bitrise-merged://` scheme, so it gets in-file go-to-definition / references / hover as a self-contained single-file workspace, with zero symbol bleed into the real `bitrise://` tree.
- **Cross-file navigation** — registers a `monaco.editor.registerEditorOpener` that maps a target `bitrise://` URI back to its tree node, opens that file's tab (`openTab`), and reveals the range once the target editor mounts. (In-file / merged-internal nav is the same-model fast path.)
- **Dependency + tooling** — bumps `@bitrise/languageserver` to the multi-file build; adds a pure-codec jest stand-in for the `@bitrise/languageserver-core` alias (Vite/tsconfig-only, unresolvable by jest).

## Companion changes (language-server repo, PR #33 branch)

A few gaps surfaced during integration were fixed on the LS side:
- Scope the Bitrise workspace to `bitrise://` models (a host `file://` YAML model — like a merged preview — must not join the tree and duplicate every symbol).
- Add a `scheme` option to `configureBitriseYaml` so a host can run a second isolated instance (the merged preview).
- Demo the merged-preview instance in the example app.

## Verification

- `tsc --noEmit` clean, `eslint` clean, **1418 jest tests pass**.
- The end-to-end multi-file + isolated-merged behavior was verified live in the language-server example app (cross-file hover/definition, include links, and the isolated merged tab). Verifying inside WFE itself needs the Go apiserver + a real modular/enterprise config loaded.

## Notes for reviewers

- Gated behind the existing `enable-wfe-bitrise-lsp-integration` flag.
- The `@bitrise/languageserver` dependency currently pins the PR #33 **branch** commit; repoint it to the squash-merge SHA once #33 lands on `main`.
- Data-contract check to confirm on a real modular project: `/config/tree` must populate `TreeNode.source` (path/repository/branch/tag/commit) per include, since `buildNodeUris` derives each `bitrise://` identity from it.


source/javascripts/hooks/useYmlLanguageServices.ts
```diff
@@ -21,71 +88,127 @@ function useYmlLanguageServices() {
...
+          ownedUris.delete(uri);
+        }
+      });
+    };
+
+    reconcile(true);
+
+    // Aggregate validation status across every owned model — in modular mode a single file's markers
+    // don't tell the whole story (a cross-file reference can be unresolved in another file).
+    const markerDisposable = MonacoUtils.onWorkspaceMarkerStatusChange(
```
AndrasEszes: **[High] Partial include modules flip global validation to `invalid` and trap the user on the YAML page.**

This sums markers across every include model. A module with only `workflows:` (e.g. the new `module1.yml`) has neither `format_version` nor `include:`, so monaco-yaml's schema reports an error (top-level `anyOf` requires one of them, and `fileMatch: ['*']` matches `bitrise://` URIs). That flips `validationStatus` to `invalid`, and `InvalidYmlRedirect` bounces the user from the visual editor to the YAML page and blocks Save. Pre-PR only the active document was watched, so this is a regression. Consider excluding partial include files from the aggregate, or gating global validity on the root/merged config only.

source/javascripts/hooks/useYmlLanguageServices.ts
```diff
@@ -1,11 +1,78 @@
...
+}
+
+// After a tab switch the target file's editor mounts asynchronously (React re-render + model swap),
+// so poll a few frames for the editor now showing that model, then reveal. Bounded so a target that
+// never mounts (e.g. tab open was blocked) can't spin forever.
+function revealWhenReady(uri: string, range: monaco.IRange, attempts = 0) {
+  const editor = monaco.editor.getEditors().find((e) => e.getModel()?.uri.toString() === uri);
+  if (editor) {
+    revealInEditor(editor, range);
+  } else if (attempts < 20) {
```
AndrasEszes: **[Medium] Cross-file reveal can silently fail on slow or large targets.**

The reveal polls at most 20 animation frames (about 330ms) for the target editor to mount. If mount takes longer, the tab opens but the cursor is never positioned, so the user lands at line 1 with no hint where the definition is. Consider revealing from the target editor's own mount / `onDidChangeModel` instead of a bounded poll, or raise the bound.

source/javascripts/hooks/useYmlLanguageServices.ts
```diff
@@ -21,71 +88,127 @@ function useYmlLanguageServices() {
...
+        if (desiredUris.has(uri)) return;
+        const model = monaco.editor.getModel(monaco.Uri.parse(uri));
+        if (model && !model.isAttachedToEditor()) {
+          model.dispose();
+          ownedUris.delete(uri);
+        }
+      });
+    };
+
+    reconcile(true);
```
AndrasEszes, line 165: **[Medium] Reconcile, marker aggregation, and the editor-opener are not behind the LSP flag.**

Only `configureBitriseLanguageServer` is gated by `enableBitriseLsp`. `reconcile`, `onWorkspaceMarkerStatusChange`, and `registerEditorOpener` always run, so the behavior changes above (including the validation redirect) ship even with the flag off. The flag is not an effective kill-switch. Consider gating this block too.

AndrasEszes (commented):
## Code review notes

Inline comments below are tagged with severity. Quick summary:

- **[High]** Partial include modules break the visual editor for modular configs (inline on `useYmlLanguageServices.ts`). This is the one to fix before merge.
- **[Medium]** Bounded reveal poll, `node.path` vs include-path URI risk, and the LSP flag not gating the new behavior.
- **[Low]** Per-keystroke re-serialization, and permanent tabs on cross-file navigation.

**[Low] Convention (not inline, the line is unchanged):** `source/javascripts/pages/YmlPage/components/ModularYmlEditor.tsx` still imports `Box` from `@bitrise/bitkit` (v1). CLAUDE.md asks to migrate v1 components to v2 in any file you touch.

Confidence note: the High finding and the `node.path` finding depend on runtime behavior I reasoned about from the schema and codec sources but did not run live in WFE. That matches the PR note that it was verified in the language-server example app rather than inside WFE.

source/javascripts/hooks/useYmlLanguageServices.ts
```diff
@@ -1,11 +1,78 @@
...
+  if (!tree) {
+    return [{ uri: ROOT_MODEL_URI, content: getYmlString() }];
+  }
+
+  // ponytail: re-serializes (toYml) every file on each store change. Fine at editor scale (a handful
+  // of include files); if trees ever get large, diff against a cached per-node serialization first.
+  const uriByNode = buildNodeUris(tree);
+  return Object.values(files).flatMap((file) => {
+    const uri = uriByNode.get(file.nodeId);
+    return uri ? [{ uri, content: YmlUtils.toYml(file.ymlDocument) }] : [];
```
AndrasEszes: **[Low] Every keystroke re-serializes all files.**

On each store change (every keypress) `desiredModels()` rebuilds `buildNodeUris(tree)` and runs `YmlUtils.toYml` on every file, then string-compares each. Cost scales with total tree size per keypress. Fine for a few small modules, but consider diffing against a cached per-node serialization if trees grow (as the `ponytail` comment notes).

source/javascripts/hooks/useYmlLanguageServices.ts
```diff
@@ -21,71 +88,127 @@ function useYmlLanguageServices() {
...
+        if (source.getModel()?.uri.toString() === uri) {
+          revealInEditor(source, range);
+          return true;
+        }
 
-    // Subscribe to marker changes
-    const markerDisposable = MonacoUtils.onModelMarkerStatusChange(model, setValidationStatus);
+        const nodeId = nodeIdForUri(uri);
+        if (!nodeId) return false; // not one of our tree models — let Monaco's default handling run
+        openTab(nodeId, { preview: false });
```
AndrasEszes, line 199: **[Low] Cross-file navigation pins a permanent tab.**

`openTab(nodeId, { preview: false })` opens a non-preview tab on every go-to-definition or include-link click, so peeking at a definition clutters the tab strip. A preview tab (`preview: true`) would match normal go-to-definition behavior.

source/javascripts/core/utils/lspModelUris.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (!parent) {
+      identity = { repo: '.' };
+    } else if (source?.repository) {
+      identity = { repo: source.repository, ref: refFromSource(source) };
+    } else {
+      identity = identityByNode.get(parent.nodeId) ?? { repo: '.' };
+    }
+
+    identityByNode.set(node.nodeId, identity);
+    uriByNode.set(node.nodeId, composeBitriseUri({ ...identity, path: node.path }));
```
AndrasEszes, line 66: **[Medium] URI is composed from `node.path`, but core resolves include edges from the include-directive path.**

`canonicalizePath` only strips `./`, `//`, and a leading `/`. If `TreeNode.path` is absolute, repo-qualified, or otherwise differs from the repo-relative path string core uses, the host and core URIs diverge and this file silently drops out of the workspace: no cross-file symbols, hover, include links, or navigation. Worth confirming `TreeNode.path` matches the include path exactly. The reviewer note references `source`, but the code keys the path on `node.path`.

source/javascripts/hooks/useYmlLanguageServices.ts
zoltanszabo-bitrise: Good catch on the intent, but I've left the opener ungated deliberately (and added a comment saying so). The cross-file targets `registerEditorOpener` acts on are produced by the LS's definition/reference/link providers, which only exist when `enable-wfe-bitrise-lsp-integration` is on. With the flag off, Monaco never hands the opener a `bitrise://` target, so registering it is inert.

The genuinely harmful flag-off behavior you flagged was the validation redirect — that's fixed directly (ec481f44): global validation now tracks the root model only, so it no longer flips to `invalid` on a partial include, flag on *or* off. `reconcile` itself stays ungated because it builds the models the editors bind to — modular editing needs them regardless of the LSP.

source/javascripts/hooks/useYmlLanguageServices.ts
zoltanszabo-bitrise: Holding off on this one. Our tab strip isn't Monaco's/VS Code's preview-tab UI (no italic + single-click-replace peek behavior), so `preview: true` wouldn't give the "peek" UX you're picturing — it'd just make the tab silently replaceable by the next navigation, which is arguably worse for go-to-definition. Leaving `preview: false` for now.

source/javascripts/hooks/useYmlLanguageServices.ts
```diff
@@ -21,71 +88,127 @@ function useYmlLanguageServices() {
...
 
-    // Create or reuse the background model
-    let model = monaco.editor.getModel(BACKGROUND_MODEL_URI);
-    if (!model) {
-      model = monaco.editor.createModel(getYmlString(), 'yaml', BACKGROUND_MODEL_URI);
-    }
+    // URIs of the models we own, so a reconcile can dispose ones whose file left the tree. Models
+    // are shared with the editor components (same URI ⇒ same Monaco model), so the worker analyzes
+    // exactly what the user sees and edits.
+    const ownedUris = new Set<string>();
```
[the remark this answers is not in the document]
zoltanszabo-bitrise: Acknowledged, skipping as low-value. The effect only re-runs when the LSP feature flag flips (its sole dependency), which doesn't happen in normal use. And by design these models live for the entire app session — the cleanup intentionally never disposes them (async workers run on them). So the only missed disposal is a model whose file left the tree *during* a mid-session flag flip — a narrow enough case that hoisting the set isn't worth the extra module-level state.

source/javascripts/core/utils/lspModelUris.ts
zoltanszabo-bitrise: By design the `bitrise://` URI is a single string that encodes everything an `include:` can describe — `repository → authority`, `ref → query`, `path → path` — so one URI unambiguously identifies any target (local, cross-repo, or ref-pinned), which is all Monaco gets to navigate with. Host and core compose that same string through the shared codec (`composeBitriseUri`), so resolution is exact-string match.

`repo` and `ref` come from the node's own include fields on both sides, so they agree by construction. The path is the one slot filled from different places — host from `node.path`, core from the include directive (`entry.path`) — and both run it through the same `canonicalizePath` (`./`, `//`, leading `/` stripped; `../` rejected). In our tree fixtures they're byte-identical, so it lines up. The only residual assumption is that the real BE always sets `node.path` to the plain repo-relative directive path (never a display/absolute form) — a data-contract item already flagged in the PR description to confirm on a real modular/enterprise project. Not adding defensive normalization, which would mask a genuine mismatch rather than surface it.

AndrasEszes (approved):

PR #1864, opened 2026-07-27
feat: unmask interface chrome in Clarity session recordings

bitrise-io-kolega:
Clarity masks text by default, so recordings redacted the navigation, page titles, tab labels and button captions along with the customer's configuration, making them hard to interpret. This applies `data-clarity-unmask="true"` to the static shell only, and leaves everything that renders configuration or user input masked. 53 files, +169/-59 — attributes and comments only, no logic changes.

**Unmasked (identical for every user/account):**
- App shell: whole sidebar navigation; the fixed "Bitrise CI" / "CI configuration" breadcrumbs; the Visual/YAML switch; the Show diff / Discard / Save changes button group.
- Page headings + intro copy: Environment Variables, Stacks & Machines, Containers, Triggers, Licenses, Secrets (incl. section headings).
- Tab strips: page-level tabs plus fixed tab labels in the workflow / step / step bundle / pipeline / step-selector drawers and panels.
- Empty, loading and error states.
- Static informational notifications, ribbons, read-only banner.
- Static table column headers.
- Fixed menu items, dialog footers, and the static parts of the delete workflow / delete step bundle dialogs.

**Left masked (unchanged):** workflow / pipeline / step bundle / step / container names and ids, step inputs, env var names and values, secrets, the project-name breadcrumb, branch and storage labels, module file paths, and the YAML editor.

Because the attribute is inherited by the whole subtree, it's attached to the narrowest element containing only fixed copy. Where a container mixes both (breadcrumb, config drawer headers, delete dialogs, pipeline toolbar, config settings menu) the static children are tagged individually and the container stays masked. A note next to the Clarity loader records the convention for future changes.

**Judgement calls a reviewer may want to overturn (also noted in the commit message):**
1. Container tab badges show a bare count of container definitions (no ids/images/workflow names), so that tab strip is treated as chrome and unmasked — one attribute to flip if counts should stay masked.
2. `LoadingState` / `TabHeader` take copy as props, but every call site passes a fixed UI string, so both are unmasked at component level.
3. The "split your N-lines YAML file" notification and the deprecated-machine notification interpolate a per-account line count / workspace grace-period dates, so both are deliberately left masked.
4. Filter empty states never echo the search term, so they're unmasked; the filter inputs themselves are untouched.

**Verification:**
- Wrote a throwaway render test confirming every touched component actually forwards the attribute to the DOM (bitkit v1 and v2 primitives) — all passed, then removed (would need jest-config changes not present in this repo; happy to add if reviewers want it kept).
- `tsc --noEmit`: clean.
- ESLint on all 53 changed files: 0 errors (3 pre-existing warnings on master, verified by stashing).
- `npx jest`: 35 suites / 1431 tests pass, unchanged.
- Production build (`MODE=WEBSITE CLARITY=true`) succeeds with the attribute present in the emitted bundles.

### DIFF
--- a/jest.config.cjs
+++ b/jest.config.cjs
@@ -24,8 +24,11 @@ module.exports = {
     '@/(.*)': '<rootDir>/source/javascripts/$1',
   },
   // node_modules is transform-ignored by default; allow @bitrise/languageserver so the real codec
-  // (raw TS shipped by the git dependency) gets compiled.
-  transformIgnorePatterns: ['/node_modules/(?!@bitrise/languageserver/)', '\\.pnp\\.[^\\/]+$'],
+  // (raw TS shipped by the git dependency) gets compiled, and the two bitkit packages so component
+  // tests can render real bitkit components — bitkit v1 ships raw TS from `src/`, bitkit-v2 ships an
+  // ESM dist. Both barrels re-export a markdown component; a test that renders them should
+  // `jest.mock('react-markdown', ...)` rather than pull in that ESM dependency tree.
+  transformIgnorePatterns: ['/node_modules/(?!@bitrise/(languageserver|bitkit(-v2)?)/)', '\\.pnp\\.[^\\/]+$'],
   setupFiles: ['<rootDir>/spec/set-node-env.ts'],
   setupFilesAfterEnv: ['<rootDir>/spec/setup-jest.ts'],
   moduleDirectories: ['node_modules', '<rootDir>/spec/__mocks__'],
--- a/source/javascripts/components/Header.tsx
+++ b/source/javascripts/components/Header.tsx
@@ -360,10 +360,20 @@ const Header = () => {
       alignItems={['flex-start', 'center']}
     >
       <Box display="flex" alignItems="center" gap="8" minWidth={0}>
+        {/* The fixed crumbs are unmasked one by one on purpose — the project name between them is
+            customer data, so the Breadcrumb itself must stay masked. */}
         <Breadcrumb hasSeparatorBeforeFirst={isMobile}>
-          {isWebsiteMode && !isMobile && <BreadcrumbLink href="/dashboard">Bitrise CI</BreadcrumbLink>}
+          {isWebsiteMode && !isMobile && (
+            <BreadcrumbLink href="/dashboard" data-clarity-unmask="true">
+              Bitrise CI
+            </BreadcrumbLink>
+          )}
           {isWebsiteMode && appPath && appName && <BreadcrumbLink href={appPath}>{appName}</BreadcrumbLink>}
-          {(!isWebsiteMode || !isMobile) && <BreadcrumbLink isCurrentPage>CI configuration</BreadcrumbLink>}
+          {(!isWebsiteMode || !isMobile) && (
+            <BreadcrumbLink isCurrentPage data-clarity-unmask="true">
+              CI configuration
+            </BreadcrumbLink>
+          )}
         </Breadcrumb>
         {isWebsiteMode && <ConfigSettingsMenu />}
       </Box>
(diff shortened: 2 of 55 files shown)


moczolaszlo (commented):
Reviewed the whole diff plus the bitkit v1 component sources. Overall this is careful work — the default-mask + explicit `data-clarity-unmask` opt-in with subtree inheritance is the right model, and the masking boundary is drawn well (the delete-workflow / delete-step-bundle dialogs keep the id-bearing `<Text>` masked and only tag the static children — nicely done).

### Forwarding actually holds on v1 (the main risk)
The mechanism depends on ~15 distinct components each forwarding an unknown `data-*` down to the correct DOM node. I checked the touched bitkit v1 composites and they all spread it to their DOM root:

- `EmptyState` → `<Box {...rest}>`; `Notification` / `Ribbon` → `<ChakraAlert {...rest}>`; `Sidebar` → `<Box as="nav" {...boxProps}>`; `List` → `<ul {...rest}>`; `BreadcrumbLink` → `<a {...props}>`; `Dialog` → `<ModalContent {...rest}>`; `DialogFooter` → `<ModalFooter {...props}>`; `MenuList` → `<ChakraMenuList {...props}>`. Table/Tabs/Text/Button/Card are Chakra primitives.

Portals are handled correctly too: where content is portalled (`Dialog` → `ModalContent`, `MenuList`, `ActionMenu.Item`) the attribute sits on the portalled root itself, not on a parent that stays in the main tree — so inheritance still holds inside the portal. That's the easiest Clarity-specific mistake to make and it's avoided.

### Non-blocking, worth addressing

1. **Keep a forwarding/boundary regression test.** The whole thing is a silent convention: if forwarding breaks it silently re-masks (fail-safe), but if a future refactor drops a dynamic child under an unmasked container it silently *leaks* (fail-open). The PR says the render test was removed because it would need jest-config changes not present here — but the repo already runs `.spec.tsx` render tests with no extra config (e.g. `useIsYmlParseError.spec.tsx`), so a lightweight test asserting the attribute reaches the DOM for the wrapped primitives is feasible and worth keeping as the guard for exactly this load-bearing behavior.

2. **The Secrets delete `Dialog` is unmasked as a whole.** It's safe *today* — the body is fixed copy and never names the secret, unlike the other two delete dialogs — but tagging the entire `Dialog` (rather than its static children, as the workflow/step-bundle delete dialogs do) is the broadest fail-open surface here and the one inconsistent pattern. Bringing it in line with the "tag the static children only" approach would make the boundary uniform and refactor-proof.

Everything else looks good to merge.

bitrise-io-kolega:
Thanks for the thorough review! Addressed both points in `2c80b30d`:

1. Re-added a regression test (`source/javascripts/lib/ClarityUnmask.spec.tsx`) pinning every unmask site and asserting `data-clarity-unmask` reaches the DOM through the bitkit v1/v2 wrappers. Needed one narrow `transformIgnorePatterns` addition in `jest.config.cjs` (allowlisting the two bitkit packages) plus mocking `react-markdown`, which the barrel otherwise pulls in as ESM — no other config churn. Verified the guards actually catch regressions via mutation testing (dropped forwarding, widened an unmask boundary, etc. — all caught).
2. The Secrets delete dialog now tags its static children (title text, error notification, footer) individually instead of the whole `Dialog`, matching the pattern in the workflow/step-bundle delete dialogs.

Full suite is green (36/36 suites, 1436 tests), typecheck/lint clean.

moczolaszlo (approved):
Both follow-up points are addressed thoroughly in `2c80b30d`.

**Regression test (`ClarityUnmask.spec.tsx`)** — not a token test: it renders every tagged bitkit v1/v2 component and asserts the label sits *inside* the unmasked subtree (`getByText(...).closest('[data-clarity-unmask="true"]')`), with an exact element count that also catches an attribute landing on the wrong/extra node. It then pins the full file→count map of every unmask site (so a new or widened one can't land unreviewed), blocklists the customer-data modules, and rejects any non-`"true"` spelling that Clarity would silently ignore. Mutation-checked per the commit message. The `transformIgnorePatterns` widening is narrow and well justified.

**Secrets delete dialog** — now tags the body copy, error notification and footer individually instead of the whole `Dialog`, matching the delete workflow / delete step bundle dialogs; the title stays masked as a plain prop. Boundary is uniform now.

CI green, typecheck/lint clean. LGTM.

PR #1869, opened 2026-08-03
fix: surface located YAML parse errors and stop corrupting brace input values

bitrise-io-kolega:
### Scope note, please read first

The reported root cause was "the Visual tab uses a stricter YAML parser than the raw/CLI parser, so unquoted brace values like `{devs,qa}` are rejected in Visual mode but accepted by the CLI." I measured that before changing anything, and **it does not hold**.

A corpus of 43 unquoted brace/curly step-input values was run through both parsers:
- Visual-mode: the `yaml` package via `YmlUtils.toDoc`, with the real options (`stringKeys: true, keepSourceTokens: true`)
- Raw/CLI: `gopkg.in/yaml.v2` + `bitrise/v2 models.BitriseDataModel` + `Normalize()` — the exact path `apiserver/service/bitrise_config.go` uses

**Zero divergence.** Every value one accepts, the other accepts:
- `deploy_path: reports/{devs,qa}` → plain string in both
- `notify_user_groups: {devs,qa}` → flow **map** (not a string) in both, no error in either
- `{devs,qa}/reports`, `{devs,qa`, `{{ .Values.x }}`, `**/*.{js,ts}` → rejected by both

So there is no stricter-parser regression to align. This PR therefore does the two things that are actually broken, the first of which is the fallback the ticket itself specifies.

### 1. Parse failures had no location (the "silent/blank Visual tab")

`BitriseYmlStore` only ever checked `doc.errors.length === 0` and stored the raw string in `__invalidYmlString`, throwing the errors away. `InvalidYmlRedirect` then bounced the user off the visual editor, and the only feedback was a generic *"Please fix the errors in your YAML configuration before navigating."* — no line, no column, no cause. That is the reported "can't open Visual, fails silently" symptom.

Now: `YmlUtils.parseErrorOf()` captures the first error's message, code and 1-based line/column (plus file path in modular mode); the store keeps it in `__invalidYmlError` in lockstep with the existing sentinel; `useYmlParseError()` exposes it; the navigation toast and a new notification on the YAML page report e.g. `Line 7, column 31: Unexpected scalar at node end`.

### 2. `{devs,qa}` was silently corrupted to `[object Object]`

An unquoted `notify_user_groups: {devs,qa}` is a well-formed flow mapping per spec, so it arrives as an object. `StepVariableService.getValue` ran `String()` over it → the field displayed `[object Object]`, and editing that field **saved `[object Object]` over the user's value**. Now rendered as compact single-line YAML (`{devs: null, qa: null}`) via `YmlUtils.toInlineYml`.

### Tests
Added 30+ cases: the parser-parity corpus (so a future divergence fails CI), `parseErrorOf`/`formatParseError`, `toInlineYml`, `getValue` on flow map/seq values, and a `useYmlParseError` hook spec. Genuinely malformed YAML is asserted to still produce a located error rather than a silent failure.

### Verification
`npx jest` 1469/1469 pass · `tsc --noEmit` clean · `eslint` clean · `npm run build` succeeds · `go vet ./...` and `go test ./...` pass.

### Assumptions flagged for the reviewer
- The new notification is deliberately **not** `data-clarity-unmask`ed — in modular mode the message names a file from the user's repo, which the repo's own ClarityUnmask policy treats as configuration data.
- Touched files use legacy `@bitrise/bitkit` v1; I matched the surrounding code rather than doing a v1→v2 migration, to keep a bug fix reviewable. Happy to split a migration out if you want it.

### DIFF
--- a/source/javascripts/components/Header.tsx
+++ b/source/javascripts/components/Header.tsx
@@ -43,6 +43,7 @@ import useIsYmlParseError from '@/hooks/useIsYmlParseError';
 import usePushBranch, { PushBranchPayload } from '@/hooks/usePushBranch';
 import useSearchParams from '@/hooks/useSearchParams';
 import useYmlHasChanges from '@/hooks/useYmlHasChanges';
+import useYmlParseErrorMessage from '@/hooks/useYmlParseErrorMessage';
 import useYmlValidationStatus from '@/hooks/useYmlValidationStatus';
 import { usePipelinesPageStore } from '@/pages/PipelinesPage/PipelinesPage.store';
 import { useStepBundlesPageStore } from '@/pages/StepBundlesPage/StepBundlesPage.store';
@@ -75,6 +76,19 @@ const Header = () => {
   // YAML view on every schema-invalid load (SSW-3087).
   const ymlStatus = useYmlValidationStatus();
   const isParseError = useIsYmlParseError();
+
+  // Each of these points at the offending line when the config genuinely failed to parse, and falls
+  // back to the generic wording otherwise — `ymlStatus` also covers schema errors that have no
+  // location to name.
+  const switchToVisualErrorMessage = useYmlParseErrorMessage(
+    "YAML can't be parsed, please fix it before switching to the Visual editor.",
+    'switching to the Visual editor',
+  );
+  const saveErrorMessage = useYmlParseErrorMessage(
+    'Please fix the errors in your YAML configuration before saving.',
+    'saving',
+  );
+  const saveButtonErrorMessage = useYmlParseErrorMessage('YAML is invalid, please fix it before saving.', 'saving');
 
   const [path, navigate] = useHashLocation();
   const [searchParams] = useSearchParams();
@@ -334,7 +348,7 @@ const Header = () => {
       if (ymlStatus === 'invalid') {
         toast({
           title: 'YAML is invalid',
-          description: 'Please fix the errors in your YAML configuration before saving.',
+          description: saveErrorMessage,
           status: 'error',
           duration: null,
           isClosable: true,
@@ -381,7 +395,7 @@ const Header = () => {
       <BitkitTooltip
         disabled={!isParseError}
         placement={isMobile ? 'bottom' : 'bottom-start'}
-        text="YAML can't be parsed, please fix it before switching to the Visual editor."
+        text={switchToVisualErrorMessage}
       >
         <BitkitSegmentedControl
           size="sm"
@@ -429,7 +443,7 @@ const Header = () => {
         <Tooltip
           isDisabled={ymlStatus !== 'invalid'}
           placement={isMobile ? 'bottom' : 'bottom-start'}
-          label="YAML is invalid, please fix it before saving."
+          label={saveButtonErrorMessage}
         >
           <Button
             size="sm"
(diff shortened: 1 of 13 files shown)


moczolaszlo (dismissed):
Reviewed thoroughly — this is solid, well-tested work, and it builds cleanly on the parse-error fallback from #1863. Two things I verified specifically:

- **The `__invalidYmlError` lockstep invariant holds.** I checked every writer of `__invalidYmlString` / `__savedInvalidYmlString` in the store (initial state, `discardBitriseYmlDocument`, `commitActiveFileDocument`, both branches of `updateBitriseYmlDocumentByString`, both branches of `initializeBitriseYmlDocument`, `updateBitriseYmlDocument`, `activeDocumentPatch`, `modularTreePatch`) — each is paired with the corresponding error write, so `useIsYmlParseError` and `useYmlParseError` can never disagree.
- **The `toInlineYml` change is save-safe.** `getValue` feeds `StepInputGroup` as `value`/`defaultValue` only; the store is written solely via `onChange` on an actual edit, so an untouched flow-map input is never rewritten, and a re-saved `{devs: null, qa: null}` round-trips to the same map. No new corruption, and scalars are unchanged vs. the old `String(x ?? '')`.

The scope-note debunking the "stricter parser" premise is well-substantiated, and pinning the parser-parity corpus so a future divergence fails CI is a nice touch.

### Requested change (minor)

`toInlineYml` serializes with the library default schema (YAML 1.2 core):

```ts
return stringify(node, { flowCollectionPadding: false }).trim();
```

Everywhere else the app serializes as YAML 1.1 (`toYml` passes `version: '1.1', schema: 'yaml-1.1'`). It's display-only so the impact is small, but a 1.1 boolean-like key (e.g. `{on: null}`) can render differently here than the same value would through `toYml`. Please pass the same `version: '1.1', schema: 'yaml-1.1'` to this `stringify` call so the inline rendering matches the rest of the app.

### Non-blocking observation (no change needed)

`__invalidYmlError` / the new notification / the redirect all key off the active-or-root document only, so a load-time parse error in a *non-active* modular include won't get a located notification (it surfaces as markers in that file's own editor, and #1863 keeps it from crashing serialization). This is consistent with the existing `__invalidYmlString` scope — flagging it only so the boundary is a conscious one.

bitrise-io-kolega:
Thanks for the thorough review — good catch on the schema mismatch. Fixed: `toInlineYml` now serializes with the same `version: '1.1', schema: 'yaml-1.1'` options as `toYml`, so inline rendering matches what an actual save would write. Added a test asserting the two paths agree on quoting (flow padding is the one deliberate difference, since inline fields are always rendered compact). Full suite (1477 tests), typecheck, lint, build, and Go checks all pass. Pushed to the same branch.

moczolaszlo:
Follow-up to my review — one more consistency gap on the located-error work.

`formatParseError(...)` is only wired into the **nav-item toast** (`Navigation.tsx`, "…before navigating"). But the same generic, location-less wording still lives in three sibling spots, all keyed off `isParseError`:

- `Header.tsx:384` — the tooltip on the **Visual/YAML toggle**: *"YAML can't be parsed, please fix it before switching to the Visual editor."*
- `Header.tsx:337` — the Save toast: *"Please fix the errors in your YAML configuration before saving."*
- `Header.tsx:432` — the Save-button tooltip: *"YAML is invalid, please fix it before saving."*

The first one is the important one: the Visual/YAML toggle tooltip is exactly the moment the user is trying to switch to Visual and can't, so it's the most natural place to tell them *where* the parse error is — yet it stays generic after this PR, while the less-prominent nav toast gets the located message. That inconsistency is a bit surprising.

Could you also surface `formatParseError(useYmlParseError())` in the Visual-toggle tooltip (`Header.tsx:384`) so the located error shows up where users actually go to switch to Visual? Folding the two Save messages (`:337`, `:432`) into the same helper would be nice for consistency, but the toggle tooltip is the one I'd consider part of the fix rather than optional.

PR #1870, opened 2026-08-06
fix: notify Intercom on in-editor navigation so SPA tours re-target

violapeter:
## Problem
Per [Intercom product tours in Bitrise: what to watch out for](https://bitrise.atlassian.net/wiki/spaces/~7120205fa5090eaf5746519410986d2d5633fd/pages/5140512927) (§5): page-targeted Intercom tours in the Workflow Editor only ever fire for whichever page loads first in a session. Navigating in-editor (e.g. Workflows → Triggers) never triggers the second page's tour.

## Root cause
Intercom boots inside the WFE's own iframe document (`source/javascripts/lib/intrcm.js`) and detects SPA page views by hooking `history.pushState`/`replaceState` **on that same window**.

`useHashLocation`'s `navigate()` — the single entry point our hash router uses for in-editor navigation — only ever calls `window.parent.history.replaceState()`. It never touches this window's own History API, so Intercom never observes the navigation, and its "Current page URL contains…" targeting rules are evaluated once at initial load and never re-evaluated.

## Fix
Call `window.Intercom('update')` (guarded for when Intercom isn't present) at the end of `navigate()`, so Intercom re-evaluates page-targeting on every in-editor route change — its documented mechanism for SPAs it can't otherwise detect.

Also added a minimal `Window.Intercom` type declaration (none existed yet).

## Verification
- `tsc --noEmit`: clean.
- `eslint`: clean.
- No existing spec for `useHashLocation`; this is a two-line, low-risk addition (an optional-chained call), so didn't add a new test harness for it — happy to follow up if wanted.

Manual verification suggestion per the doc's reproduction: set up the two tours from §5 (Workflows-tab tour + Triggers-tab tour), load Workflows first, then navigate in-editor to Triggers — the Triggers tour should now fire.

### DIFF
--- a/source/javascripts/hooks/useHashLocation.ts
+++ b/source/javascripts/hooks/useHashLocation.ts
@@ -29,8 +29,7 @@ const useHashLocation: BaseLocationHook = () => {
       // parent's location — never call this window's History API, so Intercom can't see them and
       // its "Current page URL contains…" targeting rules are evaluated once, at first load, and
       // never again. hashchange on window.parent is the one thing every kind of navigation here
-      // has in common, so nudge Intercom from there rather than from navigate() alone. See:
-      // https://bitrise.atlassian.net/wiki/spaces/~7120205fa5090eaf5746519410986d2d5633fd/pages/5140512927
+      // has in common, so nudge Intercom from there rather than from navigate() alone.
       window.Intercom?.('update');
     };
 
--- a/source/javascripts/hooks/useHashLocation.ts
+++ b/source/javascripts/hooks/useHashLocation.ts
@@ -14,14 +14,6 @@ const navigate = (to: Path, { state }: Options = {}): void => {
   window.parent.location.hash = targetUrl.hash;
   window.parent.history.replaceState(state, '', targetUrl);
   window.DD_RUM?.startView(`/app/?/workflow_editor${targetUrl.hash?.split('?')?.[0] || '#!/workflows'}`);
-
-  // Intercom boots inside this iframe's own document and hooks pushState/replaceState on ITS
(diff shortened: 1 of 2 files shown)


tothszabi (dismissed):

source/javascripts/hooks/useHashLocation.ts
```diff
@@ -14,6 +14,14 @@ const navigate = (to: Path, { state }: Options = {}): void => {
   window.parent.location.hash = targetUrl.hash;
   window.parent.history.replaceState(state, '', targetUrl);
   window.DD_RUM?.startView(`/app/?/workflow_editor${targetUrl.hash?.split('?')?.[0] || '#!/workflows'}`);
+
+  // Intercom boots inside this iframe's own document and hooks pushState/replaceState on ITS
+  // window to detect SPA page views. This router only ever touches window.parent's history, so
+  // Intercom never sees in-editor navigation and its "Current page URL contains…" page-targeting
+  // rules are only evaluated once, at first load — later pages' tours never fire. Nudge it
+  // explicitly on every route change so it re-evaluates. See:
+  // https://bitrise.atlassian.net/wiki/spaces/~7120205fa5090eaf5746519410986d2d5633fd/pages/5140512927
```
gaborszakacs: If this page contains important context, let's move that into the git history (comment or commit message), if it's not required and the comment above is enough, I'd leave that out. I'm not a fan of referring to wiki pages from code, this is a link to a personal workspace page, it can break even easier.

gaborszakacs: (not a blocker on my part, just my 2 cents)

gaborszakacs (approved):

PR #1871, opened 2026-08-06
feat(tool setup): BE-2171 absolute latest and installed strategies

marcell-vida:
## Task

Using `installed` and `latest` special keywords without prefix in the YAML (e.g. `latest` instead of `22:latest`) was previously handled, by leaving the version selector on the UI in `Any` with `Latest version of` strategy (this dropdown was introduced in the current PR stack). Two new strategies `Latest preinstalled version` and `Latest released version` represents those states without a version field shown.

## Changes

- Create two new strategies: `Latest preinstalled version` and `Latest released version`
- `latest-of` keeps only the prefixed forms, so the version control disappears for the strategies
  that have no version, and the `Any` option goes with it
- Seed a prefix in the same write as the strategy when switching onto `latest-of`, preferring the
  prefix of the version being switched away from, so `22.12.0` offers `22` rather than upgrading
- Offer `latest-of` only where a prefix can be seeded
- Publish the mock catalogs newest first, like the real API, and add java and elixir to cover the
  non semver cases


source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -25,15 +25,27 @@ type ToolRowFormValues = {
...
 
+const STRATEGIES_WITH_VERSION_INPUT: VersionStrategy[] = ['exact', 'latest-released', 'latest-installed'];
+
 const OTHER_VALUE = '__other__';
 
 const TOOL_ID_COLUMN_WIDTH = rem(160);
 const VERSION_COLUMN_WIDTH = rem(240);
 
+/** A prefix strategy with no prefix serializes to the bare keyword, which parses back as absolute. */
+function isCollapsedPrefixOf(picked: VersionStrategy, parsed: VersionStrategy): boolean {
```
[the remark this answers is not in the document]
marcell-vida: Both commited and pending absolute rows would pass an empty version string from `getVersionInputValue`, so the guard would not hold up. I would not apply any complex fixes on this, because a simple remount solves this and writing the YAML requires switching to a separate tab, then switching back.

source/javascripts/components/StacksAndMachine/StackAndMachine.stories.tsx
```diff
@@ -0,0 +1,166 @@
+import { Box } from '@chakra-ui/react/box';
```
ofalvai, line 1: Why this new story file? Maybe I missed something

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -24,138 +24,161 @@ import { paths } from '@/routes';
...
+
+const ToolVersions = ({ workflowId, stackReportUrl, isReadOnly }: Props) => {
   const scope: ToolScope = workflowId ? { type: 'workflow', workflowId } : { type: 'root' };
   const tools = useToolsForScope(scope);
   const { replace } = useNavigation();
   const { data: catalog, isLoading: isCatalogLoading, isError: isCatalogError } = useToolCatalog();
   const [hasPendingRow, setHasPendingRow] = useState(false);
-  const [pendingStrategy, setPendingStrategy] = useState<VersionStrategy>('latest-released');
+  // A new row starts on the absolute variant, because with no prefix entered yet the prefix variant
+  // serializes to bare `latest` and the dropdown would change by itself once the row committed.
```
ofalvai, line 51: Are you sure we want to change the default? This reasoning sounds like a bug or a code smell, it should not dictate a product decision (default value).

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -23,17 +23,36 @@ type ToolRowFormValues = {
...
+  'latest-released': 'A major or minor version is required',
+  'latest-installed': 'A major or minor version is required',
+};
+
 const OTHER_VALUE = '__other__';
 
 const TOOL_ID_COLUMN_WIDTH = rem(160);
 const VERSION_COLUMN_WIDTH = rem(240);
 
+/** A prefix strategy with no prefix serializes to the bare keyword, which parses back as absolute. */
```
ofalvai, line 48: I might not fully understand the whole picture, but isn't this working around something that's fundamentally wrong we should be fixing instead?

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -56,13 +75,18 @@ const ToolRow = ({
...
-  onStrategyChange,
-  onVersionChange,
+  onChange,
   onRemove,
 }: ToolRowProps) => {
   // Whether the user has explicitly picked "Other" from the tool ID dropdown.
   const [manualOther, setManualOther] = useState(false);
+  // Keeps the dropdown on the user's pick while the prefix field is still empty. Honoured only in
+  // that one ambiguous case, so any other change to the YAML still wins.
+  const [pickedStrategy, setPickedStrategy] = useState<VersionStrategy>();
```
ofalvai, line 87: Related to my above comment, it feels a bit "smelly". We are manually tracking state which should be tracked by the dropdown component I think. Other than the added complexity, it might be a sign that we are doing something fundamentally wrong.

source/javascripts/components/ToolVersions/ToolVersions.tsx
marcell-vida: To be fair the default didn't actually change, in fact, it stayed the same. Previously if user simply created a new line and "latest released" was activated it basically wrote `latest` to the yaml in case user didn't fill the third field with prefix. Now if user creates a new line, it does the same thing, since "absolute latest" now means `latest` in the yaml. We can change it back to default to the other strategy, but keep in mind, that now means the third field is required. So if user would add a new tool row they will instantly face a required field error (same behavior as before with other strategies).

source/javascripts/components/StacksAndMachine/StackAndMachine.stories.tsx
marcell-vida: In the merged-config tab the tool setup should be read-only, which was previously not respected and it looked like some bug doesn't allow editing the config. This story allows us to test the `useReadOnlyView` function instead of just passing a boolean. To achieve this I created the story for `StackAndMachine`, which was previously not present.

source/javascripts/components/ToolVersions/ToolRow.tsx
marcell-vida: It is a workaround, but I don't think there is a real "solution" for that. Unfortunately with the absolute strategies our system is designed so, that `latest` keyword can mean two states:
- user selected `Latest released version of` AND have not written a prefix in the third input yet
- user selected `Absolute latest released`

#### Previous solution

1. User selects `Latest released version`
2. yaml gets written: `golang: latest`
3. User writes prefix: `1.23`
4. yaml gets written: `golang: 1.23:latest`

#### Current solution without the workaround would be

1. User selects `Latest released version of...`
2. yaml gets written: `golang: latest`
3. UI instantly jumps over to `Absolute latest released`, since latest keyword means that. User doesn't even have time to write a prefix

With state handling this can be avoided for the time of the edits. Also this is why prefix became required field for `Latest released version of...` strategy, since `tool: latest` now means an entirely different strategy (on UI). Same applies to installed strategies.

Why I still think we should stand by the new strategies: having an optional prefix field can confuse users, if they only want the real latest version. The new approach seems to be much cleaner (for me).

**In short: we cannot write/read the yaml until we decide which strategy it will be, the workaround is handling the temporary state.**

source/javascripts/components/ToolVersions/ToolRow.tsx
marcell-vida: I tried to explain this in the other comment, please read that before this.

The problem is a bit different, since the state of the dropdown is handled by the 'single source of truth' yaml, and the component is entirely managed. The state should be owned here in my opinion, as fetching the list elements, reading the yaml values and writing the changes happens here as well. The dropdown has no connection to these and the problem is more related to modelling.

PR #1876, opened 2026-08-13
Enable stack rollback option for machine resource classes

bitrise-io-kolega:
## Summary
The "Use previous stack version" checkbox only lit up when an exact machine type was selected, because the stack payload's `rollback_version` map is keyed by exact machine type IDs, so a selected machine resource class never matched.

A machine resource class can schedule a build onto any of the exact machine types it stands for, so the rollback stack version has to be published for all of them — otherwise a build could land on a machine where that version doesn't exist. Two new API fields make that decidable: `available_on_machines` (per-exact-machine-type `latest_version`/`rollback_version`, keyed by availability group) and `machine_types` on the machine payload (the exact machine types a resource class stands for).

`StackAndMachineService.getAvailableRollbackVersion()` (new pure function in `core/services`, consumed by `StackAndMachine.tsx` in place of the previous inline lookup) offers a rollback only when every exact machine type of the selected resource class has BOTH versions published and they all agree on the same rollback version, for the selected pricing tier. A missing `latest_version` means the stack isn't available on that machine type at all, which blocks the rollback too.

## Tests
12 new cases in `StackAndMachineService.spec.ts` covering: exact machine type (existing behaviour), resource class where all exact machine types support rollback (enabled), one member missing `rollback_version` (disabled), stack not available on a member at all (disabled), members disagreeing on the rollback version (disabled), tier selection, the legacy-payload fallback, and workspace-specific availability groups being ignored. Full suite: 1454 tests / 37 suites pass, `tsc` clean, `eslint` clean. MSW mocks extended so both the enabled and the disabled resource-class case are reproducible in Storybook.

## Assumptions (worth a reviewer's eye)
1. Exact machine types still read the legacy `rollback_version` map as a fallback, so the checkbox can't regress if this ships before the backend serves the new fields.
2. Members of a resource class sitting on DIFFERENT `latest_version`s do not block the rollback as long as their `rollback_version` matches. Verified against the real production stacks config: a handful of (stack, tier, resource class) combos are exactly like this today, and rolling them back is safe — requiring identical `latest_version` would needlessly disable them.
3. Only the free & paying pricing tiers are consulted (unchanged from today); workspace-specific availability groups are ignored, so a workspace whose stack availability exists only under its own key (e.g. private cloud) gets the checkbox disabled for resource classes. Flagging this as a scope call for the reviewer — extending it to workspace-specific availability would be a behaviour change for exact machine types too, and is left as a possible follow-up.

## Depends on
The backend fields this relies on (`available_on_machines`, `machine_types`) are added in a companion backend PR. This PR is backwards compatible either order thanks to the legacy-map fallback, but the checkbox won't actually enable for resource classes until the backend change is deployed.

### DIFF
--- a/source/javascripts/core/services/StackAndMachineService.spec.ts
+++ b/source/javascripts/core/services/StackAndMachineService.spec.ts
@@ -1914,6 +1914,218 @@ describe('StackAndMachineService', () => {
     });
   });
 
+  describe('getAvailableRollbackVersion', () => {
+    const createStack = (override: Partial<Stack>): Stack => ({
+      id: 'osx-xcode-16.0.x',
+      name: 'Xcode 16.0.x',
+      status: 'stable',
+      description: '',
+      machineTypes: ['g2.mac.medium', 'm2.medium'],
+      os: 'macos',
+      ...override,
+    });
+
+    const createMachineType = (override: Partial<MachineType>): MachineType => ({
+      id: 'm2.medium',
+      name: 'M2 Medium',
+      os: 'macos',
+      isDisabled: false,
+      availableInRegions: {},
+      ...override,
+    });
+
+    describe('when an exact machine type is selected', () => {
+      it('returns the rollback version of the machine type', () => {
+        const stack = createStack({
+          availableOnMachines: {
+            paying: { 'm2.medium': { latestVersion: '2-83-0', rollbackVersion: '2-82-0' } },
+          },
+        });
+
+        expect(
+          StackAndMachineService.getAvailableRollbackVersion({
+            stack,
+            machineType: createMachineType({ id: 'm2.medium' }),
+            tier: 'paying',
+          }),
+        ).toBe('2-82-0');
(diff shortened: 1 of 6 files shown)


source/javascripts/core/services/StackAndMachineService.spec.ts
```diff
@@ -1914,6 +1914,237 @@ describe('StackAndMachineService', () => {
...
+        expect(
+          StackAndMachineService.getAvailableRollbackVersion({
+            stack,
+            machineType: machineResourceClass,
+            tier: 'paying',
+          }),
+        ).toBe('');
+      });
+
+      it('ignores workspace-specific machine availability', () => {
```
erosdome: It should not ignore workspace-specific machine availability. Though right now I'm not seeing traces of workspace-specific stuff in the pre-diff code - is it present there in some way?

erosdome (changes requested):

PR #1878, opened 2026-08-13
fix(tool setup): honour read-only views in tool setup, guard non-string versions

marcell-vida:
## Task

This PR contains two bug fixes:
- `StackAndMachine` already computes `isReadOnlyView`, which is true on the merged config preview,
on a file that comes from another repository (or the caller can force it). That value never
reached tool setup, so the rows stayed interactive while `updateBitriseYmlDocument` dropped every
mutation. Edits looked accepted and then reverted: a new row vanished because the pending row was
cleared even though the write was ignored, and picking a strategy snapped back to the stored value.
- `Tools` is typed `Record<string, string>`, but hand written YAML can hold
`python: 3.13`, which parses as a number. `parseToolVersion` then threw
`TypeError: raw.toLowerCase is not a function` and the whole tool setup section failed to render.

## Changes

- Pass `isReadOnlyView` down to tool setup
- Mark the fields `readOnly` rather than `disabled`, so values stay legible and keyboard reachable
- Hide `Add new` and disable row removal in read only views
- Reuse the tooltip copy of the stack selectors above, so both halves of the card read the same
- Separate empty state copy for read only views
- Coerce non string values at the top of `parseToolVersion`, with cases for number, null, undefined
- New `StackAndMachine` story file: editable, modular editable, forced read only, merged config,
  cross repository file, read only expandable

### DIFF
--- a/source/javascripts/components/ToolVersions/ToolRow.tsx
+++ b/source/javascripts/components/ToolVersions/ToolRow.tsx
@@ -5,6 +5,7 @@ import {
   BitkitLink,
   BitkitSelect,
   BitkitTextInput,
+  BitkitTooltip,
   IconMinusCircle,
   IconOpenInNew,
   rem,
@@ -30,6 +31,8 @@ const STRATEGY_LABELS: Record<VersionStrategy, string> = {
 };
 
 const OTHER_VALUE = '__other__';
+
+const READ_ONLY_TOOLTIP_TEXT = 'To edit, switch to the module file that defines it.';
 
 const TOOL_ID_COLUMN_WIDTH = rem(160);
 const VERSION_COLUMN_WIDTH = rem(240);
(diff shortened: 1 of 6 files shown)


ofalvai (dismissed):

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -231,7 +237,13 @@ const ToolRow = ({
           </Box>
         )}
 
-        <BitkitIconButton variant="tertiary" icon={IconMinusCircle} label="Remove tool" onClick={onRemove} />
+        <BitkitIconButton
```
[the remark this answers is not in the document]
ofalvai: When in read-only mode, it's not possible to add a new row anyway, it's not possible to end up in this state.

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -80,82 +86,94 @@ const ToolVersions = ({ workflowId }: { workflowId?: string }) => {
         )}
       </Stack>
 
-      <Stack gap="16">
-        {Object.entries(tools).map(([toolId, versionString]) => {
-          const parsed = ToolsService.parseToolVersion(versionString);
-          const versionValue =
-            parsed.strategy === 'exact' ? parsed.version : parsed.strategy === 'unset' ? '' : (parsed.prefix ?? '');
-          return (
+      <BitkitTooltip text="Read-only here — edit it in the module file that defines it." disabled={!isReadOnly}>
```
k-olly: I think the tooltip should point on the read only selects

k-olly: Nit: users already see that the selects are read-only, I think that state shouldn't be in the tooltip text. I would say something like this:
To edit, switch to the module file that defines it.

source/javascripts/components/ToolVersions/ToolVersions.tsx
```diff
@@ -80,82 +86,94 @@ const ToolVersions = ({ workflowId }: { workflowId?: string }) => {
...
+              />
+            );
+          })}
+          {hasPendingRow && (
             <ToolRow
-              key={toolId}
-              toolId={toolId}
-              strategy={parsed.strategy}
-              version={versionValue}
+              toolId=""
```
k-olly: why don't we pass the toolId?

source/javascripts/core/services/ToolsService.spec.ts
```diff
@@ -53,6 +53,12 @@ describe('ToolsService', () => {
       expect(ToolsService.parseToolVersion(':latest')).toEqual({ strategy: 'exact', version: ':latest' });
     });
 
+    it('tolerates non-string values from hand-edited YAML', () => {
+      expect(ToolsService.parseToolVersion(3.13 as unknown as string)).toEqual({ strategy: 'exact', version: '3.13' });
```
k-olly: Without the type assertions we have a cleaner code:  

expect(ToolsService.parseToolVersion(3.13)).toEqual({ strategy: 'exact', version: '3.13' });
 expect(ToolsService.parseToolVersion(null)).toEqual({ strategy: 'exact', version: '' });
      
expect(ToolsService.parseToolVersion(undefined)).toEqual({ strategy: 'exact', version: '' });

source/javascripts/components/ToolVersions/ToolVersions.tsx
marcell-vida: This is the pending row, which has no tool selected yet. Picking a tool commits it to the YALM, after that the row renders from tools along with the other committed rows above (keyed by its ID).

k-olly (approved):

PR #1881, opened 2026-08-13
feat(tool setup): BE-2172 merge latest strategies with prefix dropdown

marcell-vida:
## Task

The strategies meant the following:

- `latest-installed`: use latest version (strating with the optional prefix) installed if possible, or lateset released
- `latest-released`: use latest released version (starting with the optional prefix)

Since the fallback of the installed strategy is the other strategy, they overlap almost entirely. They also took the same prefix, so switching between them had to carry that prefix over.

## Changes

- Merge the two strategies into `latest-of`
- Expose the flag as a checkbox labelled `Prefer pre-installed version`
- Create list of prefixes from the catalog
- Add a dropdown UI element if prefix list is available
- `setTool` takes a parsed version instead of a strategy plus a loose value, so the flag has a home
- A value with no separator stands in for itself, so a list of channel names still suggests something
- Delete `nextVersionOnStrategyChange`, obsolete once the prefix move became a checkbox
- Export `serializeToolVersion` so the round trip can be asserted directly
- Versions are cut at separators for prefixes
- The option builders take the catalog alone and `withConfiguredValue` injects the value from the YAML, so picking one does not sort the catalog again
- Closest prefix suggestions are stopped at minor, since a deeper cut names one release rather than a line to follow and lists are ordered newest first
- Prefixes match at a separator boundary (which is what the default resolver does) e.g. `24.2` matches `24.2.1` but not `24.20.1`, and `2` matches nothing in a catalog of `22.x`
- The CLI's parsing is mirrored: the keyword is read as a suffix and the value is trimmed, so `a:b:latest` has prefix `a:b` and `:latest` means bare `latest`
- Storybook gains an editable YAML panel beside the rows, so a story can be driven from either side

**Two deliberate divergences from the CLI**

- The `asdf` provider matches the raw string rather than at a boundary, so a prefix straddling a separator resolves differently under `tool_config.provider: asdf`
  - The UI follows mise, which is the default
- Keyword matching here ignores case while the CLI's `(.*):latest$` does not, so `22:LATEST` renders as a prefixed keyword but is an exact version to the CLI
  - The lenient reading matches intent, so the fix belongs in the CLI

These are noted in `docs/domain.md`.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -23,13 +24,14 @@ type ToolRowFormValues = {
...
-  'latest-released': 'Latest released version',
-  'latest-installed': 'Latest preinstalled version',
+  'latest-of': 'Latest version of',
   exact: 'Exact version',
   unset: 'Do nothing (unset global setting)',
 };
 
 const OTHER_VALUE = '__other__';
+/** No prefix, i.e. the newest version overall. A sentinel, because '' means "nothing selected". */
+const ANY_PREFIX_VALUE = '__any__';
```
marcell-vida, line 36: Temporary solution, while absolute latest strategies get merged in a separate PR (built on top of this).

source/javascripts/core/models/Tools.ts
```diff
@@ -1,8 +1,11 @@
-export type VersionStrategy = 'latest-released' | 'latest-installed' | 'exact' | 'unset';
+export type VersionStrategy = 'latest-of' | 'exact' | 'unset';
 
 export type ParsedToolVersion =
-  | { strategy: 'latest-released'; prefix?: string }
-  | { strategy: 'latest-installed'; prefix?: string }
+  /**
+   * The newest version matching `prefix`, e.g. `22:latest`. `installed` picks the `:installed`
```
ofalvai: Is it just me who struggles to understand this comment? Also, we might not even need to explain anything about the *runtime behavior* here in the workflow editor code.

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -38,29 +40,30 @@ type ToolRowProps = {
   toolId: string;
   strategy: VersionStrategy;
   version: string;
+  /** Only meaningful for `latest-of`: resolve against preinstalled versions where possible. */
+  installed?: boolean;
```
ofalvai: nit: maming idea: `prefer_installed`.

source/javascripts/components/ToolVersions/ToolRow.tsx
```diff
@@ -145,17 +157,25 @@ const ToolRow = ({
...
       setVersionTouched(false);
-    } else if (newStrategy === 'exact' && newVersion.trim() === '') {
+    } else if (newStrategy === 'exact') {
       // The field was already empty, so it won't hit the branch above -> flag it immediately
       // since it's already invalid.
       setVersionTouched(true);
     }
-    onStrategyChange(newStrategy, newVersion);
+    onChange(ToolsService.toParsedToolVersion(newStrategy, '', installed));
+  };
```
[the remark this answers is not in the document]
marcell-vida: This is already covered. `versionTouched` only matters when strategy is 'exact', and switching back there always flags it again with the existing else if.

moczolaszlo (changes requested):
1. Is cutting at every separator the intended granularity? On the live catalog java's 6165 versions
   yield 8739 prefix options — more options than versions — including
   `semeru-jre-openj9-8u482-b08.1_openj9-0.57`, which isn't a version line. python is the case that
   worries me more: `3.1` ends up in the list (there are 3.1.x releases), but prefixes match as
   strings, so `3.1` resolves to `3.16-dev`. The dropdown is suggesting a value that means something
   different from what it reads as, which the free-text field on master never did. If the `-` cut is
   there for java's `zulu-musl-8` shapes, fair enough — but then it may be worth capping the depth, or
   only deriving from the semver-shaped values. Your call which way; I don't know the resolver well
   enough to pick one.

2. Not live yet, so this is a heads-up rather than an ask: on its own this PR makes the prefix control
   a `BitkitSelect` with no search for catalog-known tools, where master had a free-text input. Until
   #1883 lands that means an 8739-item dropdown for java with no way to filter it, and no way to enter
   a prefix the list doesn't offer unless it's already in the YAML. Fine if the stack merges together;
   worth knowing if this one goes to master on its own.

3. `onChange` on `BitkitCheckbox` resolves to the label's DOM handler — `BitkitCheckboxProps` extends
   `Checkbox.RootProps`, i.e. `HTMLChakraProps<"label", …>` — which is why the target needs
   `as unknown as HTMLInputElement`. `onCheckedChange` is the Ark prop and needs no cast:

   ```tsx
   // instead of
   onChange={(e) => handlePreferInstalledChange((e.target as unknown as HTMLInputElement).checked)}

   // details.checked is `boolean | "indeterminate"`, hence the `=== true`
   onCheckedChange={({ checked }) => handlePreferInstalledChange(checked === true)}
   ```

4. The checkbox state doesn't survive a round trip through another strategy: `latest-of` → `exact` →
   `latest-of` drops `preferInstalled`, since `exact` has nowhere to keep it. Consistent with the
   prefix being cleared, so fine if it's deliberate.

5. `ToolVersionsYamlPreview.stories.tsx` duplicates `ToolVersions.stories.tsx`'s meta — msw handlers,
   args, decorator — to add the YAML panel. Would it work as a decorator on the existing file, so the
   setup lives in one place?

6. Nit: `versionOptions`' memo depends on `version`, but `getVersionOptions` only uses it to unshift a
   value the catalog doesn't list — so every selection re-runs `semver.rcompare` over the whole list,
   6165 entries for java. Splitting the two halves keeps the sort across a pick:

   ```tsx
   // '' means "nothing to inject", so this half only depends on the catalog
   const catalogVersionOptions = useMemo(
     () => (isExactKnownTool ? ToolsService.getVersionOptions(toolVersions, '') : []),
     [isExactKnownTool, toolVersions],
   );
   const versionOptions = useMemo(
     () =>
       version && !catalogVersionOptions.some(({ value }) => value === version)
         ? [{ value: version, label: version }, ...catalogVersionOptions]
         : catalogVersionOptions,
     [catalogVersionOptions, version],
   );
   ```

   `getPrefixOptions` carries the same "inject the configured value at the top" logic, so if you'd
   rather keep it in the service, pulling it out as one shared helper covers both call sites.

7. Please bring the description up to date: it says the checkbox is labelled
   `Use installed versions if possible` where the code says `Prefer pre-installed version`, and it
   doesn't mention `ToolVersionsYamlPreview.stories.tsx` at all.

*AI-assisted review: the digging through source is Claude's, the judgement and every ask here are mine.
Not convinced by a point? Ask — I'll show the working.*

moczolaszlo (changes requested):
The prefix rework holds up: I re-derived the suggestions from the live catalog, and python's `3.1` now
lands on `3.1.5` instead of `3.16-dev`, while java drops from 8739 suggestions to 2807. I also walked
the depth cap across erlang `28.5.0`, java `26.0.2`, ruby `4.0.0` and elixir `1.20.4-otp` — major.minor
is the meaningful line in all four, so stopping there is the right call.

1. The description still describes the previous design. It says the checkbox is labelled
   `Use installed versions if possible` where the code says `Prefer pre-installed version`, and it
   doesn't mention what is now the substance of the PR: the separator-boundary prefix rule, the depth
   cap and numeric ordering of the suggestions, `withConfiguredValue`, the move from a colon split to a
   greedy keyword suffix, the `trim()`, the story consolidation, or the two deliberate CLI divergences
   you documented in `docs/domain.md`. Those last two are product-visible — the asdf provider matching
   prefixes as raw strings, and `22:LATEST` reading as a keyword here but as an exact version to the
   CLI — and someone reviewing from the PR alone won't know either exists.

2. `toPrefixes`' comment says "`miniconda3-3.9` is a line, `miniconda3-3` is not", but the function
   emits both: python's `miniconda3-4.7.12` yields `miniconda3`, `miniconda3-4` and `miniconda3-4.7`,
   and `miniconda3-4` does name a line — it covers 4.6.x and 4.7.x. I take the point to be why the
   regex is anchored rather than which cuts survive, so it's worth rewording to not read as a claim
   about the output.

3. The one thing I can't check from here: the parser now mirrors specific CLI internals — the
   `strings.TrimSpace` opening `ParseVersionString`, the greedy end-anchored `(.*):latest$`, and the
   handoff to `mise latest <tool>@<prefix>`. Was that read off the CLI source? It's the load-bearing
   claim behind the rewrite and the only one the review can't verify.

*AI-assisted review: the digging through source is Claude's, the judgement and every ask here are mine.
Not convinced by a point? Ask — I'll show the working.*

marcell-vida:
> 1. Is cutting at every separator the intended granularity? On the live catalog java's 6165 versions
>    yield 8739 prefix options — more options than versions — including
>    `semeru-jre-openj9-8u482-b08.1_openj9-0.57`, which isn't a version line. python is the case that
>    worries me more: `3.1` ends up in the list (there are 3.1.x releases), but prefixes match as
>    strings, so `3.1` resolves to `3.16-dev`. The dropdown is suggesting a value that means something
>    different from what it reads as, which the free-text field on master never did. If the `-` cut is
>    there for java's `zulu-musl-8` shapes, fair enough — but then it may be worth capping the depth, or
>    only deriving from the semver-shaped values. Your call which way; I don't know the resolver well
>    enough to pick one.

The dropdown was good, `3.1` really did resolve to `3.1.5`, but `isPrefixInCatalog` worked incorrectly. In the background mise gets the prefix untouched and matches part by part, the CLI's own filter is okay, it also has a test pinning that 18.1 does not match 18.10.0. Implemented your suggestions. Result dropped a lot of options, Java `8739` -> `2807` python `789` ->`348`. It is now also asserted that every prefix offered actually matches. asdf provider works a bit different: it uses an unbounded HasPrefix, so a prefix straddling a boundary resolves differently there. Written up as a trap in `docs/domain.md` rather than modelled.

> 2. Not live yet, so this is a heads-up rather than an ask: on its own this PR makes the prefix control
>    a `BitkitSelect` with no search for catalog-known tools, where master had a free-text input. Until
>    #1883 lands that means an 8739-item dropdown for java with no way to filter it, and no way to enter
>    a prefix the list doesn't offer unless it's already in the YAML. Fine if the stack merges together;
>    worth knowing if this one goes to master on its own.

#1883 adds search to both controls, so I would rather not conflict with it. We will only start Beta testing the feature once all of the PRs in the stack get merged.

> 3. `onChange` on `BitkitCheckbox` resolves to the label's DOM handler — `BitkitCheckboxProps` extends
>    `Checkbox.RootProps`, i.e. `HTMLChakraProps<"label", …>` — which is why the target needs
>    `as unknown as HTMLInputElement`. `onCheckedChange` is the Ark prop and needs no cast:
>    ```tsx
>    // instead of
>    onChange={(e) => handlePreferInstalledChange((e.target as unknown as HTMLInputElement).checked)}
>    
>    // details.checked is `boolean | "indeterminate"`, hence the `=== true`
>    onCheckedChange={({ checked }) => handlePreferInstalledChange(checked === true)}
>    ```

Thanks for noticing, I modified it.

> 4. The checkbox state doesn't survive a round trip through another strategy: `latest-of` → `exact` →
>    `latest-of` drops `preferInstalled`, since `exact` has nowhere to keep it. Consistent with the
>    prefix being cleared, so fine if it's deliberate.

It is deliberate, same as the prefix being cleared on that switch, since exact has no resolution intent to preserve. Rename keeps the flag on purpose.

> 5. `ToolVersionsYamlPreview.stories.tsx` duplicates `ToolVersions.stories.tsx`'s meta — msw handlers,
>    args, decorator — to add the YAML panel. Would it work as a decorator on the existing file, so the
>    setup lives in one place?

Yes, panel was moved to `storyutils/` and the meta switches on a yamlPreview parameter. It is editable now as well, so you can type YAML and watch the rows react. I also found a bug with the changes: set() on a shallow copy writes into the shared nested workflows, so two stories scoped to a workflow overwrote each other.

> 6. Nit: `versionOptions`' memo depends on `version`, but `getVersionOptions` only uses it to unshift a
>    value the catalog doesn't list — so every selection re-runs `semver.rcompare` over the whole list,
>    6165 entries for java. Splitting the two halves keeps the sort across a pick:
>    ```tsx
>    // '' means "nothing to inject", so this half only depends on the catalog
>    const catalogVersionOptions = useMemo(
>      () => (isExactKnownTool ? ToolsService.getVersionOptions(toolVersions, '') : []),
>      [isExactKnownTool, toolVersions],
>    );
>    const versionOptions = useMemo(
>      () =>
>        version && !catalogVersionOptions.some(({ value }) => value === version)
>          ? [{ value: version, label: version }, ...catalogVersionOptions]
>          : catalogVersionOptions,
>      [catalogVersionOptions, version],
>    );
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    `getPrefixOptions` carries the same "inject the configured value at the top" logic, so if you'd
>    rather keep it in the service, pulling it out as one shared helper covers both call sites.

Took the service option to cover both call sites. The builders take the catalog alone and `withConfiguredValue` injects the configured value. The two branches are mutually exclusive, so it ended up as one list rather than two.

> 7. Please bring the description up to date: it says the checkbox is labelled
>    `Use installed versions if possible` where the code says `Prefer pre-installed version`, and it
>    doesn't mention `ToolVersionsYamlPreview.stories.tsx` at all.

Of course, rewrote it with the updates.

marcell-vida:
> The prefix rework holds up: I re-derived the suggestions from the live catalog, and python's `3.1` now lands on `3.1.5` instead of `3.16-dev`, while java drops from 8739 suggestions to 2807. I also walked the depth cap across erlang `28.5.0`, java `26.0.2`, ruby `4.0.0` and elixir `1.20.4-otp` — major.minor is the meaningful line in all four, so stopping there is the right call.
> 
> 1. The description still describes the previous design. It says the checkbox is labelled
>    `Use installed versions if possible` where the code says `Prefer pre-installed version`, and it
>    doesn't mention what is now the substance of the PR: the separator-boundary prefix rule, the depth
>    cap and numeric ordering of the suggestions, `withConfiguredValue`, the move from a colon split to a
>    greedy keyword suffix, the `trim()`, the story consolidation, or the two deliberate CLI divergences
>    you documented in `docs/domain.md`. Those last two are product-visible — the asdf provider matching
>    prefixes as raw strings, and `22:LATEST` reading as a keyword here but as an exact version to the
>    CLI — and someone reviewing from the PR alone won't know either exists.

Yes, they are the product visible part and neither is discoverable from the diff, description is rewritten. One thing to note: `ToolVersionsYamlPreview.stories.tsx` only existed on the current branch, so there is nothing to consolidate from the diff's point of view. The bullet says what Storybook gains instead.

> 2. `toPrefixes`' comment says "`miniconda3-3.9` is a line, `miniconda3-3` is not", but the function
>    emits both: python's `miniconda3-4.7.12` yields `miniconda3`, `miniconda3-4` and `miniconda3-4.7`,
>    and `miniconda3-4` does name a line — it covers 4.6.x and 4.7.x. I take the point to be why the
>    regex is anchored rather than which cuts survive, so it's worth rewording to not read as a claim
>    about the output.

Good catch! The comment was wrong. miniconda3-4 is emitted and does name a line. Reworded to be about anchoring.

> 3. The one thing I can't check from here: the parser now mirrors specific CLI internals — the
>    `strings.TrimSpace` opening `ParseVersionString`, the greedy end-anchored `(.*):latest$`, and the
>    handoff to `mise latest <tool>@<prefix>`. Was that read off the CLI source? It's the load-bearing
>    claim behind the rewrite and the only one the review can't verify.

Yes, these are the sources:
- `TrimSpace` opening `ParseVersionString`, `toolprovider/version.go:14`
- Patterns, `models/toolprovider.go:28`
- `tools:` entry point and the mise default, `toolprovider/config.go:41` and `:68`
- Handoff, `toolprovider/mise/resolve.go:52`
- Bounded filter and its test, `toolprovider/list_versions.go:32` and `list_versions_test.go:135`

Worth mentioning: `list_versions.go` is for a bitrise tools versions list command, not part of the resolver. It performs a `TrimRight(prefix, ".")`, but at install time it is Mise's authority to handle the prefix, therefore the CLI does no matching of its own.

moczolaszlo (approved):

PR #1882, opened 2026-08-14
feat(tool setup): BE-2187 resolve latest version

marcell-vida:
## Task

A prefix says which versions are acceptable, not which one you get, so the row could not answer
"what runs today?". It now resolves the configured prefix against the catalog and shows the result.

Resolution mirrors what Bitrise does: the first catalog entry that starts with the prefix. The
catalog is published newest first, numerically, so the first match is the highest one. This is
deliberately not semver range matching, which would disagree with the resolver for anything that is
not clean semver, and would exclude most of java, python, ruby and erlang entirely.

## Changes

- `getLatestVersion` returns the first version starting with the prefix, or the newest without one
- Show `Currently resolves to 22.12.0` under the prefix
- Widen the version list fetch to any catalog known tool, not only rows already showing a version
- No hint for the installed variant, the catalog lists released versions only
- No hint for tools outside the catalog, there is no version list to resolve against

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


moczolaszlo (changes requested):
1. The hint doesn't follow the rule the description states, and it's wrong on the live catalog.
   `getLatestVersion` searches `getVersionOptions(...)`, whose ordering is a display choice — semver
   sorted first, everything else appended — so every version that isn't clean 3-part semver is
   demoted below every version that is. This has nothing to do with pre-releases: erlang's and java's
   4-part builds are not valid semver, so

   | tool | prefix | first catalog match | what the hint shows |
   |---|---|---|---|
   | erlang | `28` | `28.5.0.6` | `28.4.3` |
   | erlang | `27.3` | `27.3.4.17` | `27.3.4` |
   | java | `26` | `26.0.2.1` | `26.0.2` |
   | python | (Any) | `3.16-dev` | `3.14.7` |
   | ruby | (Any) | `4.1-dev` | `4.0.6` |

   erlang `28` is a whole minor out. Iterating `toolVersions.versions` with `matchesPrefix` is both
   what the description says and what the catalog ordering supports.

2. `getLatestVersion` re-sorts the catalog on every render: it calls `getVersionOptions`, which runs
   `semver.rcompare` over the whole list, and the call site is a plain `const` rather than a memo — so
   a java row sorts 6165 entries on every render. That's the cost the `catalogOptions` memo removed in
   #1881, back again. Fixing 1 removes it too, since a scan of the raw list needs no sort.

3. The behaviour that changed has no test. Every `getLatestVersion` case either starts with a catalog
   whose newest entry is already the newest semver (`['24.2.0', '22.12.0', …, 'lts-iron']`) or one
   where nothing is semver at all (`['zulu-musl-8.96.0.19', 'zulu-musl-8.94.0.17', '18.0.1.1']` — all
   three fail `semver.valid`), so they pass identically under either ordering. The case that matters is
   a mixed catalog where the newest match is the non-semver one, e.g. `['28.5.0.6', '28.4.3']`.

4. Worth pinning down before this ships: with catalog order restored, `python: latest` predicts
   `3.16-dev` and `ruby: latest` predicts `4.1-dev`. Since the CLI hands the prefix to
   `mise latest <tool>@<prefix>`, the hint is only right if mise picks dev builds too. "Currently
   resolves to 3.16-dev" under the default row is alarming if it isn't what actually gets installed —
   and it's the one row most people will see.

5. The description predates the #1881 rework. It says resolution is "the first catalog entry that
   starts with the prefix", but matching is at a separator boundary now, not `startsWith`. The
   `Widen the version list fetch to any catalog known tool` bullet also has no hunk behind it —
   `useToolVersions(canonicalToolId, isExactKnownTool || hasPrefixDropdown)` is identical on master and
   on this branch.

*AI-assisted review: the digging through source is Claude's, the judgement and every ask here are mine.
Not convinced by a point? Ask — I'll show the working.*

marcell-vida:
> 1. The hint doesn't follow the rule the description states, and it's wrong on the live catalog.
>    `getLatestVersion` searches `getVersionOptions(...)`, whose ordering is a display choice — semver
>    sorted first, everything else appended — so every version that isn't clean 3-part semver is
>    demoted below every version that is. This has nothing to do with pre-releases: erlang's and java's
>    4-part builds are not valid semver, so
>
> | tool   | prefix   | first catalog match | what the hint shows |
> |--------|----------|---------------------|---------------------|
> | erlang | `28`     | `28.5.0.6`          | `28.4.3`            |
> | erlang | `27.3`   | `27.3.4.17`         | `27.3.4`            |
> | java   | `26`     | `26.0.2.1`          | `26.0.2`            |
> | python | (Any)    | `3.16-dev`          | `3.14.7`            |
> | ruby   | (Any)    | `4.1-dev`           | `4.0.6`             |
>
>    erlang `28` is a whole minor out. Iterating `toolVersions.versions` with `matchesPrefix` is both
>    what the description says and what the catalog ordering supports.

Fixed in `fix: resolve a prefix the way mise does`. The resolver no longer goes through `getVersionOptions`. The catalog is read in order, which is what Mise does, and semver only ranks a line where it can read every candidate. The above table now resolves, but there are two thing to note:

- `erlang 27.3` gives `27.3`, not `27.3.4.17`, because an exact catalog hit wins before anything else (validated with `mise latest` command).
- python/ruby rows are working based on the answer to question `4`

> 2. `getLatestVersion` re-sorts the catalog on every render: it calls `getVersionOptions`, which runs
>    `semver.rcompare` over the whole list, and the call site is a plain `const` rather than a memo — so
>    a java row sorts 6165 entries on every render. That's the cost the `catalogOptions` memo removed in
>    #1881, back again. Fixing 1 removes it too, since a scan of the raw list needs no sort.

Fixed, it was indeed fixed by `1` and memo was added in perf: `memoize the resolved version hint`.

> 3. The behaviour that changed has no test. Every `getLatestVersion` case either starts with a catalog
>    whose newest entry is already the newest semver (`['24.2.0', '22.12.0', …, 'lts-iron']`) or one
>    where nothing is semver at all (`['zulu-musl-8.96.0.19', 'zulu-musl-8.94.0.17', '18.0.1.1']` — all
>    three fail `semver.valid`), so they pass identically under either ordering. The case that matters is
>    a mixed catalog where the newest match is the non-semver one, e.g. `['28.5.0.6', '28.4.3']`.

Good catch, added other tests with the mixed case you describe for prefix 26, plus prerelease skipping, the exact-hit rule, and a flutter build-metadata line.

> 4. Worth pinning down before this ships: with catalog order restored, `python: latest` predicts
>    `3.16-dev` and `ruby: latest` predicts `4.1-dev`. Since the CLI hands the prefix to
>    `mise latest <tool>@<prefix>`, the hint is only right if mise picks dev builds too. "Currently
>    resolves to 3.16-dev" under the default row is alarming if it isn't what actually gets installed —
>    and it's the one row most people will see.

Thanks for spotting! I checked and mise does not pick dev builds. `latest` searches a  list already filtered regex (located in `src/plugins/mod.rs`), which drops `-dev`, `-rc`, `-src` and similar version entries. It looks like the best option is to copy that regex rather than writing custom filters, since it's hard to match the same results. However it could drift overtime, so we have to check it regularly. I would still keep it this way, as it's a helper, but does not guarantee anything ("**Currently** resolves to" already gives the tone of uncertainty), but we could possibly add a UI warning, that the resolution is approximate?

> 5. The description predates the #1881 rework. It says resolution is "the first catalog entry that
>    starts with the prefix", but matching is at a separator boundary now, not `startsWith`. The
>    `Widen the version list fetch to any catalog known tool` bullet also has no hunk behind it —
>    `useToolVersions(canonicalToolId, isExactKnownTool || hasPrefixDropdown)` is identical on master and
>    on this branch.

Correct, I'm updating the description to match.

moczolaszlo (changes requested):
Thanks for the thorough rework — all five points check out, and I compared `PRERELEASE_MARKER` and the
`find_match_in_list` description against mise's source: both match. To test the mirror end to end, I
simulated mise's list path (`fuzzy_match_versions` → `find_match_in_list`, catalog reversed as the doc
describes) against `getLatestVersion` over every offered prefix of 13 live catalogs. Ten tools agree on
every prefix, and "Any" agrees on all 13. The disagreements come from three places:

1. A line holding nothing but prereleases answers here but not in mise. `candidates = stable.length ?
   stable : line` falls back to the prereleases, but mise only falls back for `latest` —
   `if matches.is_empty() && query == "latest"` in `latest_version_for_query_with_selection_options`
   (`src/backend/mod.rs`). For any other prefix mise finds nothing and the install fails with "no
   matching version". So `python: 3.16:latest` shows "Currently resolves to 3.16-dev" for a config that
   fails at build time; the dropdown offers 14 such python prefixes (`3.16`, `3.15`, `pypy3.3`, …) and 6
   ruby ones (`4.1`, `3.5`, `truffleruby-1`, …). The hint is wrong in the direction that reassures.
   Showing no hint there — mise's answer — seems right, and the
   `falls back to the newest entry when a line holds nothing but prereleases` test would flip.

2. `semver.prerelease` isn't mise's rule. For semver entries `isPrerelease` drops anything with a
   prerelease part, but mise drops only what `VERSION_REGEX` matches. ruby's patch releases are stable
   to mise and prereleases to `semver`: `semver.prerelease('2.0.0-p648')` is `['p648']`, and nothing in
   the regex matches `-p648`. So ruby `2.0` hints `2.0.0-rc2` where mise gives `2.0.0-p648`, `1.9` hints
   `1.9.3-rc1` where mise gives `1.9.3-p551`, and `1` hints `1.8.7` where mise gives `1.9.3-p551`.
   Running `PRERELEASE_MARKER` for every entry is what mise does.

3. The `rsort` contradicts the doc's own rule. The doc says mise doesn't sort, takes the last entry in
   list order, and "last there is first here" — which makes mise's answer for `flutter@v1.12` the first
   catalog entry, `v1.12.13+hotfix.7`, the answer the Trap calls wrong. Either flutter's catalog isn't
   mise's list reversed, in which case the mirror is approximate there whichever way it sorts, or the
   sort makes the hint disagree with mise. You checked erlang `27.3` with `mise latest`;
   `mise latest flutter@v1.12.13` would settle this one too.

4. Not from this diff, but the hint is where it shows: `matchesPrefix` treats `_` and `+` as separators
   everywhere, while `fuzzy_match_versions` uses `[+\-.]` for numeric queries, `[\-.]` otherwise, and
   never `_`. The java dropdown offers 288 prefixes cut at `_` — `semeru-openj9-8u472-b08` and the like —
   that `mise latest` can't resolve, and the hint names a concrete version for each. A `v`-prefixed query
   also matches the unprefixed versions in mise (`query_without_v_regex`), so flutter `v1` is `1.22.6`
   there and `v1.12.13+hotfix.9` here. This lives in `matchesPrefix`/`toPrefixes` from #1881, so fixing
   it here or separately is your call.

On the approximate-resolution warning: I wouldn't add a blanket one. Where the hint is wrong today it
is confidently wrong — 1 and 4 name a version for configs that fail — and a disclaimer doesn't change
that; showing nothing where mise would find nothing does. "Currently" carries the right tone for the
rest. The one row that is inherently less certain is "Any": mise tries the backend's upstream
`latest_stable_version` first and only then the list, and the catalog can't see that. It agreed on all
13 tools today, though.

<details><summary>How the comparison was run</summary>

mise side: the default `fuzzy_match_versions` — exact hit first, `VERSION_REGEX` filter, numeric
queries as `^v?<query>([+\-.].+)?$`, other queries as `^<query>([\-.].+)?$`, plus the
`query_without_v_regex` variant — over the catalog reversed into mise's ascending order, then
`find_match_in_list`. Editor side: `getLatestVersion` and `getPrefixOptions` as on this branch. Points
1, 2 and 4 don't depend on the reversal assumption, since they're about whether anything matches or how
an entry is classified; point 3 does, which is why it's a question.
</details>

*AI-assisted review: the digging through source is Claude's, the judgement and every ask here are mine.
Not convinced by a point? Ask — I'll show the working.*

PR #1891, opened 2026-09-02
Handle YAML aliases and empty nodes in document reads

bitrise-io-kolega:
### Root cause

`YmlUtils.getMapIn`/`getSeqIn` threw `Expected a YAMLMap/YAMLSeq at path "..."` for two shapes of perfectly loadable `bitrise.yml`. Both reported message variants reproduce exactly:

1. **YAML anchors/aliases.** `Document.getIn`/`hasIn` stop at an `Alias` node because an alias is not a collection, while `toJSON()` — what the UI renders from — expands aliases. So the UI renders a step written as `- *common_step`, but every document-level read of it hands back the raw `Alias`, which the collection getters reject. Reproduces `Expected a YAMLMap at path "steps.2", but found Alias`.
2. **A key written with no value** (`meta:` / `bitrise.io:` on its own line, a bare `-` seq item) parses to a *null scalar*, not to a missing key — so `hasIn` said "present" while there was nothing to return, and even `createIfNotExists: true` threw instead of filling it in. Reproduces `Expected a YAMLMap at path "workflows.<wf>.meta.bitrise.io", but found Scalar` (from `StackAndMachineService.updateFieldValue`).

The throw happens on a render path (`StepCard` → `useContainerReferences` → `ContainerService` → `StepService.getStepOrThrowError` → `getMapIn`), so it took the whole editor down instead of degrading one card.

**Why one session produced ~27.8k occurrences:** the root `ErrorBoundary`'s `PassThroughFallback` called `resetError()` unconditionally in an effect. A *deterministic* render error re-throws on the very next render, gets caught, resets again — an unbounded reset/render loop, each iteration re-reporting to RUM. That is the amplifier, and it would amplify any future render-time throw.

### Fix

- `YmlUtils.getIn` resolves aliases at every hop of the path (last matching anchor declared before the alias — YAML's own rule). Anchor list cached per root object, same pattern/caveat as `collectPaths`. Zero cost when no alias is present.
- `getMapIn`/`getSeqIn` treat a null-valued key as absent: `undefined` on a plain read; with `createIfNotExists` they fill the collection in rather than throwing.
- `StepService.getSourceOrThrowError`/`getStepOrThrowError` read via `getIn` and raise their own "not found" instead of surfacing the low-level YAML shape error. The step lookup is rooted at the document so an aliased step resolves against anchors anywhere in the file.
- `ContainerService.getContainerReferenceFromInstance` uses a new read-only `findStepData` — neither throws nor materializes a missing option map (the old path mutated the store's document during render). `parseContainerReference` returns `undefined` for a reference naming no container; a non-sequence `service_containers` is skipped instead of thrown on. Reporting malformed YAML is the validator's job, not a render read's.
- `PassThroughFallback` gives up after 3 resets inside a 1s window. Terminal state is the same blank fallback the loop already produced, minus the storm.

### Tests

+24 tests across `YmlUtils.spec`, `StepService.spec`, `ContainerService.spec`, `StackAndMachineService.spec`: alias-at-step / alias-at-workflow / alias mid-path / dangling alias / empty node (read + create) / plain-string step / empty container reference / non-sequence `service_containers`. Full suite green: **1512 passed, 40 suites**; `tsc --noEmit` clean; eslint clean (only 7 pre-existing warnings); `knip` output byte-identical to master.

Two existing `ContainerService` tests that asserted `getContainerReferenceFromInstance` **throws** for a missing step index now assert it returns `undefined` — that function runs during render and must not throw. That is the one intentional behavior change to flag for review.

### Assumptions / decisions worth a reviewer's eye

- **Writing through an alias mutates the anchor target** (shared by every other alias user). That is what YAML means, and it round-trips losslessly (there is a test asserting the `*common_meta` alias survives a stack change). The alternative — deep-copying on write — is a bigger product decision I did not make unilaterally.
- **Out of scope, flagged not fixed:** an alias with *no matching anchor* makes the document unserializable — `doc.toString()` and `doc.toJSON()` both throw `ReferenceError: Unresolved alias`, so `YmlUtils.toYml`/`isEquals`/`toJSON` and the store subscriber crash. Different error signature, different Datadog issue, and the backend rejects such a config anyway; typing `*foo` before `&foo` in the YML editor would hit it. Worth a follow-up ticket.

### DIFF
--- a/source/javascripts/core/services/ContainerService.ts
+++ b/source/javascripts/core/services/ContainerService.ts
@@ -257,45 +257,64 @@ function getAllContainers(containers: Containers, selector: (container: Containe
     .filter(selector);
 }
 
-// `undefined` for a reference that names no container (an empty map, an empty seq item, a number).
-// Both callers run during render, where throwing would take down the whole card.
-function parseContainerReference(value: unknown): ContainerReference | undefined {
-  if (typeof value === 'string') {
-    return value ? { id: value, recreate: false } : undefined;
-  }
+/** The plain value behind a node, resolving an alias first. */
+function scalarValueOf(doc: Document, node: unknown) {
+  const resolved = YmlUtils.resolveAlias(doc, node);
 
-  if (typeof value !== 'object' || value === null) {
-    return undefined;
-  }
-
-  const [containerId, config] = Object.entries(value as Record<string, { recreate?: boolean }>)[0] ?? [];
-
-  if (!containerId) {
-    return undefined;
-  }
-
-  return { id: containerId, recreate: config?.recreate === true };
+  return isScalar(resolved) ? resolved.value : resolved;
 }
 
-function getContainerReferences(type: ContainerType, yamlMap: YAMLMap): ContainerReference[] | undefined {
-  if (type === ContainerType.Execution) {
-    const node = yamlMap.get(ContainerReferenceField.Execution);
+/**
+ * A container reference — a bare id (`ubuntu`) or `{ id: { recreate: true } }` — or `undefined` when
+ * the node names no container (an empty map, an empty seq item, a number).
+ *
+ * Read off the nodes rather than via `toJSON()`, and resolved against `doc` at every level: a
+ * reference can be written as `*anchor` (or hold one), and a subtree containing an unresolved alias
+ * serializes to `{ source: '<anchor>' }`, which would parse as a container literally named "source".
+ * Every caller runs during render, where throwing would take down the whole card.
+ */
+function parseContainerReference(doc: Document, node: unknown): ContainerReference | undefined {
+  const resolved = YmlUtils.resolveAlias(doc, node);
 
-    if (!node || (!isMap(node) && typeof node !== 'string')) {
+  if (isMap(resolved)) {
+    const pair = resolved.items[0];
+    const id = pair ? scalarValueOf(doc, pair.key) : undefined;
+
+    if (typeof id !== 'string' || !id) {
       return undefined;
     }
 
-    const value = isMap(node) ? node.toJSON() : node;
-    const reference = parseContainerReference(value);
+    const options = YmlUtils.resolveAlias(doc, pair?.value);
+    const recreate = isMap(options) ? scalarValueOf(doc, options.get('recreate', true)) : undefined;
+
+    return { id, recreate: recreate === true };
+  }
+
+  const value = isScalar(resolved) ? resolved.value : resolved;
+
+  return typeof value === 'string' && value ? { id: value, recreate: false } : undefined;
+}
+
+// `doc` is the alias-resolution root: an anchor referenced from a step is almost always declared
+// elsewhere in the file, so resolving against `yamlMap` alone would never find it.
+function getContainerReferences(
+  type: ContainerType,
+  yamlMap: YAMLMap,
+  doc: Document,
+): ContainerReference[] | undefined {
+  if (type === ContainerType.Execution) {
+    const reference = parseContainerReference(doc, yamlMap.get(ContainerReferenceField.Execution, true));
+
     return reference ? [reference] : undefined;
   }
 
   if (type === ContainerType.Service) {
-    // `services:` is only usable as a sequence; any other shape (a bare scalar, a map) is left to
-    // the YAML validator to report rather than thrown from here, mid-render.
-    const node = YmlUtils.getIn(yamlMap, [ContainerReferenceField.Service], true);
-    const references = (isSeq(node) ? (node.toJSON() as unknown[]) : [])
-      .map(parseContainerReference)
+    // Resolved first: `*anchor` pointing at a sequence is a config the CLI runs, not a bad shape.
+    // What's left after that — a bare scalar, a map — is genuinely unusable here, and reporting it
+    // is the YAML validator's job rather than something to throw over mid-render.
+    const node = YmlUtils.resolveAlias(doc, yamlMap.get(ContainerReferenceField.Service, true));
+    const references = (isSeq(node) ? node.items : [])
+      .map((item) => parseContainerReference(doc, item))
       .filter((reference): reference is ContainerReference => Boolean(reference));
 
     if (references.length > 0) {
@@ -313,7 +332,7 @@ function getContainerReferencesFromStepBundleDefinition(sourceId: string, type: 
     return undefined;
   }
 
-  return getContainerReferences(type, yamlMap);
+  return getContainerReferences(type, yamlMap, doc);
 }
 
 function getContainerReferenceFromInstance(
@@ -334,7 +353,7 @@ function getContainerReferenceFromInstance(
     return undefined;
   }
 
-  return getContainerReferences(type, yamlMap);
+  return getContainerReferences(type, yamlMap, doc);
 }
 
 // The container id a reference node points at — a bare scalar id, or the first key of a
--- a/source/javascripts/core/services/ContainerService.ts
+++ b/source/javascripts/core/services/ContainerService.ts
@@ -359,6 +359,11 @@ function getContainerReferenceFromInstance(
 // The container id a reference node points at — a bare scalar id, or the first key of a
 // `{ id: {opts} }` map — or undefined if it isn't a container reference. The inverse of the shapes
 // `referencesContainer` matched (a scalar equal to the id, or a map whose first key is the id).
+//
+// Read the node with `YmlUtils.getIn`, never `doc.getIn`: the paths handed to this come from
+// `getMatchingPaths`, which enumerates `toJSON()` and so happily matches a leaf written as
+// `*anchor`. A native read hands back the raw `Alias` — neither a map nor a scalar — and the
+// reference silently disappears from the usage counts.
 function referencedContainerId(node: unknown): string | undefined {
   if (isMap(node)) {
     return node.items.length > 0 ? String(node.items[0]?.key) : undefined;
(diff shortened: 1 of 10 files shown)


moczolaszlo (changes requested):
Verified the repro end to end: with this PR's specs copied onto master, 25 of them fail with exactly the reported messages (`Expected a YAMLMap at path "workflows.primary.meta.bitrise.io", but found Scalar`, `... "steps.1", but found Alias`), and all pass here. Full suite 1512/40 green, `tsc` clean.

**One gap, in the function this PR rewrote for that very reason.** `getContainerReferences` still reads both fields alias-blind, so the document/UI split fixed everywhere else survives here:

```yaml
_defaults:
  image: &img ruby-3
  svcs: &svc [postgres]
workflows:
  wf:
    steps:
      - script@1:
          container: *img
          services: *svc
```

`toJSON()` gives `{container: 'ruby-3', services: ['postgres']}`, but `getContainerReferenceFromInstance` returns `undefined` for both: `yamlMap.get(Execution)` hands back the `Alias`, which `!isMap(node) && typeof node !== 'string'` rejects, and `isSeq(node)` is false for the aliased `services`. The card renders without its container chips instead of crashing — better than before, still the wrong render. Both call sites already hold `doc`, so rooting these two reads there (or resolving the alias first) closes it, with a test alongside the ones you already added. The `services` comment reads as deliberate — "any other shape is left to the validator" — but an alias to a sequence isn't a malformed shape, it's a config the CLI runs.

Separately, not a change request: writing through the anchor is the correct YAML semantics, but it means editing one aliased step or meta block silently rewrites every other workflow sharing that anchor, with nothing in the UI saying so. That deserves an explicit product decision before it ships rather than after the first support ticket.

_AI-assisted review: the digging through source is Claude's, the judgement and every ask here are mine. Not convinced by a point? Ask — I'll show the working._

source/javascripts/core/utils/YmlUtils.ts
```diff
@@ -253,6 +345,12 @@ function setIn(root: Root, path: Path, value: unknown, stringToTypedValue = true
     throw new Error('Path cannot be empty when setting a value');
   }
 
+  const rebased = rebaseThroughAlias(root, path);
```
zoltanszabo-bitrise, line 407: `setIn` now reads `existingNode` through the alias-resolving `getIn` (line 384), and `toScalar` reuses that node by mutating it in place. When the **final** path segment holds an alias to a scalar, the write lands on the anchor and the anchored node is then re-inserted at the alias position.

Verified on `6112c58`:

```yaml
app:
  envs:
  - FOO: &v "x"
  - BAR: *v
```

`YmlUtils.setIn(doc, ['app', 'envs', 1, 'BAR'], 'changed')` →

```yaml
app:
  envs:
    - FOO: &v "changed"   # <- corrupted
    - BAR: &v "changed"   # <- alias gone, anchor name declared twice
```

On `master` the same call gives `- FOO: &v "x"` / `- BAR: changed`: the edit was local.

This isn't the shared-write semantics the description flags. A last-segment alias is the value being *replaced*, not a hop on the way to a deeper key — and the map-valued equivalent already behaves that way (`meta: *common_meta` set to a new map correctly replaces the reference and leaves the anchor alone). Only the scalar path corrupts the anchor, and it also emits a duplicate anchor name.

Suggest skipping the `existingNode` reuse when the raw node at `path` is an `Alias` — e.g. read it alias-blind for this purpose, or `isAlias(rawNodeAt(path)) ? undefined : getIn(root, path, true)`.

source/javascripts/core/utils/YmlUtils.ts
```diff
@@ -438,16 +545,17 @@ function getSeqIn(root: Root, path: Path, createIfNotExists = false) {
...
-  }
+  let node = getIn(root, path, true);
+
+  if (isEmptyNode(node)) {
+    if (!createIfNotExists) {
+      return undefined;
+    }
 
-  if (!hasIn(root, path) && createIfNotExists) {
     setIn(root, path, new YAMLSeq());
```
zoltanszabo-bitrise, line 639: `resolveAlias` can only see anchors reachable from `root`, but most `getMapIn`/`getSeqIn` callers pass a **nested collection**, not the document. For those, a perfectly valid `*anchor` resolves to `undefined`, `isEmptyNode` reports "absent", and this branch replaces the alias with a fresh empty collection — deleting the data it pointed at. Same at `getMapIn` (line 585).

This is the root cause behind the (now fixed) ContainerService review comment, *"rooting this lookup at the step option map cannot resolve a normal alias whose anchor is elsewhere in the document"*. The doc-root fix landed in `ContainerService` only; every nested-root `YmlUtils` caller still has it — and on the create path it destroys data rather than returning nothing.

Two confirmed callers, both verified on `6112c58`:

**1. `StepService.updateStepInput` → `getSeqIn(stepData, ['inputs'], true)` (StepService.ts:521)**

```yaml
workflows:
  base:
    steps:
    - script@1:
        inputs: &common_inputs
        - content: echo hi
        - other: 1
  wf:
    steps:
    - script@1:
        inputs: *common_inputs
```

Editing input 0 of `wf`'s step leaves `wf` with `inputs: [{content: echo bye}]` — `other: 1` is gone from the document.

**2. `TriggerService` → `getMapIn(entity, ['triggers'], true)` (TriggerService.ts:596)**

`wf3.triggers: *common_triggers` becomes `{enabled: false}` — the whole `push` list dropped from that workflow.

Same shape (not probed) at StepBundleService.ts:374, :466, :503.

On `master` both threw `Expected a YAMLSeq/YAMLMap at path "...", but found Alias` — an error toast, no data loss. This branch turns that into silent destruction of user config, which is a worse failure mode than the crash being fixed.

The read-only variant is affected too: `WorkflowService.ts:296` reads an aliased `before_run`/`after_run` as absent, so the chain renders empty instead of throwing.

Fix needs both halves: resolve against the owning document (thread the doc through, or keep a back-reference), **and** never overwrite a node whose raw value is an `Alias` that couldn't be resolved — that case is "can't handle", not "absent".

zoltanszabo-bitrise (commented):
Two defects introduced by the alias resolution itself, both verified against `6112c58` with a before/after on `master`. Neither is covered by the existing comments (those all predate `6112c58` and are addressed by it).

Both share one cause: `resolveAlias` is applied in two places that previously wanted the *raw* node.

source/javascripts/core/services/ContainerService.ts
```diff
@@ -290,14 +345,15 @@ function getContainerReferenceFromInstance(
...
+  // bundle defined in another module) is absent from the active document, and a step can be written
+  // in a shape this can't read. Either way the card renders without container references.
+  const yamlMap = findStepData(doc, source, sourceId, stepIndex);
+  if (!yamlMap) {
     return undefined;
   }
-  const yamlMap = getStepDataOrThrowError(doc, source, sourceId, stepIndex);
-  return getContainerReferences(type, yamlMap);
+
+  return getContainerReferences(type, yamlMap, doc);
```
zoltanszabo-bitrise, line 356: Third one, in this file: the *other* container-reference read still bypasses the alias resolution this PR added, so the two layers disagree inside a single function.

`getWorkflowsUsingContainers` (lines 396-417) pairs the alias-**aware** `YmlUtils.getMatchingPaths` with a **native** `doc.getIn(path)` (lines 400 and 413). `getMatchingPaths` happily matches a path whose leaf is an alias, and then `doc.getIn` hands `referencedContainerId` the raw `Alias` — which is neither a map nor a scalar, so it returns `undefined` and the reference is dropped.

Verified on `6112c58`:

```yaml
_defaults:
  image: &img ruby-3
containers:
  ruby-3:
    image: ruby:3
workflows:
  aliased:
    steps:
    - script@1:
        execution_container: *img
  plain:
    steps:
    - script@1:
        execution_container: ruby-3
```

`ContainerService.getWorkflowsUsingContainers(doc, ['ruby-3'])` → `[["ruby-3", ["plain"]]]`. The `aliased` workflow is missing.

User-visible through `useContainerWorkflowUsage` → `WorkflowService.getUsedByText`: the containers list under-reports "Used by N Workflows", and a container referenced *only* through aliases reads as unused — so it can be deleted without the usage warning, leaving dangling `execution_container: *img` references behind.

Fix is the same shape as the one applied to `getContainerReferences` in this hunk: read the node alias-aware (`YmlUtils.getIn(doc, path, true)` + `YmlUtils.resolveAlias`, or just have `referencedContainerId` resolve first) instead of `doc.getIn(path)`.

---

Zooming out, since this is the third instance of one pattern: alias resolution is transparent where a caller passes `doc`, and absent where it doesn't. Across the codebase that's 28 of 46 `getMapIn`/`getSeqIn` call sites doc-rooted (transparent), 18 nested-rooted (scoped to the subtree, external anchors invisible — 7 of them on the `createIfNotExists` path, which is the silent data loss in my other comment), plus 7 native `doc.hasIn`/`doc.getIn` calls that bypass `YmlUtils` altogether (`ContainerService.ts:181/400/413/508`, `WorkflowService.ts:173/185/245`).

The result is that the same YAML resolves or doesn't depending on which root a caller happened to pass, and nothing at the call site makes that visible. Both uniform positions are coherent — "alias-tolerant" (never crash, never resolve) or "alias-transparent" (resolution is a document-level invariant). This is in between. If transparency is the target, the invariant needs to be "always rooted at the document", which means threading the doc into the nested-root callers and retiring those native calls — a larger change than this PR, not a smaller one. Worth deciding explicitly before this ships, alongside the shared-write question already raised.

zoltanszabo-bitrise (commented):
One more verified gap, plus the pattern the three findings share.

zoltanszabo-bitrise:
Btw transparent alias handlig is more like a feature request than a bugfix, and it would require product decision. #1899 will fix the crash and render loop. I would hand the alias handling to one of the CI teams. @AndrasEszes @moczolaszlo 

I would close this PR for now until a decision is made about supporting aliases in the WFE.

PR #1894, opened 2026-09-03
fix(env vars): address sortable env var rows by id, not a stale render index

bitrise-io-kolega:
## Summary

Removing an environment variable — on a Workflow's **Env Vars** tab or the project-level list — could throw:

```
Error: Environment variable not found in Workflow '<workflow_id>', index <N> is out of bounds
```

The index a row's handlers used was captured when the row was rendered, and stopped being valid the moment the list changed. This PR makes every mutation resolve the row's *current* position from its stable row id — including across the 250 ms debounce on key/value edits — and tightens the bounds validation in `EnvVarService` so a stale or malformed index can no longer surface an error or corrupt the document.

## Root cause

`EnvVarService` addresses env vars by their position in the YAML sequence (`workflows.<id>.envs.<N>`), so `useSortableEnvVars` has to hand it an index. It did that by baking the render-time index into each row's handlers:

```tsx
{envs.map((env, index) => (
  <SortableEnvVarItem onRemove={onRemove(index)} ... />
))}
```

```ts
const onRemove = (index: number) => () => {
  setEnvs(envs.filter((_, i) => i !== index));   // non-functional: reads the render-time list
  EnvVarService.remove({ source, sourceId, index });
};
```

Both the local list and the YAML document are mutated straight from the handler, but `envs` only catches up on the next render. So two handler calls that land in the same React batch both resolve against the pre-update list:

1. Rows are `[A, B, C]`. The user clicks Remove on `A`, then on `C` before React re-renders (a double click on one row does the same thing, and so does a click while a drag settles — the reported traces come from the sortable list).
2. The first call removes document index `0`. The document is now `[B, C]`.
3. The second call still carries the index `C` was *rendered* at — `2` — which is now past the end.
4. `getEnvVarOrThrowError` throws, out of a click handler where nothing catches it.

The same staleness applies over a longer window to key and value edits, which are debounced by 250 ms: a row removed or reordered inside that window meant the flushed write carried the row's old position.

`onDragEnd` had a related weakness: it passed raw `findIndex` results into `arrayMove`, and `arrayMove` reads a `-1` as an offset from the end, so a row that had gone missing would silently reorder unrelated rows.

## Changes

**`useSortableEnvVars` — identity instead of position**

- Handlers are keyed by the row's `uniqueId` and resolve its index at call time. A row that is already gone resolves to `-1` and the handler is a no-op.
- The list is mirrored in a ref that is updated synchronously alongside `setEnvs`, so a handler sees the effect of an earlier call in the same batch instead of the last rendered list. This also fixes a latent lost-update bug: two edits to different rows in one batch previously kept only one.
- **The debounced key and value writes carry the `uniqueId` across the debounce boundary** and resolve the position when they actually run, dropping the write if the row is gone by then. `oldKey` / `key` are still captured at call time on purpose: they describe the row itself — what the caller knows the document holds for it — not where it sits.
- **A pending write also re-checks the file it was scheduled against, and whether the list is still mounted**, and declines to land otherwise. See the next section.
- `onDragEnd` bails out when either dragged row is no longer in the list.

**`EnvVarService` — validate at the boundary**

- `remove` validates the index and returns the document untouched when it is out of bounds, instead of throwing. An index past the end already describes the requested end state. The genuine errors (missing `sourceId`, unknown Workflow, no `envs` section) still throw as before.
- `reorder` now requires the index set to be an actual permutation: matching length, unique, and every entry an in-range integer. Length alone didn't make it one — `[0, 0]` on a two-entry list passed the previous check, copied the first entry over the second and dropped it. It throws *before* the new sequence is built, so a malformed set can't touch the document.

## Where a pending write is allowed to land

Two findings from review, both pre-existing on `master` and both about the debounce window rather than the index inside it.

**The write can end up in the wrong file.** `EnvVarService.updateKey`/`updateValue` → `updateBitriseYmlDocument` → `editableActiveSlice()`, which reads `selectedNodeId` *when the mutator runs*. In modular mode the same `sourceId` can live in several files, so a file-tab switch inside the 250 ms window redirects the write. I reproduced it: with `bitrise.yml` and an included `child-a.yml` both defining `workflows.wf1.envs` with a matching key at the same index, a rename scheduled against the root landed in `child-a.yml` and left `bitrise.yml` untouched — no error, no sign anything went wrong.

Worth being precise about what fixes it, because I got this wrong once on the way: reading the active file from a rendered value, or from a ref an effect keeps up to date, is **not** enough. A file switch reaches the store before React re-renders, and the timer can fire in that gap — that's exactly the shape of the reproduction above. The pending write now records the active file by reading the store, and re-reads the store when it flushes, so the check doesn't depend on React having caught up. There's a test for each side of that: one where the switch has been through a render, and one where it hasn't.

**The write can outlive the list.** The debounce timer survives unmount, because `useDebounceCallback`'s unmount cancel targets a debounce instance built in its own `useEffect` — not the one it hands back and that actually gets called. So closing the drawer within 250 ms of a keystroke still wrote, and if the Workflow was gone by then `getWorkflowOrThrowError` threw from inside a `setTimeout`, where no error boundary is listening. A pending write now checks it is still mounted before doing anything, which holds regardless of how many debounce instances have live timers.

## The debounce does not actually coalesce — and hoisting the options object is not a safe fix alone

The review is right about the mechanism: `{ leading: false }` is a fresh object every render, `useDebounceCallback` memoises on `[func, delay, options]`, and the `Input` is controlled by `env.key`, so every keystroke builds a new debounce instance. Nothing is coalesced; each keystroke schedules its own 250 ms write.

I did not hoist the options object to a module constant, because on its own it makes things worse, and I checked rather than assumed. With one stable instance the second keystroke of a burst cancels the first, and the surviving write carries an `oldKey` captured from local state — the *first* keystroke's value, which was never written to the document. `updateKey`'s `has(oldKey)` check then fails:

```
Environment variable key is not matching "N"
```

That is the throw the review itself names as evidence that coalescing isn't happening today. There's now a test pinning this (`lands every keystroke of a burst, because each rename builds on the previous one`) so the coupling is executable rather than folklore — it fails the moment someone hoists the constant.

Making coalescing correct needs two more things, and together they're a redesign of this write path rather than a line move:

1. **Burst-scoped `oldKey`** — the surviving write has to carry the key the document held *before* the burst, not the previous keystroke's local value.
2. **Per-row debouncing** — one shared debounced function coalesces *across* rows too, so typing in row A and then row B inside the window cancels A's write and silently loses that edit. Today's broken memoisation is what hides this.

I'd rather do that as its own PR, with its own write-volume before/after, than fold it in here — the writes-per-keystroke behaviour is unchanged by this PR either way, and the harms the review identified (wrong file, post-unmount write, uncaught `setTimeout` throw) are fixed directly above and don't depend on the debounce topology. Happy to be overruled on the sequencing.

## What `remove`'s bounds check does and does not promise

An earlier revision of this description called `remove` "idempotent". That was wrong, and the review was right to push on it. Addressing is positional, so repeating a *still-valid* index removes whatever moved into that position — with `[A, B]`, `remove(0)` twice removes both entries.

The narrowed, accurate contract: **an out-of-bounds index is a no-op** — which is exactly the crash this PR fixes — and nothing more. Removing one specific row is the caller's responsibility, which is what the `uniqueId` addressing in `useSortableEnvVars` provides. Both the code comment and the tests now say this, and there's a test pinning the positional behaviour so nobody infers idempotence from the no-op case.

Making the service resolve rows by identity itself was the alternative, and I don't think it's available: the YAML document has no stable per-entry id to resolve against, and keys can repeat and are empty on a freshly added row (`{"": ""}`), so they can't stand in for one.

## Correction on the failure mode

An earlier revision of this description said the throw took the editor down "along with every unsaved change on the page". That was also wrong. The app's top-level error boundary renders a pass-through fallback that immediately calls `resetError()`, and the YAML document lives in a Zustand store outside React — so the tree recovers and unsaved work survives. The user-visible effect is a failed operation and a surfaced error, not a lost session. Worth fixing either way, but the severity claim was overstated and the code comment that repeated it has been reworded.

## Testing

- `EnvVarService.spec.ts`: the two "throws when the index is out of bounds" cases for `remove` are replaced with cases asserting the document is left untouched (out-of-bounds and negative index, project- and Workflow-level); a repeated-index case renamed to say what it actually covers (a repeat that has gone past the end); a new case pinning the positional semantics of a repeated still-valid index; and `reorder` cases for out-of-bounds, **duplicate**, negative and non-integer index sets, each asserting it throws and leaves the document unchanged.
- `useSortableEnvVars.spec.tsx` covers the production sequence directly: two removals in one `act()` remove exactly the two intended rows; a repeated removal of the same row is ignored; an `is_expand` toggle still lands on the row it was toggled on after a preceding row is removed; and eight fake-timer cases for the debounce window — a value and a key write landing on the right row after a preceding row is removed, a pending write dropped for a row removed before it flushes, a pending write dropped after unmount, the burst invariant above, and three modular-config cases (a file switch that has been through a render, a file switch that has only reached the store, and a positive control confirming a write still lands when the file *hasn't* changed).
- Each new case was checked against the pre-fix code: the three debounce cases and the duplicate `reorder` case fail without these changes and pass with them. The other three `reorder` cases were already caught by the previous check, so the new validation is a strict superset.
- `npm test` (1504 tests) and `npx tsc --noEmit` pass. `npm run lint` reports 0 errors, and one *fewer* warning than `master`: the seeding effect now sets state through `updateEnvs`, which `react-hooks/set-state-in-effect` no longer flags. That's incidental indirection, not a deliberate silencing — the effect still seeds the list on mount, as it must.

## Remaining known limitations

- The local list still isn't re-seeded when the document changes from elsewhere (e.g. an edit on the YAML tab while an env var list is mounted). That desync can misaddress a row that is still in bounds, which no amount of index validation can catch — it needs the list to re-seed on external document changes, and doing that without clobbering in-flight debounced edits is its own piece of work.
- The debounce still doesn't coalesce, for the reasons set out above. Fixing that properly (per-row debouncing plus burst-scoped `oldKey`) is the follow-up I'd like to do separately.

Neither affects the reported crash, which is fully covered.

---
_Found via a routine Datadog FE error-log check (service:wfe, last 24h). Fixed by an RDE coding agent, orchestrated by Kolega._

### DIFF
--- a/source/javascripts/components/SortableEnvVars/useSortableEnvVars.ts
+++ b/source/javascripts/components/SortableEnvVars/useSortableEnvVars.ts
@@ -1,14 +1,25 @@
 import { DragEndEvent, DragStartEvent } from '@dnd-kit/core';
 import { arrayMove } from '@dnd-kit/sortable';
 import { useCallback, useEffect, useRef, useState } from 'react';
-import { useDebounceCallback } from 'usehooks-ts';
+import { useDebounceCallback, useUnmount } from 'usehooks-ts';
 
 import { SortableEnvVar } from '@/components/SortableEnvVars/SortableEnvVarItem';
 import { EnvVar, EnvVarSource } from '@/core/models/EnvVar';
 import EnvVarService from '@/core/services/EnvVarService';
+import { bitriseYmlStore } from '@/core/stores/BitriseYmlStore';
 import useBitriseYmlStore from '@/hooks/useBitriseYmlStore';
 
 import { listenToEnvVarCreated } from './SortableEnvVars.events';
+
+/**
+ * The file a write lands in. `updateBitriseYmlDocument` reads exactly this when the mutator runs, so
+ * it is read from the store — not from a rendered value or a ref an effect maintains — at both ends
+ * of a debounced write: a file switch reaches the store before React re-renders, and the debounce
+ * timer can fire in between.
+ */
+function activeWriteTarget() {
+  return bitriseYmlStore.getState().selectedNodeId;
+}
 
 type UseSortableEnvVarsProps = {
   source: EnvVarSource;
@@ -48,14 +59,44 @@ export const useSortableEnvVars = ({
     return envsRef.current.findIndex((env) => env.uniqueId === uniqueId);
   }, []);
 
-  // The debounced writes carry the row's `uniqueId` rather than its index, and resolve the position
-  // only when they actually run: the row can move or go away during the delay (a preceding row
-  // removed, a reorder), and a position captured before the delay would write to whatever ended up
-  // there. `oldKey`/`key` describe the row itself, so those are still captured at call time — they
-  // are what the caller knows the document holds for that row, whatever position it is at.
+  const isMountedRef = useRef(true);
+
+  useUnmount(() => {
+    isMountedRef.current = false;
+  });
+
+  /**
+   * Where a pending write is still allowed to land: the row's current position, or -1 when it may
+   * not be applied at all.
+   *
+   * A debounce timer outlives the interaction that scheduled it, and the write resolves both its
+   * row and its target document only when it runs:
+   *
+   * - The row moves when a preceding row is removed or the list is reordered, so a position
+   *   captured before the delay would write to whatever ended up there.
+   * - `updateBitriseYmlDocument` picks the file from the active tab as the mutator runs. In modular
+   *   mode the same `sourceId` can live in several files, so a file switch inside the window would
+   *   redirect the write into a document it was never meant for — silently, when that file happens
+   *   to hold a matching key. See `activeWriteTarget` for why the store is read directly here.
+   * - After unmount there is nothing left to write for. (`useDebounceCallback`'s own unmount cancel
+   *   targets a different debounce instance than the one it hands back, so the timer still fires.)
+   */
+  const resolvePendingRow = useCallback(
+    (at: { uniqueId: string; fileId?: string }) => {
+      if (!isMountedRef.current || at.fileId !== activeWriteTarget()) {
+        return -1;
+      }
+
+      return indexOf(at.uniqueId);
+    },
+    [indexOf],
+  );
+
+  // `oldKey`/`key` stay captured at call time: they describe the row itself — what the caller knows
+  // the document holds for it — not where it sits, so they survive the row moving.
   const flushKeyUpdate = useCallback(
-    (newKey: string, at: { uniqueId: string; oldKey: string }) => {
-      const index = indexOf(at.uniqueId);
+    (newKey: string, at: { uniqueId: string; fileId?: string; oldKey: string }) => {
+      const index = resolvePendingRow(at);
 
       if (index === -1) {
         return;
@@ -63,12 +104,12 @@ export const useSortableEnvVars = ({
 
       EnvVarService.updateKey(newKey, { source, sourceId, index, oldKey: at.oldKey });
     },
-    [indexOf, source, sourceId],
+    [resolvePendingRow, source, sourceId],
   );
 
   const flushValueUpdate = useCallback(
-    (value: string, at: { uniqueId: string; key: string }) => {
-      const index = indexOf(at.uniqueId);
+    (value: string, at: { uniqueId: string; fileId?: string; key: string }) => {
+      const index = resolvePendingRow(at);
 
       if (index === -1) {
         return;
@@ -76,7 +117,7 @@ export const useSortableEnvVars = ({
 
       EnvVarService.updateValue(value, { source, sourceId, index, key: at.key });
     },
-    [indexOf, source, sourceId],
+    [resolvePendingRow, source, sourceId],
   );
 
   const updateKeyDebounced = useDebounceCallback(flushKeyUpdate, 250, { leading: false });
@@ -161,7 +202,7 @@ export const useSortableEnvVars = ({
 
     const oldKey = envsRef.current[index].key;
     updateEnvs((current) => current.map((env, i) => (i === index ? { ...env, key } : env)));
-    updateKeyDebounced(key, { uniqueId, oldKey });
+    updateKeyDebounced(key, { uniqueId, fileId: activeWriteTarget(), oldKey });
   };
 
   const onValueChange = (uniqueId: string) => (value: string) => {
@@ -173,7 +214,7 @@ export const useSortableEnvVars = ({
 
     const { key } = envsRef.current[index];
     updateEnvs((current) => current.map((env, i) => (i === index ? { ...env, value } : env)));
-    updateValueDebounced(value, { uniqueId, key });
+    updateValueDebounced(value, { uniqueId, fileId: activeWriteTarget(), key });
   };
 
   const onIsExpandChange = (uniqueId: string) => (isExpand: boolean) => {
--- a/source/javascripts/components/SortableEnvVars/useSortableEnvVars.ts
+++ b/source/javascripts/components/SortableEnvVars/useSortableEnvVars.ts
@@ -48,8 +48,39 @@ export const useSortableEnvVars = ({
     return envsRef.current.findIndex((env) => env.uniqueId === uniqueId);
   }, []);
 
-  const updateKeyDebounced = useDebounceCallback(EnvVarService.updateKey, 250, { leading: false });
-  const updateValueDebounced = useDebounceCallback(EnvVarService.updateValue, 250, { leading: false });
+  // The debounced writes carry the row's `uniqueId` rather than its index, and resolve the position
+  // only when they actually run: the row can move or go away during the delay (a preceding row
+  // removed, a reorder), and a position captured before the delay would write to whatever ended up
+  // there. `oldKey`/`key` describe the row itself, so those are still captured at call time — they
+  // are what the caller knows the document holds for that row, whatever position it is at.
+  const flushKeyUpdate = useCallback(
+    (newKey: string, at: { uniqueId: string; oldKey: string }) => {
+      const index = indexOf(at.uniqueId);
+
+      if (index === -1) {
+        return;
+      }
+
+      EnvVarService.updateKey(newKey, { source, sourceId, index, oldKey: at.oldKey });
+    },
+    [indexOf, source, sourceId],
+  );
+
+  const flushValueUpdate = useCallback(
+    (value: string, at: { uniqueId: string; key: string }) => {
+      const index = indexOf(at.uniqueId);
+
+      if (index === -1) {
+        return;
+      }
+
+      EnvVarService.updateValue(value, { source, sourceId, index, key: at.key });
+    },
+    [indexOf, source, sourceId],
+  );
+
+  const updateKeyDebounced = useDebounceCallback(flushKeyUpdate, 250, { leading: false });
+  const updateValueDebounced = useDebounceCallback(flushValueUpdate, 250, { leading: false });
 
   // `initialEnvs` is fixed to a specific file, so the active-file change must not re-seed it;
   // only re-seed on active-file changes when the list is read from the store.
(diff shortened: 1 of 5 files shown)


source/javascripts/core/services/EnvVarService.ts
```diff
@@ -220,7 +228,15 @@ function reorder(newIndices: number[], at: { source: EnvVarSource; sourceId?: st
...
     }
 
-    envs.items = newIndices.map((i) => envs.get(i));
+    const reordered = newIndices.map((i) => envs.get(i));
+
+    // `envs.get()` yields undefined for an index past the end, and assigning that into the sequence
+    // would write empty entries over the user's env vars. Bail out before touching the document.
+    if (reordered.some((env) => env === undefined)) {
+      throw new Error(`The indices (${newIndices.join(', ')}) don't address every environment variable exactly once`);
+    }
```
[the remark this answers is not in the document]
bitrise-io-kolega: Good catch — adding a uniqueness check alongside the existing bounds check so a duplicate-index permutation throws before the document is touched. Follow-up commit incoming.

source/javascripts/components/SortableEnvVars/useSortableEnvVars.ts
```diff
@@ -31,32 +31,80 @@ export const useSortableEnvVars = ({
...
+    envsRef.current = updater(envsRef.current);
+    setEnvs(envsRef.current);
+  }, []);
+
+  /** Current position of a row, or -1 once it is gone from the list. */
+  const indexOf = useCallback((uniqueId: string) => {
+    return envsRef.current.findIndex((env) => env.uniqueId === uniqueId);
+  }, []);
+
+  // The debounced writes carry the row's `uniqueId` rather than its index, and resolve the position
```
zoltanszabo-bitrise: The `{ leading: false }` literal is a new object on every render, and `useDebounceCallback` memoises on `[func, delay, options]` — so this returns a brand-new debounce instance each render. `Input` is controlled by `env.key`, so every keystroke re-renders, which means nothing is ever coalesced: each keystroke schedules its own 250 ms write.

Verified with a hook test on this branch: two successive `onKeyChange` calls both flush. (With a single coalesced flush the second `oldKey` wouldn't match the document and `updateKey` would throw — it doesn't, and the YAML ends at the second value.)

Two consequences:

- Several stale-index writes can be in flight at once, so the window flagged on 133/145 is wider than "the last keystroke", not narrower.
- `useDebounceCallback`'s `useUnmount` cancels `debouncedFunc.current` — a *different* instance built in its own `useEffect` and never invoked — so the instance actually called keeps its timer. Closing the drawer within 250 ms of a keystroke still writes; if the Workflow is deleted in between, `getWorkflowOrThrowError` throws from a `setTimeout` (uncaught).

Hoisting the options object to a module-level constant restores real debouncing and makes the unmount cancel effective.

source/javascripts/components/SortableEnvVars/useSortableEnvVars.ts
```diff
@@ -89,27 +137,53 @@ export const useSortableEnvVars = ({
...
   const onAdd = () => {
-    setEnvs([...envs, { uniqueId: crypto.randomUUID(), ...EnvVarService.EMPTY_ENV_VAR }]);
+    updateEnvs((current) => [...current, { uniqueId: crypto.randomUUID(), ...EnvVarService.EMPTY_ENV_VAR }]);
     EnvVarService.create({ source, sourceId });
   };
 
-  const onRemove = (index: number) => () => {
-    setEnvs(envs.filter((_, i) => i !== index));
+  const onRemove = (uniqueId: string) => () => {
+    const index = indexOf(uniqueId);
```
zoltanszabo-bitrise, line 186: Distinct from the stale-index problem already raised on this line and 133: the debounced flush also resolves the **target document** at flush time, not just the index. `EnvVarService.updateKey`/`updateValue` → `updateBitriseYmlDocument` → `editableActiveSlice()`, which reads `selectedNodeId` when the timer fires. In modular mode the same `sourceId` can live in several files — the comment on lines 30-31 says exactly this — so a file-tab switch inside the 250 ms window redirects the write.

Verified with a modular-mode hook test: root `bitrise.yml` and included `child-a.yml` both define `workflows.wf1.envs`; render the list on root, call `onKeyChange(ROOT_B)('RENAMED')`, then `selectNode('child-a')`, then advance timers →

- if `child-a`'s entry at that index carries the same key, the rename silently lands in `child-a.yml` and `bitrise.yml` is left untouched — cross-file corruption, no error;
- if the keys differ, `updateKey` throws `Environment variable key is not matching "ROOT_B"` from a `setTimeout` (so no ErrorBoundary toast) and the edit is dropped.

Carrying the `uniqueId` through the debounce boundary fixes the index but not this — the pending flush also has to be cancelled (or rebound) when the active file changes. Worth adding to the assumptions list either way: the current wording ("it would rename the wrong row rather than error") understates both modes.

zoltanszabo-bitrise (commented):
Two additional findings on the debounced path, both verified with hook tests on this branch and both pre-existing on `master` — neither is covered by the existing review comments.

source/javascripts/components/SortableEnvVars/useSortableEnvVars.ts
bitrise-io-kolega: Excellent catch, verified with a real hook test — hoisting the options object to a module-level constant so memoization and unmount-cancel actually work as intended. Follow-up commit incoming.

source/javascripts/components/SortableEnvVars/useSortableEnvVars.ts
bitrise-io-kolega: Good find — this is a real correctness gap beyond the index issue. Looking at cancelling/rebinding the pending debounced write when the active file changes, and will update the assumptions section to accurately describe both failure modes either way.

source/javascripts/components/SortableEnvVars/useSortableEnvVars.ts
bitrise-io-kolega: Diagnosis confirmed — but hoisting the options object alone regresses things, and I checked rather than assumed. With one stable instance, the second keystroke of a burst cancels the first, and the surviving write carries an `oldKey` from local state (the first keystroke's value) that was never written to the document — `updateKey`'s `has(oldKey)` check then throws `Environment variable key is not matching "N"`, which is the exact throw your note cites as evidence coalescing isn't happening today.

Real coalescing needs burst-scoped `oldKey` *and* per-row debouncing (today's broken memoisation also hides that one shared debounced function coalesces across rows, silently losing edits) — that's a redesign of this write path, not a line move. I'd rather do it as its own PR with its own write-volume before/after. I fixed the two harms that don't depend on debounce topology instead (see the wrong-file reply below, and a post-unmount write that could throw from an uncaught `setTimeout`), and added a test pinning the burst-cancellation coupling so it fails the moment anyone hoists the constant without fixing the rest. Happy to fold the redesign into this PR instead if you'd rather not split it — let me know.

source/javascripts/components/SortableEnvVars/useSortableEnvVars.ts
bitrise-io-kolega: Reproduced exactly as described — fixed now. One correction to flag on the mechanism: reading the active file from a rendered value, or from a ref an effect keeps in sync, is NOT enough. A file switch reaches the store before React re-renders, and the timer can fire in that gap, which is exactly the shape of your repro. The pending write now reads the store (not a rendered/ref value) both when it's scheduled and again when it flushes, so the check doesn't depend on React having caught up. Added tests for both a switch that's been through a render and one that's only reached the store, plus a positive control confirming a write still lands when the file hasn't changed.

Side note in case it's useful: on this branch, an `act()`-wrapped version of the repro passes even without the fix, because a file switch re-seeds the list with fresh `uniqueId`s, so the pending write's row becomes unfindable — which may be why "carrying `uniqueId` fixes the index but not this" read differently than what I measured. If your repro calls `selectNode` outside `act()`, that's likely why it reproduces there. Either way the underlying finding was correct and is now explicitly guarded rather than relying incidentally on `uniqueId` regeneration.

PR #1899, opened 2026-09-03
Stop unreadable YAML shapes from crashing the workflow editor

bitrise-io-kolega:
### Root cause

`getMapIn`/`getSeqIn` throw when a path holds something other than the collection the caller asked for. Two shapes of loadable `bitrise.yml` hit that, and one of them did so from inside a render, which took down the whole editor rather than degrading one card.

**A step the editor can't read.** `StepCard` renders from `toJSON()`, so it draws a card for any step in the list, then `useContainerReferences` reads the same step out of the document to look up its container references. When that node isn't a `{cvs: options}` map — a step written as a plain string, an empty `-` item, a YAML alias — the read threw during render.

**A key written with no value.** `meta:` or `bitrise.io:` alone on a line parses to a null scalar, not to a missing key, so the getters saw "present" with nothing to return and threw — including under `createIfNotExists`, while trying to create the very node they were asked for. That's what `updateStackAndMachine` hit on a workflow with an empty `meta.bitrise.io`.

**The reset loop.** The root `ErrorBoundary`'s fallback called `resetError()` unconditionally, so a deterministic render error re-threw, was caught, reset, and re-threw again — an unbounded loop re-reporting the same error every iteration. This is the reason a single affected session generated tens of thousands of duplicate error reports.

### Fix

- The container reads are now shape-tolerant, since they run on every `StepCard` render and must never throw: a new read-only `findStepData` returns `undefined` instead of throwing or materializing a missing option map; `parseContainerReference` returns `undefined` for a reference naming no container; a non-sequence `service_containers` is skipped; a step bundle definition of an unexpected shape returns no references. A card that can't be read renders without container badges. `StepService`'s lookups report their own "not found" instead of surfacing the low-level YAML shape error.
- Both getters now treat a null-valued key as absent: `undefined` on a read, and filled in on a create. `hasIn` was only used for that check and is dropped.
- `PassThroughFallback` gives up after three resets inside a one-second window. The terminal state is the same blank fallback the loop already produced, without the storm. This is defence-in-depth: it bounds any render-time throw, not just this one.

### Scope note

This deliberately does **not** resolve YAML aliases. An aliased step renders without its container badges rather than crashing, and `getMapIn`/`getSeqIn` still reject an alias exactly as before, which keeps the create path from ever overwriting one. Making the editor read and write through aliases is a larger change and is being handled separately (follow-up PR, #1891).

### Tests

Full suite green, `tsc --noEmit` clean, eslint clean, `knip` output unchanged vs master.

### DIFF
--- a/source/javascripts/core/services/ContainerService.spec.tsx
+++ b/source/javascripts/core/services/ContainerService.spec.tsx
@@ -1807,6 +1807,57 @@ describe('ContainerService', () => {
         ).toBeUndefined();
       });
 
+      it('skips an aliased reference entry instead of naming a container after the anchor key', () => {
+        updateBitriseYmlDocumentByString(yaml`
+          containers:
+            pg: &pg
+              image: postgres
+          workflows:
+            wf1:
+              steps:
+              - script:
+                  service_containers:
+                  - *pg
+                  - redis
+        `);
+
+        // Serializing the sequence as a whole would render the alias as `{ source: 'pg' }`, which
+        // would parse as a container called "source".
+        expect(
+          ContainerService.getContainerReferenceFromInstance(
+            'workflows',
+            'wf1',
+            0,
+            ContainerType.Service,
+            bitriseYmlStore.getState().ymlDocument,
+          ),
+        ).toEqual([{ id: 'redis', recreate: false }]);
+      });
+
+      it('returns undefined when the only reference entry is an alias', () => {
+        updateBitriseYmlDocumentByString(yaml`
+          containers:
+            pg: &pg
+              image: postgres
+          workflows:
+            wf1:
+              steps:
+              - script:
+                  service_containers:
+                  - *pg
+        `);
+
+        expect(
+          ContainerService.getContainerReferenceFromInstance(
+            'workflows',
+            'wf1',
+            0,
+            ContainerType.Service,
+            bitriseYmlStore.getState().ymlDocument,
+          ),
+        ).toBeUndefined();
+      });
+
       it('returns undefined for a step bundle definition of an unexpected shape', () => {
         updateBitriseYmlDocumentByString(yaml`
           step_bundles:
--- a/source/javascripts/core/services/ContainerService.spec.tsx
+++ b/source/javascripts/core/services/ContainerService.spec.tsx
@@ -1325,7 +1325,7 @@ describe('ContainerService', () => {
         ).toBeUndefined();
       });
 
-      it('should throw error when step index does not exist', () => {
+      it('should return undefined when step index does not exist', () => {
         updateBitriseYmlDocumentByString(yaml`
         workflows:
           wf1:
@@ -1333,7 +1333,7 @@ describe('ContainerService', () => {
               - script: {}
       `);
 
-        expect(() =>
+        expect(
           ContainerService.getContainerReferenceFromInstance(
             'workflows',
             'wf1',
@@ -1341,7 +1341,7 @@ describe('ContainerService', () => {
             ContainerType.Execution,
             bitriseYmlStore.getState().ymlDocument,
           ),
-        ).toThrow('Step at index 5 not found in workflows.wf1');
+        ).toBeUndefined();
       });
     });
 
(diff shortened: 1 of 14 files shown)


zoltanszabo-bitrise:
## Review: 4 findings

The core fix holds up. `findStepData`, the shape-tolerant `parseContainerReference`, the null-scalar handling in both getters and the reset budget's rolling-window arithmetic all do what the description claims, and dropping `getStepDataOrThrowError` from the read path also removes a store mutation that was happening during render.

The problem is that `getMapIn`'s contract changed and three callers outside the diff still depend on the old one. It now answers *"is there a readable map here?"*, but those three use it to ask *"does this key exist?"* — and for a key written with no value those are now different questions.

### 1. `StepBundleService.ts:198` — `changeStepBundleId` can write an unparseable yml

`throwIfStepBundleAlreadyExists` uses `getMapIn`, so it no longer sees a bundle written as a bare `b:`.

```yaml
step_bundles:
  b:
  a:
    steps: []
```

`changeStepBundleId('a', 'b')` now passes the guard — before this PR `getMapIn` threw ``Expected a YAMLMap at path "step_bundles.b", but found Scalar``, which blocked it. `updateKeyByPath` only rewrites the key in place, so the result is:

```yaml
step_bundles:
  b:
  b:
    steps: []
```

That reparses with 1 duplicate-key error (verified against the repo's `yaml` build), i.e. the editor writes a bitrise.yml that no longer parses, and `a`'s body is lost on reload. The same helper guards `createStepBundle` (`:230`) and the clone path (`:297`), where it silently overwrites the stub instead of reporting "already exists".

### 2. `PipelineService.ts:259` — same duplicate-key corruption in `renamePipeline`

```yaml
pipelines:
  b:
  a:
    workflows: {}
```

`renamePipeline('a', 'b')` passes `if (YmlUtils.getMapIn(doc, ['pipelines', newName]))`, then `updateKeyByPath` produces two `b:` keys. Same 1 parse error, `a`'s definition silently dropped.

### 3. `PipelineService.ts:230` — `createPipeline` overwrites a valueless key

With `pipelines:\n  release:` in the yml, creating a pipeline named `release` now succeeds and `setIn(doc, ['pipelines', 'release'], EMPTY_PIPELINE)` replaces the existing entry, instead of surfacing `Pipeline release already exists`.

**Fix for all three:** `doc.hasIn([...])`, which is exactly what `WorkflowService.createWorkflow` / `renameWorkflow` already use — that's why those two didn't regress. One small change, not three. Worth grepping every `getMapIn` in boolean position before merge, since the question it answers has changed.

### 4. `ContainerService.ts:294` — an aliased `service_containers` entry becomes a reference named `source`

`node.toJSON()` is called on the sequence with no document context, so an alias entry serializes to `{ source: '<anchor>' }` and `parseContainerReference` reads `source` as the container id.

```yaml
containers:
  pg: &pg
    image: postgres
workflows:
  wf1:
    steps:
    - script:
        service_containers:
        - *pg
        - redis
```

`getContainerReferenceFromInstance(..., ContainerType.Service, doc)` returns `[{ id: 'source', recreate: false }, { id: 'redis', recreate: false }]` (verified). `StepContainersTab` renders a service-container chip labelled `source`, and its Remove calls `removeContainerReference(..., 'source')` against a container that does not exist — the same "bogus chip with working actions behind it" the new `Array.isArray` guard three lines up exists to prevent.

This shape is pre-existing rather than introduced here, but it contradicts the scope note: an aliased entry renders a *wrong* badge, not no badge. Either reject alias nodes in the sequence (`isAlias`) or serialize through the document so the alias resolves.

### Checked and not flagged

- The remaining `getMapIn` / `getSeqIn` callers (`EnvVarService`, `TriggerService`, `ToolsService`, `EntityIndexService`, `WorkflowService`) are all read-then-throw. The new contract just swaps a low-level shape error for a domain "not found" there — no corruption path.
- `StrictMode` double-invokes the fallback effect in dev, so `maxResets: 3` is ~2 recoveries there. The loop still terminates, so it doesn't matter.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

PR #1913, opened 2026-09-16
Render the CI configuration header with BitkitPageHeader [FRONTEND-604]

doomsayer13:
## Summary

The CI configuration header is now rendered by `BitkitPageHeader` instead of a hand-assembled flex row. The design file calls this exact layout `PageHeader-WFE`, and the component's `controls` slot spreads the row into the three groups this header already wanted: the breadcrumb trail, the Visual/YAML switcher, and the save actions. Pins `@bitrise/bitkit-v2` to `0.3.416` exactly rather than with a caret, so a breaking release cannot arrive unreviewed. That is a stable release: `BitkitPageHeader` shipped, so the branch no longer rides the beta it started on.

**Jira:** [FRONTEND-604](https://bitrise.atlassian.net/browse/FRONTEND-604)

<sub>📋 Full detail — changes, Storybook before/after, test plan — in the first comment below.</sub>

[FRONTEND-604]: https://bitrise.atlassian.net/browse/FRONTEND-604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


doomsayer13:
## Changes

- Pin `@bitrise/bitkit-v2` to `0.3.416` (exact, was `^0.3.388`). The branch started on the `0.3.402-beta.2587` cut of [`BitkitPageHeader` #344](https://github.com/bitrise-io/bitkit-v2/pull/344) and [`BitkitPageSidebar` #351](https://github.com/bitrise-io/bitkit-v2/pull/351); both have since shipped, so it is on a stable release.
- `Header.tsx` renders `BitkitPageHeader` with `variant="minimal"` and the `controls` slot instead of its own flex row, border and padding.
- v1 → v2 for the components in that row: `Breadcrumb`/`BreadcrumbLink` → `BitkitBreadcrumb`, `Button` → `BitkitButton`, `Tooltip` → `BitkitTooltip`. `Box` now comes from Chakra v3, matching the v2 component it wraps.
- The conditional crumbs are collected into an array — `BitkitBreadcrumb` types its children as breadcrumb elements only, so `{cond && <Item/>}` would widen them to `false` and fail the type-check.

- `ClarityUnmask.spec.tsx` — the reviewed-unmask-site map now expects 6 for `Header.tsx` instead of 4, and the v2 forwarding test gained `BitkitButton` and both breadcrumb items (see the masking note below).

`Navigation.tsx` is deliberately untouched and stays on bitkit v1 — see "Notes for Reviewers".

---

## Type of Change

- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [x] ♻️ Refactor
- [ ] 📝 Documentation
- [ ] 🎨 Style / formatting
- [x] 🧪 Tests
- [x] 🔧 Chore / maintenance
- [ ] 💥 Breaking change

---

## Storybook — what to review

- **This PR** (rebuilt on each push): https://storybook.services.bitrise.dev/projects/wfe/pr-1913/
- **Production** (`master`): https://storybook.services.bitrise.dev/projects/wfe/production/

The header is a layout change, so the rendered stories say far more than the diff: the same three groups have to land in the same places, at the same widths, in both runtime modes.

| Component | Story | This PR | Production | What changed / what to look at |
|---|---|---|---|---|
| `Header` | Website (monolith mode) | [open](https://storybook.services.bitrise.dev/projects/wfe/pr-1913/?path=/story/javascripts-components-header--website) | [open](https://storybook.services.bitrise.dev/projects/wfe/production/?path=/story/javascripts-components-header--website) | The three-group row — trail + settings menu, view switcher, actions. Row height and the gaps are the component's now, not ours |
| | CLI | [open](https://storybook.services.bitrise.dev/projects/wfe/pr-1913/?path=/story/javascripts-components-header--cli) | [open](https://storybook.services.bitrise.dev/projects/wfe/production/?path=/story/javascripts-components-header--cli) | CLI mode drops the Bitrise CI and project crumbs, so the trail is just the current item |
| | Invalid YAML | [open](https://storybook.services.bitrise.dev/projects/wfe/pr-1913/?path=/story/javascripts-components-header--invalid-yml) | [open](https://storybook.services.bitrise.dev/projects/wfe/production/?path=/story/javascripts-components-header--invalid-yml) | Disabled Save plus its tooltip, and the blocked Visual switch with its own tooltip — the states moved from `isDisabled`/`isLoading` to `state` |

<sub>Resize any of the three to mobile as well — see the note below, that width behaves differently now. Deep-link IDs were computed with Storybook's own `getStoryTitle`/`toId` against this repo's config; if one 404s before the rebuild finishes, navigate the sidebar under *javascripts/components/Header*. Not logged in? Open a base URL first, then the deep links resolve. The canonical PR base is whatever the `Storybook uploaded to:` bot comment says.</sub>

---

## Test Plan

- ✅ `tsc --noEmit` — clean
- ✅ `eslint` — clean on both changed files
- ✅ `ClarityUnmask.spec.tsx` — 5/5 passing (the exact-count guard and the stray-attribute guard both cover this change)
- [ ] Both runtime modes in Storybook (Website / CLI) — the crumbs, the settings menu and the nav items all branch on `isWebsiteMode`
- [ ] Save / Discard / Show diff enabled and disabled states, and the saving spinner
- [ ] `Cmd/Ctrl+S` still saves, and all seven dialogs still open

---

## Notes for Reviewers

**One deliberate behaviour change, on mobile.** The old header always rendered "CI configuration" as a breadcrumb item — in CLI mode at every width, in website mode only on desktop. `BitkitPageHeader` only collapses into the single row from `tablet` up; below that it stays expanded and renders `title` as a visible heading. So the trail now drops its current item below `tablet` and the title shows as a heading there instead. Net effect: mobile gains a real page title (and an `h1` it never had), and nothing shows twice. Worth a look at mobile width.

**Clarity masking: worth the closest look in this PR — a review round found a real bug here.**

The trail is masked by default and the fixed crumbs are opted out one by one, because the project name between them is customer data. That intent is unchanged, but the mechanism had to change: `BitkitBreadcrumb.Item` and `.CurrentItem` destructure only `href`/`children` and drop every other prop, so the attribute passed to the crumb never reached the DOM. Two labels that used to be readable in recordings had quietly gone back to being masked. The tag now sits on a `span` inside each label, which does render.

The source-count guard could not catch that — it counts occurrences in the file, and all six were there, just not all six in the DOM. So the v2 forwarding test now covers `BitkitButton` (which does forward) and both breadcrumb items in the `span` shape, which is the guard that actually asserts the attribute lands. If a future bitkit version starts forwarding, or a call site stops wrapping, that test fails instead of the trail silently re-masking.

The action buttons also moved from one shared opted-out wrapper to one attribute each, since the actions are the component's own slot now — four reviewed sites to six. All three render fixed labels, so nothing carrying configuration or user input was opted out.

The one deliberate gap left is the page title: it is passed as a string and the component exposes no hook onto its `h1`, so the only way to opt it out would be tagging the root — which would take the project name in the trail with it. It stays masked, which costs a little readability and leaks nothing. From `tablet` up the heading is `srOnly` anyway and the visible label is the trail's current item, which is opted out.

**`isDanger` has no v2 equivalent as a flag.** Discard was `variant="secondary" isDanger`; it is now `variant="danger-secondary"`, which is the v2 recipe's own name for that pairing.

**`hasSeparatorBeforeFirst` is gone.** The v1 breadcrumb took it to draw a leading separator on mobile; `BitkitBreadcrumb` has no such prop and owns its own separator rules. This is the one prop that didn't carry over.

**`Navigation.tsx` is untouched on purpose.** It is still bitkit v1 and still an 88px icon-only rail on mobile. `BitkitPageSidebar.Item` requires a text label and the root has no narrow mode, so migrating it now would lose that rail — it is blocked on the mobile navigation design question tracked on [FRONTEND-606](https://bitrise.atlassian.net/browse/FRONTEND-606).


[FRONTEND-606]: https://bitrise.atlassian.net/browse/FRONTEND-606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

source/javascripts/components/Header.tsx
```diff
@@ -346,103 +346,108 @@ const Header = () => {
...
+        }
+        title="CI configuration"
+        variant="minimal"
       >
-        <Button
+        <BitkitButton
           size="sm"
           className="diff"
           variant="secondary"
+          data-clarity-unmask="true"
```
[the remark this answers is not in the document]
doomsayer13: Good catch — fixed in 07b6120b. The map now expects 6 for `Header.tsx`, with a note recording why it widened: the actions are `BitkitPageHeader`'s own slot now, so the single wrapper that opted all three buttons out became one attribute per button. All three render fixed labels (Show diff, Discard, Save changes), so nothing carrying configuration or user input was opted out — which is the confirmation the map's comment asks for.

source/javascripts/components/Header.tsx
```diff
@@ -346,103 +346,108 @@ const Header = () => {
...
+              <BitkitSegmentedControl.Item icon={IconWebUi} value="visual" disabled={isParseError}>
+                Visual
+              </BitkitSegmentedControl.Item>
+              <BitkitSegmentedControl.Item icon={IconCode} value="yaml">
+                YAML
+              </BitkitSegmentedControl.Item>
+            </BitkitSegmentedControl>
+          </BitkitTooltip>
+        }
+        title="CI configuration"
```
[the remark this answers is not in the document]
doomsayer13: Deliberate, and now documented at the call site in 07b6120b. `title` is typed as a string and the component exposes no hook onto its `h1`, so the only way to opt the heading out is to tag the root — which opts out the entire subtree, including the project name in the trail. That trades a slightly less readable recording for not leaking customer data, which seems the right way round. Worth noting the blast radius is small: from `tablet` up the heading is `srOnly` and the visible label is the trail's current item, which is opted out — the masked heading only renders below `tablet`.

moczolaszlo (dismissed):
The `<span data-clarity-unmask>` wrappers are working around a bitkit-v2 bug: `BitkitBreadcrumb.Item` and `.CurrentItem` forward nothing but `href`/`children`, and the mobile layout drops the rest a second time when it rebuilds the trail from a two-key pick. That one is ours, and bitrise-io/bitkit-v2#410 fixes it — nothing to change here, but worth knowing the wrappers are temporary.

One thing to fix: **in CLI mode below `tablet` the `breadcrumb` slot renders an empty wrapper.** All three crumb conditions are false there — the first two are gated on `isWebsiteMode`, the current item on `!isMobile` — so `BitkitBreadcrumb` returns `null`, and `ConfigSettingsMenu` is website-only too. The prop is still a `<Box>` element though, which `isEmptyNode` counts as present, so the header applies its `hasBreadcrumb` variant (`paddingBlockStart` 24 instead of 16) and the empty wrapper still consumes a 16px `headerBlock` gap. The title ends up ~24px lower over nothing. Pass `undefined` when there is nothing to put in it.

*AI-assisted review: the digging through source is Claude's, the judgement and every ask here are mine. Not convinced by a point? Ask — I'll show the working.*

doomsayer13:
Good catch on the empty wrapper — fixed in e464ee23.

You're right that all three conditions collapse in CLI below `tablet`: the first two crumbs are gated on `isWebsiteMode`, the current item on `!isMobile`, and `ConfigSettingsMenu` is website-only as well. Since `BitkitPageHeader` decides its layout on the slot's *presence* rather than its contents, the empty `Box` still bought the `hasBreadcrumb` padding and a `headerBlock` gap. The trail is now built as a value that collapses to `undefined` when neither the crumbs nor the settings menu would render, so the slot is genuinely absent in that case.

Thanks for the pointer to bitrise-io/bitkit-v2#410 — good to know the `<span data-clarity-unmask>` wrappers have an expiry date. I'd found the forwarding gap the hard way: the attribute was originally on the crumbs themselves, the source-text count guard passed because all six occurrences were in the file, and only the DOM told the truth. I've left the shape pinned in `ClarityUnmask.spec.tsx`'s v2 forwarding test, so once #410 lands that test is what will flag the wrappers as removable rather than them quietly lingering.

moczolaszlo (dismissed):

moczolaszlo (dismissed):

AndrasEszes (commented):
Arriving after the approvals, so just the gaps. One that isn't line-anchorable: the description and the Changes comment both still say `0.3.402-beta.2587` and link it to bitkit-v2 #344/#351 as "the beta cut", but `package.json` pins `0.3.416` and the lock resolves that. The bot's own summary already disagrees with the text — worth fixing in both places.

source/javascripts/components/Header.tsx
```diff
@@ -346,103 +346,125 @@ const Header = () => {
...
-      alignItems={['flex-start', 'center']}
-    >
+  // BitkitBreadcrumb types its children as breadcrumb elements only, so the conditional crumbs are
+  // spread into an array rather than rendered with `&&`, which would widen them to `false`.
+  //
+  // The fixed crumbs are unmasked one by one on purpose — the project name between them is customer
+  // data, so the Breadcrumb itself must stay masked.
+  //
+  // The tag goes on a span inside the label rather than on the crumb: BitkitBreadcrumb.Item and
+  // .CurrentItem destructure only `href`/`children` and drop every other prop, so tagging the crumb
```
AndrasEszes: #410 landed in 0.3.407 and this PR pins 0.3.416, so `Item` and `CurrentItem` both spread `{...rest}` onto the link now — tagging the crumb directly does reach the DOM. Should the spans go and `data-clarity-unmask` move back onto the two crumbs, with this comment removed? The count stays 6 either way, so the `EXPECTED` map is unaffected.

source/javascripts/lib/ClarityUnmask.spec.tsx
```diff
@@ -198,11 +205,32 @@ describe('Clarity unmasking', () => {
               action menu item
             </BitkitActionMenu.Item>
           </BitkitActionMenu>
+          <BitkitButton data-clarity-unmask="true">bitkit button</BitkitButton>
+          {/* Tagged on a span inside the label, not on the crumb: BitkitBreadcrumb.Item and
+              .CurrentItem destructure only `href`/`children` and drop every other prop, so tagging
+              the crumb is silently inert. Pinned in this shape so that a version which starts
```
AndrasEszes: One thing about this being the guard: the span carries the attribute itself, so `getByText(...).closest(UNMASK_SELECTOR)` finds the span whether or not the crumb forwards. The test can't tell you when the wrappers become removable — which is probably why the 0.3.416 bump slipped past. Reword, or drop it together with the spans?

source/javascripts/components/Header.tsx
```diff
@@ -346,103 +346,125 @@ const Header = () => {
...
-          )}
-          {isWebsiteMode && appPath && appName && <BreadcrumbLink href={appPath}>{appName}</BreadcrumbLink>}
-          {(!isWebsiteMode || !isMobile) && (
-            <BreadcrumbLink isCurrentPage data-clarity-unmask="true">
-              CI configuration
-            </BreadcrumbLink>
-          )}
-        </Breadcrumb>
+        <BitkitBreadcrumb>{breadcrumbItems}</BitkitBreadcrumb>
         {isWebsiteMode && <ConfigSettingsMenu />}
```
AndrasEszes, line 389: The PageHeader `breadcrumb` slot is `overflow: hidden`, and the Breadcrumb recipe gives `minWidth: 0` + truncate only to the crumb with nothing after it — here the fixed "CI configuration". So a long project name never truncates, this Box overflows, and `ConfigSettingsMenu` gets clipped off the right edge. I read that from the recipes rather than seeing it happen — could you try the Website story at ~1280px with a 30-character project name, and reach for `BitkitBreadcrumb.OverflowMenu` if it does?

source/javascripts/components/Header.tsx
```diff
@@ -346,103 +346,125 @@ const Header = () => {
...
+  // spread into an array rather than rendered with `&&`, which would widen them to `false`.
+  //
+  // The fixed crumbs are unmasked one by one on purpose — the project name between them is customer
+  // data, so the Breadcrumb itself must stay masked.
+  //
+  // The tag goes on a span inside the label rather than on the crumb: BitkitBreadcrumb.Item and
+  // .CurrentItem destructure only `href`/`children` and drop every other prop, so tagging the crumb
+  // itself is silently inert. `ClarityUnmask.spec.tsx` pins this shape.
+  const breadcrumbItems = [
+    ...(isWebsiteMode && !isMobile
```
AndrasEszes: `!isMobile` here looks like a v1 leftover: v1's `Breadcrumb` had no mobile collapsing, but `BitkitBreadcrumb` folds every non-last `Item` into an overflow menu below `tablet`. Dropping it would put the link back within reach on mobile instead of removing it — worth doing, or keep as is?

source/javascripts/components/Header.tsx
```diff
@@ -346,103 +346,125 @@ const Header = () => {
...
           onClick={onDiscard}
-          isDisabled={!hasChanges || isSavingConfig}
+          state={!hasChanges || isSavingConfig ? 'disabled' : undefined}
         >
           Discard
-        </Button>
-        <Tooltip
-          isDisabled={ymlStatus !== 'invalid'}
+        </BitkitButton>
+        <BitkitTooltip
```
AndrasEszes, line 447: Pre-existing, not something you broke: the button is `disabled` in exactly the state this tooltip explains, and a disabled `<button>` gets no pointer events and no focus, so "YAML is invalid…" never shows. v1 had the same gap. `BitkitTooltip` has `wrapTrigger` for this now — add it while you're in the file, or leave it for its own PR?

source/javascripts/components/Header.tsx
doomsayer13: You are right, and I checked the installed 0.3.416 rather than taking #410 on trust: `BreadcrumbItem` renders `<Breadcrumb.Link ref {...rest} href>` and `BreadcrumbCurrentItem` the same onto `Breadcrumb.CurrentLink`, so the attribute does reach the DOM now.

Spans dropped in 3685dabf, `data-clarity-unmask` back on the two crumbs, and the stale paragraph is gone from the comment. Count is unchanged at 6, so the `EXPECTED` map is untouched, and the spec still passes.

source/javascripts/lib/ClarityUnmask.spec.tsx
doomsayer13: Agreed — as written it could not fail. With the attribute on the span, `closest(UNMASK_SELECTOR)` found the span itself no matter what the crumb did, which is exactly why the version bump went unnoticed.

Updated with the spans in 3685dabf: the fixture now tags `BitkitBreadcrumb.Item` and `.CurrentItem` directly, so the assertion only passes while the components forward, and the comment says that is what it guards.

source/javascripts/components/Header.tsx
doomsayer13: Checked the component rather than assuming: `BitkitBreadcrumbRoot` takes `useBreakpointValue({ base: true, tablet: false })` and, when mobile, renders the last link plus the preceding ones in an overflow menu. So the guard was removing the link rather than letting it collapse — dropped in 3685dabf.

The `!isMobile` on the `CurrentItem` stays. It is there for a different reason (the title is a visible heading below tablet, so the trail would repeat it), and as it happens `flattenItemProps` only collects `Item`/`OverflowMenu`, so a `CurrentItem` contributes nothing on mobile anyway.

source/javascripts/components/Header.tsx
doomsayer13: Worth doing here — it is one prop and the file is already open. `BitkitTooltip` has `wrapTrigger`, and its own doc describes this case; added in 3685dabf.

Flagging the caveat from that doc so it is on the record: the wrapper is a plain focusable span, so focus announces the trigger label and the tooltip text, but not the button's disabled state. Conveying that too would mean `aria-disabled` instead of `disabled`, which is a bigger change than this PR should carry.

source/javascripts/components/Header.tsx
doomsayer13: Confirmed the mechanism from the recipes, but not the rendering — jsdom does no layout, so I could not watch it happen either.

What I can verify: the PageHeader `breadcrumb` slot is `display: flex, maxWidth: 100%, minWidth: 0, overflow: hidden`, and the Breadcrumb recipe gives `minWidth: 0` only via `item: { "&:not(:has(~ .breadcrumb__item))": { minWidth: 0 } }` — the crumb with nothing after it. On this trail that is the fixed "CI configuration", so the project name sits in a flex item that cannot shrink, and `truncateStyle` on the link never gets the chance to ellipse it. Your reading holds.

I stopped short of reaching for `OverflowMenu` because that changes the trail's shape and I could not confirm the overflow first. Note `Item` spreads its rest props onto the link, not the `<li>`, so `minWidth` cannot be passed from the call site as a cheaper fix.

Instead 3685dabf adds a `WebsiteWithLongProjectName` story (31-character name) so the case is one click away at 1280px. If it clips, `OverflowMenu` is the change — happy to make it once we have seen it.

AndrasEszes (approved):

